PR #10, opened 2017-06-18
Drop recompose, fix types

jaredpalmer:
- Drops `recompose`
- Hoist non-React statics to preserve static vars declared on child component
- Improve TypeScript types
- `displayName` is now optional


src/index.tsx
```diff
@@ -44,42 +39,77 @@ export function validateFormData<T>(
...
+  [field: string]: any;
+}
+
+// @todo make this limited to keys of values
+export interface FormikErrors {
+  [field: string]: string;
 }
 
+// @todo make this limited to keys of values
+// interfact FormikTouched<Values, Field extends keyof Values> ??
```
jaredpalmer: There is probably a nice way to express this in TS. All we are doing is mapings keys to boolean.

PR #47, opened 2017-07-11
Use rollup instead of webpack

jaredpalmer:
With this PR, Formik is just `4.23 KB`, `1.58 KB` minified gzipped.


miriam-z:
Does it mean that webpack would not work with Formik then?

TrySound:
@miriam-z It's internal change. Rollup just gives smaller bundle for fork itself. Usage is not changed.

PR #53, opened 2017-07-13
0.8

jaredpalmer:
- [x] Adds deprecation warnings to `setError`
- [x] Adds deprecation warnings to `handleChangeValue`
- [x] Adds deprecation warning to `mapValuesToPayload`
- [x] Adds `setFieldValue`, `setFieldError`, `setFieldTouched` methods
- [x] Adds `status` and `setStatus`
- [x] Adds warning about using `handleChange` with React Native
- [x] Adds new `validate` method. It will run sync or async depending on what is returned (object or Promise)
- [x] Makes `validationSchema` optional, Yup is now optional too
- [x] Adds `validateOnChange?: boolean` option to config
- [x] Adds `validateOnBlur?: boolean` option to config
- [x] Improved tests
- [x] Update documentation

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -1,36 +1,8 @@
 import * as React from 'react';
 
-import { isPromise, isReactNative, touchAllFields } from './utils';
+import { isPromise, isReactNative } from './utils';
 
 import { hoistNonReactStatics } from './hoistStatics';
-
-/**
- * Transform Yup ValidationError to a more usable object
- */
-export function yupToFormErrors(yupError: any): FormikErrors {
-  let errors: FormikErrors = {};
-  for (let err of yupError.inner) {
-    if (!errors[err.path]) {
-      errors[err.path] = err.message;
-    }
-  }
-  return errors;
-}
-
-/**
- * Validate a yup schema.
- */
-export function validateYupSchema<T>(data: T, schema: any): Promise<void> {
-  let validateData: any = {};
-  for (let k in data) {
-    if (data.hasOwnProperty(k)) {
-      const key = String(k);
-      validateData[key] =
-        (data as any)[key] !== '' ? (data as any)[key] : undefined;
-    }
-  }
-  return schema.validate(validateData, { abortEarly: false });
-}
 
 /**
  * Values of fields in the form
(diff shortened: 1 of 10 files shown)


src/formik.tsx
```diff
@@ -217,75 +255,125 @@ Formik cannot determine which value to update. See docs for more information: ht
...
+
+        if (validateOnChange) {
+          this.validateFormik({ ...this.state.values as any, [field]: value });
+        }
+      };
+
+      handleChangeValue = (field: string, value: any) => {
+        if (process.env.NODE_ENV !== 'production') {
+          console.warn(`
+          Warning: Formik\'s handleChangeValue will be deprecated in future releases. Please use Formik's setFieldValue(field, value) together with setTouched(field, isTouched) instead.
```
eonwhite: "is deprecated and may be removed in future releases"

src/formik.tsx
```diff
@@ -314,10 +404,13 @@ Formik cannot determine which value to update. See docs for more information: ht
             {...this.props as any}
             {...this.state as any}
             setError={this.setError}
+            setFieldError={this.setFieldError}
             setErrors={this.setErrors}
             setSubmitting={this.setSubmitting}
             setTouched={this.setTouched}
+            setFieldTouched={this.setFieldTouched}
             setValues={this.setValues}
+            setFieldValues={this.setFieldValue}
```
eonwhite: should be `setFieldValue=`

src/formik.tsx
jaredpalmer: Need to also add a link to the correct place in readme before this gets pushed.

src/formik.tsx
```diff
@@ -217,75 +255,125 @@ Formik cannot determine which value to update. See docs for more information: ht
...
           touched: {
-            ...touched as object,
-            [field]: true,
+            ...prevState.touched,
+            [field]: touched,
+          },
+        }));
+
+        // Async validate
+        if (!validateOnChange) {
```
eonwhite: What if we add a `validateOnTouch` option so you can separately control whether you want a change, a touch, both, or neither, to validate?

src/formik.tsx
```diff
@@ -91,14 +95,20 @@ export interface FormikState<V> {
...
   /* Manually set isSubmitting */
   setSubmitting: (isSubmitting: boolean) => void;
-  /* Manually set touched fields */
+  /* Manually set touched object */
   setTouched: (touched: FormikTouched) => void;
-  /* Manually set values  */
+  /* Manually set values object  */
   setValues: (values: V) => void;
+  /* Set value of form field directly */
+  setFieldValue: (field: string, value: any) => void;
```
eonwhite: Weren't we going to add a `touched?: boolean` to the end of this signature? I think I like it better without it actually on the basis of single responsibility principle, but we should acknowledge that the consequence is that a number of components will have to make calls to both `setFieldValue` and `setFieldTouched`, meaning double the number of state updates, which is sort of annoying.

jaredpalmer: Hrmmm... React should batch the two calls together (in theory). Remember `setState` is just a _request_ to change state. https://facebook.github.io/react/docs/react-component.html#setstate

My thinking is that you are already into custom component land here, explicit > implicit. Maybe we  have this wrong and it should just be
```tsx
setField(field: string, partialStates: {
   value: any;
   touched: boolean;
   error: string
})
```
or we can throw our hands up and just give `setFormikState()` and let people do whatever.

jaredpalmer: The reason we got into this whole `setXThing` in the first place is because Formik 0 was built with recompose which create an HOC and an updater function for each state key. This is no . longer needed

eonwhite: I don't mind the idea of `setField()` as an additional option, as long as `setFieldValue()`, `setFieldTouched()`, and `setFieldError()` are still also available. I think you still need the latter three because `setField()`'s signature would be so specific to Formik and it's a little more cumbersome to use if you're just setting one thing.

Let's definitely not give up and add `setFormikState()` yet -- that pushes a ton of inappropriate responsibility to the caller. We may need it as an escape hatch at some point but it shouldn't be the blessed way to do anything.

jaredpalmer: Another thing that blows about `touched?` is that you have no clue what it is doing when you read it in a file.

eonwhite: Yeah, agree. Let's not include it in the signature. I like these functions being single-purpose. Up to you if you want to add `setField()`, I'm on the fence.

jaredpalmer: Let's pause on `setField` too have feeling it will bite later.

src/formik.tsx
jaredpalmer: @eonwhite this would mean we would need to pass two flags to each one of our forms internally (if the default stays as onBlur = true).

eonwhite: We should think this one through a bit. As you know there's also a request out there to validate only individual fields at a time, which we might want to incorporate as well. If not in this version, we should at least set up well for that in the future. Perhaps there's a `ValidationConfig` object where you can set the general rules around validation at a top level, and make some mechanism for overriding ValidationConfig for individual fields if you want.

Issue #61, opened 2017-07-17
[0.8] Add validate and validateAsync methods


jaredpalmer:

cloud-walker:
Why two methods? Why not a unified promise based solution?

jaredpalmer:
Good idea! Should just be able see if it's a promise or not.

Issue #42, opened 2017-07-10
Remote submit


anhldbk:
How to achieve `remote submit` with formik? This useful feature from [Redux-form](http://redux-form.com/6.6.3/examples/remoteSubmit/)

jaredpalmer:
See the discussion in https://github.com/jaredpalmer/formik/pull/25 .

Since Formik does not use context, whatever ultimately calls `submitForm` must be a descendant of the form anyways. Not sure having a remote submit will help you. 

However, I am 100% open to a `submitForm` if you can provide concrete example of when this would be preferred over the current API.

anhldbk:
Thank you @jaredpalmer

rhyek:
@jaredpalmer I am building my own Form component where what I want it to do is have its own `handleSubmit` that would call Formik's `submitForm`, grab any http errors that might arise, and if the errors contain validation errors from the server, do `setErrors` to get that consistent visual feedback.

I am currently reusing the same `yup` validation schema with a few changes to it on the backend.

The only way for me to achieve this would be for `submitForm` to return a promise.

The idea would be something like:

```js
class Form extends React.Component<Props> {
  handleSubmit = async event => {
    event.preventDefault()
    const { formik } = this.context
    try {
      await formik.submitForm()
    } catch (error) {
      if (error.response && error.response.data && error.response.data.validationErrors) {
        formik.setErrors(error.response.data.validationErrors)
      } else {
        throw error
      }
    }
  }

  render() {
    const { children } = this.props
    return <form onSubmit={event => this.handleSubmit(event)}>{children}</form>
  }
}

Form.contextTypes = {
  formik: PropTypes.object,
}
```

The issue there, obviously, is `submitForm` does not return a promise I can `await` on and so it rejects after I've exited my `try/catch` leaving me with an unhandled promise rejection.

Is there any way to achieve what I want, currently? Or would this, indeed, require changes to `submitForm`?

RodolfoSilva:
Small example of Remote Submit with Formik.

https://codesandbox.io/s/w6nj3zl2l7

![55b64728-4629-4da6-92f5-683183a8a725](https://user-images.githubusercontent.com/1703020/48573747-f75fa800-e8eb-11e8-89b7-da7c7facab5e.gif)

RodolfoSilva:
Small example of Remote Submit with Formik.

https://codesandbox.io/s/w6nj3zl2l7

[...]

PR #25, opened 2017-07-03
Adds 'submitForm' to FormikBag

ctrlplusb:
Similar to the custom `resetForm` action.

Useful in cases where you aren't depending on a DOM element to perform the submit.

I was looking to add a test, but I see you are still forming your strategy here.

--

Great library @jaredpalmer and @eonwhite 👍

PR #63, opened 2017-07-18
[0.8] Add singular validate method

jaredpalmer:
- Adds singular `validate?: (values: Values, props: Props) => FormikErrors | Promise<any>` configuration option. Validate may either return a map of error messages `{ [field: string]: string}` or a promise whose error is a map of error messages. Unfortunately TypeScript does not let me express the shape of the error object of the promise with a type signature. So that kind of sucks. #61 
-  Adds `submitForm` to `FormikBag`. #25 #42

Contrary to my initial opinion described in #25 and #42 (and while not that useful in the component itself), having a `submitForm` function turns out to be _extremely_ useful as way to test Formik submissions without dealing with or simulating a submission event. This needs to be documented as well as other testing recipes in a guide.

The problem is that a React synthetic submission event is synchronous. This is fine for when validate is also sync, but when it comes to testing an async validation function the event, the call to the consumer's `handleSubmit()` is swallowed. However, if we encapsulate validation and submission into its own function (`submitForm()`) and thus move it away from the synthetic event, we can just [test that function using enzyme's mount](https://github.com/jaredpalmer/formik/blob/cbbb08f2bd7aaaed33ccea687fd96aa0dc56ca86/test/formik.test.tsx#L249) for the async test case + and test [`e.preventDefault()`](https://github.com/jaredpalmer/formik/blob/cbbb08f2bd7aaaed33ccea687fd96aa0dc56ca86/test/formik.test.tsx#L132) is getting called and we are :100:.

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 
+import { isPromise, isReactNative, touchAllFields } from './utils';
+
 import { hoistNonReactStatics } from './hoistStatics';
-import { isReactNative } from './isReactNative';
 
 /**
  * Transform Yup ValidationError to a more usable object
@@ -17,19 +18,9 @@ export function yupToFormErrors(yupError: any): FormikErrors {
 }
 
 /**
- * Given an object, map keys to boolean
- */
-export function touchAllFields<T>(fields: T): { [field: string]: boolean } {
-  let touched: { [field: string]: boolean } = {};
-  for (let k of Object.keys(fields)) {
-    touched[k] = true;
-  }
-  return touched;
-}
-/**
  * Validate a yup schema.
  */
-export function validateFormData<T>(data: T, schema: any): Promise<void> {
+export function validateYupSchema<T>(data: T, schema: any): Promise<void> {
   let validateData: any = {};
   for (let k in data) {
     if (data.hasOwnProperty(k)) {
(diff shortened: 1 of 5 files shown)


jaredpalmer:
@ctrlplusb you were right. `submitForm` is very handy for testing. As this PR / #53 is quite a large shift from your original one #25 , I went ahead and just added it (but without docs).

src/formik.tsx
```diff
@@ -248,42 +254,40 @@ export function Formik<Props, Values extends FormikValues, Payload>({
...
+       */
       runValidations = (values: Values) => {
         if (validate) {
-          this.runValidationSync(values);
-        }
-        if (validateAsync) {
-          this.runValidationAsync(values);
+          const maybePromisedErrors = validate(values, this.props);
+          if (isPromise(maybePromisedErrors)) {
+            (validate(values, this.props) as Promise<any>).then(
```
Andreyco: Second call to `(validate(values, this.props) as Promise<any>)` causes second validation run here. It feasible to run `maybePromisedErrors.then()` here - not sure whether you loose some of typehints here...

src/formik.tsx
```diff
@@ -382,34 +386,46 @@ Formik cannot determine which value to update. See docs for more information: ht
...
-          const errors = validate(this.state.values);
-          this.setState({ errors, isSubmitting: errors.keys.length > 0 });
-
-          // only call if there are no errors errors
-          if (errors.keys.length === 0) {
-            this.submitForm();
+          const maybePromisedErrors =
+            validate(this.state.values, this.props) || {};
+          if (isPromise(maybePromisedErrors)) {
+            (validate(this.state.values, this.props) as Promise<any>).then(
```
Andreyco: Second call to (validate(values, this.props) as Promise<any>) causes second validation run here. It feasible to run maybePromisedErrors.then() here - not sure whether you loose some of typehints here...

Issue #50, opened 2017-07-12
Testing problem with Enzyme, validationSchema ignored?


lukaszgadomski:
Hey, 
first the ball, thanks for nice form library.
I'm struggling with enzyme. I want to test that bad input values are properly validated. 
After simulate submit on `<form>` with enzyme, Formik component have state isSubmitting == true, and my  validationSchema is ignored. 
Here is the code of my form component:
```

// component/form/index.js
import React from "react";
import { Formik } from "formik";
import Yup from "yup";

const SimpleForm = ({
  values,
  touched,
  errors,
  error,
  handleChange,
  handleSubmit,
  handleBlur,
  handleReset,
  isSubmitting
}) => (
  <form onSubmit={handleSubmit}>
    <div className="basic-grey">
      <input
        type="text"
        name="firstname"
        value={values.firstname}
        onChange={handleChange}
        onBlur={handleBlur}
        placeholder="firstname"
      />
      {errors.firstname && <div className="error-msg">{errors.firstname}</div>}
      <div className="center buttons">
        <button
          className="button button-reset"
          onClick={handleReset}
          disabled={isSubmitting}
        >
          Reset
        </button>
        <button
          className="button button-submit"
          type="submit"
          disabled={isSubmitting}
        >
          Submit
        </button>
      </div>
    </div>
  </form>
);

export default Formik({
  displayName: "validationFAIL",
  validationSchema: Yup.object().shape({
    firstname: Yup.string().required()
  }),
  mapPropsToValues: props => ({
    firstname: props.firstname || ""
  })
})(SimpleForm);
```

and here is test for component, jest framework, (project is created using create-react-app):
```
// component/form/index.test.js
import React from "react";
import MyUserForm from "./index.js";
import { shallow, mount, render } from "enzyme";
import { expect } from "chai";

describe("MyUserForm", () => {
  describe("validate", () => {
    it("validate correctly", () => {
      const myUserForm = mount(<MyUserForm />);
      expect(myUserForm.state().isSubmitting).to.equal(false); //ok false == false
      myUserForm.find("form").simulate("submit");
      expect(myUserForm.state().isSubmitting).to.equal(false); //FAIL true == false WHY? :C
    });
  });
});
```

jaredpalmer:
[...]

PR #85, opened 2017-07-24
Add testing recipes to docs

### LINKED ISSUE #65: [Docs] Add testing recipes / guide
Reference: #68, 

### Story:
As user of Formik, I want to know how to test my Formik forms and inputs with popular test frameworks. 

### Rough Outline
- Using Jest snapshots
  - This is a quick way to test `mapPropsToValues`
- Using Enzyme
  - Explain when/how to use `mount` and `shallow`
  - Explain ways to simulate and test `change`, `blur`, and `submit` events
  - Explain how to use `submitForm()` to test async `handleSubmit`'s
  - Add tips about using `data-test-id=".."` attributes
- Using Karma / Selenium / Nightmare / Headless Chrome idk??
  - (may need help with this)

jaredpalmer:
Closes #65 
Related #50 

This is work in progress. ... Right now it's basically a recipe, but ideally this section should go into detail explaining the big gotcha: why promises inside of Enzyme's sync `simulate()`  will be swallowed (see due to Enzyme's  https://github.com/airbnb/enzyme/issues/823). 

This is going to result in confusion to newcomers when testing updates to inner form ui because the results of an async `validate` or `validationSchema` will neither resolve nor seem to fully update their UI intuitively. More specifically... 

- Using `simulate('submit', {..})` on an inner form component with an async `validate` function or `validationSchema` Formik option can only test updates as far [as this `setState` call](https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L392). 

- Depending on a form's `validateOnChange` or `validateOnBlur` options (defaults are `false` and `true` respectively), either the `simulate('change')` or `simulate('blur')` will have a similar issue with testing UI updates related to `errors`. Synchronous calls to `setState()` avoid this issue entirely and so testing changes to `values` and `touched` via `simulate()` do not have this problem.

Note: The above does not impact form's that have synchronous `validate` functions.

The suggested approach expose helpers that allow various events to be tested outside an event callback.

From Jordan Harband of Airbnb:

> No, there's no way to await the outcome of an event listener. Separately, `this.setState` is async and takes a callback, so for your promise to work properly it'd need to pass a callback and resolve the wrapping new Promise when `setState` is done....**What I'd recommend is having your event listener call into an instance method; spy on the instance method to test the event listener, and then unit-test the instance method.**
>  Source: https://github.com/airbnb/enzyme/issues/823#issuecomment-282213559 (emphasis added)

- [x] Recipe for `submitForm()`
- [ ] Recipe for testing `validationSchema`

Want to open this up for comment about the best approach to documenting all of this...

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,10 @@ You can also try before you buy with this **[demo of Formik on CodeSandbox.io](h
   - [React Native](#react-native)
     - [Why use `setFieldValue` instead of `handleChange`?](#why-use-setfieldvalue-instead-of-handlechange)
     - [Avoiding a Render Callback](#avoiding-a-render-callback)
+  - [Testing Formik](#testing-formik)
+    - [Dummy Form](#dummy-form)
+    - [Simulating input](#simulating-input)
+    - [Simulating form submission](#simulating-form-submission)
 - [Authors](#authors)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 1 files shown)


Ciaran0:
Has this testing documentation been removed?

janhesters:
This is great! Could you also add how to test the `renderForm()` passed to the `render` prop?

PR #92, opened 2017-07-26
Bugfix: status should be reset on a resetForm

eonwhite:
Formik 0.8 introduced the `status` field, which we're switching to rather than the deprecated `error`. However, `status` was not added to the list of fields to reset on a `resetForm` or `handleReset` (both these functions correctly reset `error`) This PR addresses that.


jaredpalmer:
Pushing 8.5 now

PR #97, opened 2017-07-28
Closes #94. Add isValid prop and isInitialValid option

### LINKED ISSUE #94: RFC: isValid prop for the whole form
Hey guys!

First of all, we've really enjoyed using Formik in our products. It's made form submission so much easier. One pain point that we have in our usage though is that there's no easy way to always tell if the whole form is valid. 

This would be nice for doing something like disabling the submit button based off one prop instead of having to check for other things. For example, as far as I can tell nothing shows up in the errors array until at least one field has been typed in. So although the form is invalid, you can't tell that just by checking the errors array.

Just wanted to see if anyone else thought something like that would be a good idea... Thanks again for the awesome library!

jaredpalmer:


src/formik.tsx
```diff
@@ -61,6 +61,8 @@ export interface FormikConfig<
   validateOnChange?: boolean;
   /** Tells Formik to validate the form on each input's onBlur event */
   validateOnBlur?: boolean;
+  /** Tell Formik if initial form values are valid or not on first render */
+  isInitialValid?: (props: Props) => boolean;
```
jaredpalmer: I couldn't get typescript to allow me to make this a`boolean | (props: Props) => boolean` unless we make it a required option (which we don't want). Thoughts? Am I missing something?

eonwhite: Can you do 'boolean | (props: Props) => boolean | undefined'?

jaredpalmer: what do we want the default to be? `false`?

eonwhite: Yeah, I would think so. More common to have a form that needs filling out, than an edit-form that starts life valid.

jaredpalmer: Had to cast but figured it out.

bcarroll22:
🙌 just saw this, very cool. Pretty sure this is exactly what we're looking for. Thanks @jaredpalmer!

Issue #99, opened 2017-07-31
Custom `validate` method runs twice


Andreyco:
- bug appears for async validation only.
- subsequent validation attempts are not handled properly
```
Possible Unhandled Promise Rejection (id: 0):
Object {
  "email": "email validation failed on email",
}
```
- This bug is caused by duplicated call of `validate` method. See comment on https://github.com/jaredpalmer/formik/pull/63/files

PR #100, opened 2017-07-31
Fix double invocation of async `validate` method.

Andreyco:
This also fixes incorrect Promise handling, when multiple runs of `validate` method caused Promises to remain unhandled.


Andreyco:
Fixes #99

jaredpalmer:
Good catch!

Andreyco:
Thanks for very quick response & merge! Gotta love Formik with such care...

Issue #87, opened 2017-07-24
Who uses Formik?


jaredpalmer:
If your company is using Formik in production, please let us know who are you so we can mention you in the new README.

eonwhite:
Stardog Ventures (https://www.stardog.io)

iralution:
elbstack (https://www.elbstack.com)

bcarroll22:
[Prolio](www.prolio.com) uses Formik pretty extensively. Not only that but we retroactively changed all of our forms over to it in very little time. Lots of messy class components that we got to simplify to functional components because of Formik 👍

guilhermedecampo:
Astrocoders (https://astrocoders.com/). Formik is awesome!!!

cloud-walker:
[Wanderio](https://www.wanderio.com/)!

drewlustro:
I'm working to replace **Formsy**, which is unfortunately not maintained frequently, with **Formik** at [Capsule](https://www.capsulecares.com/).

Has anyone used **Formik** by wrapping it in another custom higher-order component to grant standard enhancements for all your forms?

Magellol:
I've just opened a RFC on our [company](http://www.automat.ai/) application repo to move to Formik :)

<img width="1051" alt="screen shot 2017-08-16 at 9 15 24 am" src="https://user-images.githubusercontent.com/3819093/29365055-9b0a5ebe-8263-11e7-8202-d999ba4f808d.png">

Magellol:
@drewlustro Since HOCs are just functions, I think you could compose Formik and yours together to provide a set of props for all your forms.

```javascript
import { withState, compose } from 'recompose';
import { Formik } from 'formik';

const withSuperFormPowers = compose(
  withState('myField', 'myStateUpdater', 'hello world'),
  Formik({...}),
);

const App = withSuperFormPowers(({
  myField,
  myStateUpdater,
  ...formikProps,
}) => {
  console.log(formikProps);
  return <div>{myField}</div>;
});
```

drewlustro:
@Magellol v cool - Digging into _recompose_ as a result!

julienvincent:
[OneDayOnly](www.onedayonly.co.za)

dphaener:
FermentAble - www.getfermentable.com

morrisallison:
[OpenTable](https://www.opentable.com)

cjbell:
[Frame.io](https://frame.io) 😎

jonthomp:
[...]

PR #103, opened 2017-07-31
Fix isValid logic and add tests

### LINKED ISSUE #102: `isValid` is true even when there are invalid fields
#97 introduces `isInitialValid` which marks the form invalid when it's initialized. This gives a confusing result as soon as the form is dirty.

~Because `isValid` works by counting the number of items in `errors`, it doesn't check for fields which haven't had blur events.~ In our project we observed this as a submit button going from disabled to enabled after entering a valid value into an input, even though the other fields weren't valid. This feels like an unexpected behaviour.

Do you have a pattern to deal with this? I understand that you don't want fields to show errors until they're been blurred, but for this case I'd expect formik to keep track of invalid fields even if they haven't been touched

Thanks
Adam

jaredpalmer:
Fixes #102 . Adds tests to all logic branches as well to ensure correct future behavior

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -532,7 +532,8 @@ Formik cannot determine which value to update. For more info see https://github.
             dirty={dirty}
             isValid={
               dirty
-                ? this.state.errors && Object.keys(this.state.errors).length > 0
+                ? this.state.errors &&
+                  Object.keys(this.state.errors).length === 0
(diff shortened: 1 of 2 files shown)


skattyadz:
👍 Looks good to me

jaredpalmer:
@skattyadz Thanks! 

Make sure to add your company to #87 !

Issue #108, opened 2017-08-01
Crash on React Native Android


Andreyco:
Formik crashes on hoisting of statics.

Related to:
- apollographql/react-apollo#832
- mridgway/hoist-non-react-statics#32

jaredpalmer:
Just got to the office. Will push a patch right meow 🐱

Andreyco:
My pleasure, patch has been in progress already. Thanks for attention! :)

PR #109, opened 2017-08-01
Fix hoisting of statics

Andreyco:


Andreyco:
Fixes #108

src/hoistStatics.tsx
```diff
@@ -11,6 +11,16 @@ const REACT_STATICS: any = {
   type: true,
 };
 
+var KNOWN_STATICS: any = {
```
jaredpalmer: nit: use `let` here

Andreyco: I think `const` will serve even better here.

jaredpalmer: Yes, my b

jaredpalmer:
LGTM

Andreyco:
1450460008joys-of-block-scoping-with-es602-make-let-not-var

Issue #119, opened 2017-08-10
isSubmitting gets set to true when the form is submitted, even when there are validation errors


scbrady:
When submit a form with formik, the `isSubmitting` prop is set to `true`, triggering a render and causing submit logic to fire in my view. This all works great unless there are validation errors. When I submit a form that has validation errors, `isSubmitting` is set to `true`, a render happens, then it is set to `false`. This causes weird glitches.

Is there a reason why `isSubmitting` is set to true, even when it isn't actually submitting? If not, I'd be happy to submit a PR.

jaredpalmer:
Feel free to submit a PR

jaredpalmer:
Right so `isSubmitting` is set to `true` because the user is trying to submit. It is set `false` if the form is found to be invalid. I guess it is somewhat confusing in the sense that `handleSubmit` may or may not be called. Not sure about the double render you are seeing. Can you give an example?

scbrady:
Sure! I have a custom button:

```
<Button showsLoading submit isSubmitting={isSubmitting}>
```

That fires a redux action to make our page show a loading state as soon as `isSubmitting` is set to `true`. I then hide the loading state at the end of the `handleSubmit` function.

I've already worked around this problem in my code, but it just seemed like a bug and confused me for a while until I put in breakpoints to see what was happening. I just figured I'd check in to see if this was the intended behavior or not.

jaredpalmer:
It is. Although you do raise a point about naming confusion

PR #121, opened 2017-08-10
Set isSubmitting to true on executeSubmit

scbrady:
This makes sure that isSubmitting is only true when it is actually submitting the form.

Proposed fix for issue #119

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -401,7 +401,6 @@ Formik cannot determine which value to update. For more info see https://github.
       submitForm = () => {
         this.setState({
           touched: touchAllFields<FormikTouched<Values>>(this.state.values),
-          isSubmitting: true,
         });
 
         if (validate) {
@@ -436,6 +435,10 @@ Formik cannot determine which value to update. For more info see https://github.
       };
 
       executeSubmit = () => {
+        this.setState({
+          isSubmitting: true,
+        });
+
         const values = mapValuesToPayload
           ? mapValuesToPayload(this.state.values)
           : this.state.values;


marcneander:
Why was this never merged?

Lets say I have this login form below. I want to disable and show a loader ontop of my submit button when the handleSubmit is executing. Without this PR I get rerenders on my submit button when I click on it causing bad UX in mobile (browser never has the time to trigger button:active styling).

I can of course solve this by creating my own submitting prop but I just thought it was a weird behaviour as default.

@jaredpalmer 

```js
async submitHandler(values, { setSubmitting }) {
    const response = await this.props.authenticate({
        username: values.username,
        password: values.password
    });

    if (response.error) {
        setSubmitting(false);
    }
}


<Formik
    onSubmit={this.submitHandler}
    validationSchema={yup.object().shape({
        username: yup.string().required('Username missing'),
        password: yup.string().required('Password missing')
    })}
    validateOnBlur={false}
>
    {props => (
        <CustomForm onSubmit={handleSubmit} noValidate name="LoginForm">
            <CustomField
                type="text"
                name="username"
                error={props.errors.username}
                onChange={props.handleChange}
                onBlur={props.handleBlur}
                value={props.values.username}
            />
            <CustomField
                type="password"
                name="password"
                error={props.errors.password}
                onChange={props.handleChange}
                onBlur={props.handleBlur}
                value={props.values.password}
            />
            <Button type="submit" disabled={props.isSubmitting} className={props.isSubmitting ? 'btn-loading' : ''}>
                {props.isSubmitting ? <i className="fa fa-circle-o-notch fa-spin" /> : 'Log in'}
            </Button>
        </CustomForm>
    )}
</Formik>
```

freemanon:
In any scenario, currently submitting an invalid form re-renders the form twice.

Issue #116, opened 2017-08-09
Issues with Formik


bduff9:
Hi there!  I really like this library you have set up and have had little issue integrating in place of existing forms.  I have a couple concerns, that are possible I am doing something wrong with.  Since I love the idea of the plugin, I wanted to verify what I might be missing before anything else.

1. Are checkboxes supported?  I don't see any mention of them in the docs nor in the issues, and they don't seem to work as I would expect.  For instance, if I have 3 checkboxes with name "color" and each has a value of a color (i.e. red, blue, green) I would expect this to be retrieved as an array (i.e. if red and blue were checked, ['red', 'blue'])  Currently, the value is ignored and its returned as true or undefined.  This also makes validation difficult as in order to have checkboxes, I have to give them multiple names and then somehow validate that at least one of the names is checked, instead of validating that its an array of length that is at least 1.

2. What about radio buttons?  These seem to be better handled as the actual value of the selected radio button is returned, however until one is changed, no value is returned, and I had an issue in my brief test of changing the value back after making a selection.  For instance, if I have yes and no radio buttons, with no checked by default, the form first returns undefined for the radio buttons even though no is checked.  Once I select Yes, the form does return yes, but then I am unable to re-select no.

3. Form elements in general seem to have some issues upfront.  For instance, a select that I default the value of does not return a value upon submit.  Similar to radio buttons, making a change does update it but this is counter intuitive.  Additionally, the examples all list checking if the field has an error and is touched before showing an error which makes perfect sense to me ({errors.field_name && touched.field_name && ...})  However, if a field that is required is not touched, when submitted the form fails to submit but no error is shown.  I think the initialization of the form in general needs some work, but in this one case, I would've imagined that if I ever click the submit button, Formik would set every field's touched value to true to avoid this issue.

I hope the above is clear without too much code as I am currently evaluating this library.  If not, please let me know and I would be happy to provide additional details.  Thanks again!

jaredpalmer:
@bduff9 Formik handle's checkboxes exactly the way they are handled in the React documentation with `true` / `false`:  https://facebook.github.io/react/docs/forms.html#handling-multiple-inputs . However, you could probably write your own custom inputs to get the behavior you described above.

I am working on a patch to fix touch behavior with select, radio, checkboxes now.

jaredpalmer:
So after exploring various form libraries, the general consensus is to not automatically touch select, radio, and checkboxes onChange. 

[...]

PR #123, opened 2017-08-14
#114 Add touchOnBlur and touchOnChange options

### LINKED ISSUE #114: Should a change event mark the value as touched?
Some background:
I have a form where I validate on change, and use the isValid prop to control whether the submit button is enabled or disabled.

The problem for me is that isValid is computed based on the dirty values, and a field isn't considered touched/dirty until a blur happens. I want the submit button to be enabled as soon as the form is in valid state, even while typing. Right now this doesn't happen until an input loses focus.

This feels weird to me. validateOnChange gives me the impression that it in some way should have a say over the isValid prop.

This would be fixed by marking a value as touched when an onChange happens, which makes sense to me at least. While typing, I am definitely "touching" it.

I could provide a pull request for this if you are interested.

And finally, thank you so much for this great library!

jaredpalmer:
Fix #114 and #116 

Adds `touchOnBlur?: boolean = true` and `touchOnChange?: boolean = false` options to Formik configuration. This gives much needed control over updates to `touched`.

**To Discuss**

- I wonder if `setFieldValue()` should accept an optional third argument that will set `touched` regardless of config like so:

```js
props.setFieldValue(field: string, value: any, setTouched?: boolean) => void;
```
- Are these defaults correct?

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ You can also try before you buy with this **[demo of Formik on CodeSandbox.io](h
         - [The "FormikBag":](#the-formikbag)
       - [`isInitialValid?: boolean | (props: Props) => boolean`](#isinitialvalid-boolean--props-props--boolean)
       - [`mapPropsToValues?: (props: Props) => Values`](#mappropstovalues-props-props--values)
+      - [`touchOnBlur?: boolean`](#touchonblur-boolean)
+      - [`touchOnChange?: boolean`](#touchonchange-boolean)
       - [`validate?: (values: Values, props: Props) => FormikError<Values> | Promise<any>`](#validate-values-values-props-props--formikerrorvalues--promiseany)
       - [`validateOnBlur?: boolean`](#validateonblur-boolean)
       - [`validateOnChange?: boolean`](#validateonchange-boolean)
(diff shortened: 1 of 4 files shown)


jaredpalmer:
cc @eonwhite I think this and #126 should get merged, we push a patch release, and then we move all efforts to renderProps and new docs.

jaredpalmer:
@eonwhite despite this PR, which maintains existing default behavior related to setting `touched`. I think in 0.9/1.0 renderprops we should default to `touchOnChange: true` and `touchOnBlur: false`.

eonwhite:
It's always good to offer some flexibility, but I'm not sure what the usecase is here. Certainly we shouldn't change the default to `touchOnChange: true`.

The `touched` field should normally be used to determine whether to display an error message, a success state (e.g. a green checkbox), or neither.

It's wrong to show people an error message, or a success state, the moment they hit their first keystroke when they're first typing into a blank field. I've seen forms that do this in the wild, but generally not on well-designed websites. It's totally distracting.

Instead you want to show an error or success, after the user has indicated that they're done typing their input. The best way to determine whether the user has indicated that they're done is a blur event.

Of course, if a field has already been fully entered and it's an error, then it makes sense for the user to get realtime typing feedback as they go back and edit the field. This is the advantage of using `validateOnChange:true`, which we should probably make the default. But realtime validation feedback is not appropriate for all forms, so `validateOnChange/Blur` made sense as an option. I don't understand the form that wants `touchOnChange` to be default behavior for text inputs.

jaredpalmer:
Going to close for now and move efforts to #130

third774:
@eonwhite - I'm not sure `validateOnChange` is useful for this because it validates the entire form, right? Won't it generate error messages for inputs that haven't been touched yet?

In my use-case, I have a `FieldFeedback` component that I want to debounce display of errors in, and I'm running into issues where I can't rely on checking if an error exists, and I can't rely on checking whether the field is touched or not because of this.

third774:
Perhaps the solution is to have `dirty` be an object instead of a boolean similar to how `touched` is.

PR #126, opened 2017-08-15
Add `isValidating` prop to indicate async validation process

husa:
add `isValidating` processing to `runValidations` and `submitForm`
add `setValidating` method as manual setter
remove trailing whitespace

In order to somehow indicate that async validation is in the process(show spinner or smth) it would be great to have `isValidating` prop.

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -37,13 +37,13 @@ export interface FormikConfig<
   displayName?: string;
   /** Map props to the form values */
   mapPropsToValues?: (props: Props) => Values;
-  /** 
-   * Map form values to submission payload 
+  /**
+   * Map form values to submission payload
    * @deprecated since 0.8.0
    */
   mapValuesToPayload?: (values: Values) => DeprecatedPayload;
-  /** 
-   * Validation function. Must return an error object or promise that 
+  /**
+   * Validation function. Must return an error object or promise that
    * throws an error object where that object keys map to corresponding value.
    */
   validate?: (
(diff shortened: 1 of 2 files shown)


husa:
@jaredpalmer any thoughts?

jaredpalmer:
Not merging anything into master until renderprops gets merged.

jaredpalmer:
This has nothing to do with merits of this PR. Just putting in hold until then

aodinok:
Hi @jaredpalmer, first of all thank you for cool form library)
I see that many issue was referenced to this PR, and having isValidating prop to show loading indicator on form while async validation is in progress sound like a good idea. I would like to add this indicator too.

What about reopen this? I can test it if you want.
Would be glad to help.

Thanks.

jaredpalmer:
yes I think it's time to reopen for v1

husa:
Any updates required?

jaredpalmer:
I

jaredpalmer:
@eonwhite should we change isSubmitting to only be true if and only if validation passes? Right now it turns true whenever submission is attempted (regardless of validation outcome). Would be breaking but is what ppl expect. Thoughts?

EdwardDrapkin:
Why not have a status field with enum values like VALIDATING, FAILED, VALID? That way we can handle all three potential states clearly.

husa:
I think, setting `isSubmitting` to `true`, while we're actually validating, does seem like a bug. So, yeah, for me it makes total sense to leave `isSubmitting` as `false` until validation succeeds.

SirNerdBear:
Would love this feature.

The problem with having isSubmitting to false while validating is behaviorally, if validation is taking more than a fraction of a second (such as an API call to verify username availability) then the user can spam click the submit button. And if the button was disabled during verification than it could become disabled while typing values in the form back and forth. 

isSubmitting should be true during validation on submit and return to false if validation fails.

PR #130, opened 2017-08-16
Next

jaredpalmer:
- [x] Make sure that `mapPropsToValues` backwards compat in `withFormik`. 
- [x] Update docs
- [ ] Rewrite the walk through
- [ ] Update sandboxes
- [ ] Determine if `<Field/>` and `<Form/>` should be the "blessed" approach in the docs
- [ ] Consider `<FieldError name="thing" />` and `<FieldTouched name="thing" />` which only render children if `!!errors[name]` and `!!touched[name]`
- [x] Write migration guide
- [ ] Seriously consider using `react-broadcast` instead of context directly


VladShcherbin:
Maybe `getInitialValues` needs to be changed to `initialValues` as it's more pleasant to write and to be consistent with other codebase (`validationSchema` instead of `getValidationSchema`) ?

jaredpalmer:
@VladShcherbin agree 100%.

VladShcherbin:
```jsx
const CustomInputComponent = () => (
  <p>input</p>
)

<Field component={CustomInputComponent} />
```

The above code gives a warning when using `alpha.9`:

```
`Warning: Failed prop type: Invalid prop `component` supplied to `Field`.
```

Seems like `component` prop [here](https://github.com/jaredpalmer/formik/blob/v0.9.0-alpha.9/src/Field.tsx#L42) expects `string` or `element` and `element` expects this to pass:

```jsx
<Field component={<CustomInputComponent />} />
```

A possible fix is to change `PropTypes.element` to `PropTypes.func`.

jaredpalmer:
Hmm thought I copied the props from React Router exactly. Let me look again

jaredpalmer:
@VladShcherbin good catch! pushing out a new release now

VladShcherbin:
@jaredpalmer that was fast, thank you.

VladShcherbin:
@jaredpalmer hey, what do you think about adding `touched`, `error` directly to `Field`'s `field` prop (`field.touched` instead of `form.touched[field.name]`), same as `value`, 👍  or 👎 ?

VladShcherbin:
I also found that in 0.9 by default `Field` is [validated](https://github.com/jaredpalmer/formik/blob/v0.9.0-beta.1/src/formik.tsx#L168-L169) `onChange` and is not validated `onBlur`. Not sure, if this was intentional or a typo.

jaredpalmer:
>  hey, what do you think about adding touched, error directly to Field's field prop (field.touched instead of form.touched[field.name]), same as value, 👍 or 👎 ?

problem is that field is meant to be spread `{...field}`

VladShcherbin:
@jaredpalmer oh, I see. We pass `{...field}` as props for `input` so all of this props must be supported by `input` component.

The reason why I asked is that in redux-form there is a similar [meta prop](http://redux-form.com/7.0.3/docs/api/Field.md/#meta-props) (now I understand, why it is separated there too). It's useful in such situations:

```jsx
const FormError = ({ touched, error }) => (
  (touched && error) ? <p>{error}</p> : null
)

// with meta prop
const CustomInput = ({ field, meta, ...rest }) => (
  <div className={generateFieldClassName(meta)}>
    <input {...field} />
    <FormError state={meta} />
  </div>
)

// current way
const CustomInput = ({ field, form, ...rest }) => {
  const fieldState = {
    touched: form.touched[field.name],
    error: form.errors[field.name]
  }

  <div className={generateFieldClassName(fieldState)}>
    <input {...field} />
    <FormError state={fieldState} />
  </div>
}
```

If some kind of `meta` prop is added, it can already contain some useful field info:

- touched - from `form.touched[field.name]`
- error - from `form.errors[field.name]`
- initialValue - didn't find a way to get it

jaredpalmer:
For all those watching this issue. Pretty happy with current api, will probably merge tomorrow

VladShcherbin:
@jaredpalmer check [this one](https://github.com/jaredpalmer/formik/pull/130#issuecomment-326712397), it still differs with the docs.

jaredpalmer:
Going to take another swing at docs tomorrow too

jaredpalmer:
here. we. go.

Issue #139, opened 2017-08-24
Checkbox input validation inconsistency


morrisallison:
Firstly, thanks for this project, it's pretty cool.

Our form is using `validationSchema`, with both `validateOnChange` and `validateOnBlur` on. Among other inputs, the form has a few checkboxes that are validated with `Yup.boolean()` in the schema.

Reproduce:

A. Toggle a checkbox

1. The checkbox's `value` property is set in `values`
1. The `values` are validated
1. Validation fails: `"on"` is not a boolean
1. The checkbox's `checked` property is set in `values`

B. Click outside the checkbox

1. The values are validated
1. Validation is successful: `true` is a boolean

This only seems to be an issue with `validateOnChange` turned on.

jaredpalmer:
This has been fixed in 0.9.1

vuhrmeister:
I'm experiencing exactly the same with 0.9.4

PR #165, opened 2017-09-20
Close #162, #139. Fix validation defaults

### LINKED ISSUE #162: validateOnBlur & validateOnChange defaults on v0.9
First time submitting an issue on Github. Please let me know if I'm doing it wrong!

I've noticed that 

validateOnBlur now defaults to false
& validateOnChange now defaults to true

but documentation states the opposite for both of them.

[https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx](url)

jaredpalmer:
Due to consensus, this PR sets `validateOnBlur` and `validateOnChange` to true by default in Formik. 

The suggested method for showing feedback to the user is officially:
```
{errors.field && touched.field && <div>{errors.field}</div>}
```

It's worth considering codifying this by creating a `<FieldFeedback name render />` component. Im not sure it's worth it as it's more likely people will include standardized errors in a custom `Field component/>`. However, an optimized `<FieldFeedback >` might be more performant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,7 +707,7 @@ const validate = (values, props) => {
 
 #### `validateOnBlur?: boolean`
 
-Default is `false`. Use this option to run validations on `blur` events. More specifically, when either [`handleBlur`], [`setFieldTouched`], or [`setTouched`] are called.
+Default is `true`. Use this option to run validations on `blur` events. More specifically, when either [`handleBlur`], [`setFieldTouched`], or [`setTouched`] are called.
 
 #### `validateOnChange?: boolean`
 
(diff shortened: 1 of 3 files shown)


eonwhite:
LGTM

jaredpalmer:
@eonwhite i figured out a way to save a render for those using sync validation. It's gross, but it works...

```jsx
handleChange = e => {
    e.persist();
    let asyncValidationPromise;
    // only want to update the value that has changed
    this.setState(
      state => {
        const newState = {
          ...state,
          values: {
            ...state.values,
            [e.target.name]: e.target.value,
          },
        };

        if (
          this.props.validateOnChange &&
          this.props.validate
        ) {
          const maybeErrors = this.props.validate({
            ...state.values,
            [e.target.name]: e.target.value,
          });

          if (!isPromise(maybeErrors)) {
            newState.errors = {
              ...state.errors,
              ...maybeErrors,
            };
          } else {
            asyncValidationPromise = maybeErrors;
          }
        }

        return newState;
      },
      () => {
        if (!!asyncValidationPromise) {
          asyncValidationPromise.then(
            () => this.setState({ errors: {} }),
            errors => this.setState({ errors })
          );
        }
      }
    );
  };
```

VladShcherbin:
@jaredpalmer in release notes

> After some discussion `validateOnChange` is now set to true by default.

probably `validateOnChange` -> `validateOnBlur`

PR #169, opened 2017-09-21
Fix check to call resetForm on componentWillReceiveProps

### LINKED ISSUE #164: componentWillReceiveProps calls resetForm even if initialValues hasn't changed
https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L236
```js
  componentWillReceiveProps(nextProps: Props) {
    // If the initialValues change, reset the form
    if (nextProps.initialValues !== this.props.initialValues) {
      this.resetForm(nextProps.initialValues);
    }
  }
```
```js
{email: 'foobar@email.com'} !== {email: 'foobar@email.com'} // true
```
It seems like the intent is shallow object comparison instead of object equality
```js
const shallowCompare = (a, b) => {
    for(var key in a) {
        if(!(key in b) || a[key] !== b[key]) {
            return false
        }
    }
    for(var key in b) {
        if(!(key in a) || a[key] !== b[key]) {
            return false
        }
    }
    return true
}

!shallowCompare({email: 'foobar@email.com'}, {email: 'foobar@email.com'}) // false
```
```js
if (!shallowCompare(nextProps.initialValues, this.props.initialValues)) {
  this.resetForm(nextProps.initialValues);
}
```

skvale:
initialValues comparison was using an object equality check with !== this was replaced with a deepequal check from lodash.isEqual

Fixes #164

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "gen-docs": "all-contributors generate && doctoc README.md"
   },
   "dependencies": {
-    "lodash.isequal": "^4.5.0",
+    "lodash.isequal": "4.5.0",
     "prop-types": "^15.5.10",
     "warning": "^3.0.0"
   },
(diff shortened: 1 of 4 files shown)


jaredpalmer:
Merging this now btw

PR #176, opened 2017-09-25
Run validations on set values and touched methods in a callback

### LINKED ISSUE #106: Old error shown when setFieldValue and setFieldTouched are executed together
After executing the following code, Formik shows an error based on the previous value. The order of the two lines doesn't make a difference.

```js
this.props.setFieldValue('rating', value);
this.props.setFieldTouched('rating', true);
```

Wrapping `setFieldTouched` in `setImmediate` solves the problem so I presume it's some kind of race condition?

jontansey:
Where multiple set values and touched methods are called in sequence
React will batch the setState meaning validation will run on previous
state. Putting the code which executes the validations in a callback on
set state should solve the issue

Fixes #106

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -268,17 +268,19 @@ export class Formik<
   };
 
   setTouched = (touched: FormikTouched) => {
-    this.setState({ touched });
-    if (this.props.validateOnBlur) {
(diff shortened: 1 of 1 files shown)


jaredpalmer:
LGTM

PR #179, opened 2017-09-26
Check `e.persist` exists before calling

willfarrell:
In `react-material-ui@1.0.0-beta.x` the `Select` component doesn't return an event w/ `persist()` causing an error to occur.


jaredpalmer:
If e doesn't exist, then you should use setFieldValue() and use a custom component.

Issue #161, opened 2017-09-18
Pristine values


tkvw:
I think it would be nice to remember the pristine(initial) values of the form and allow dirty to check if the new value is changed compared to this pristine value. Currently dirty is actually isTouched. Or is this a to complicated use case for formik?

jaredpalmer:
yeah i think this is a good idea. I would keep this outside of state in a class property like `this.initialValues`

Magellol:
We've actually enhanced Formik and provided a different version of dirty for our use cases.
We're using `componentWillReceiveProps` and we're comparing `props.values` with a property like `this.initialValues`. Then, once the form has been submitted, we're overring initialValues with the latest values.

jaredpalmer:
@Magellol this is going to be built into Formik now with #181

Magellol:
@jaredpalmer with the change in `componentWillReceivedProps`, does that mean the `dirty` prop will be updated correctly?

NamPNQ:
I think https://github.com/jaredpalmer/formik-persist is what u want

Issue #163, opened 2017-09-20
Option to not reset all form state when only some initialValues change


bogdansoare:
I have a Formik component which initials values depend on props

```jsx
<Formik
  initialValues={{
    foo: this.props.foo,
    bar: this.props.bar
  }}
  ...
```

I save each field `onBlur`, which also changes the corresponding prop of the component to keep in sync with the updated value on server (using graphql with apollo-client). This change of prop triggers reset form https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L236-L238 and with it losing all errors & touched info. Is there any way to shallow compare the values of the initialValues and only reset the ones that changed?

jaredpalmer:
Can pasted the component your having issues with?

jaredpalmer:
FYI: We did this because we couldn't come up with a pattern where this wasn't the desired result. So very curious to understand and dig into the code.

bogdansoare:
@jaredpalmer sure, will share tomorrow a codesandox with an example

bogdansoare:
Hey @jaredpalmer . Sorry it took so so long.

Here is the codesandox https://codesandbox.io/s/n31p94m320 . 

Here is a demo with the problem, when only one prop changes, the whole forms reset 
![formik-issue](https://user-images.githubusercontent.com/2750170/30782207-f31ecce8-a135-11e7-873b-d79eb9880d50.gif)

I think the solution would be to have an option which prevents the form from refreshing dirty values on initialValues change like @VladShcherbin said in his issue #168 . He made some great points, and how redux-form solves those issues.


Another related issue is #172 which mainly has the same problems

jaredpalmer:
closed by #169

bogdansoare:
Hey @jaredpalmer.

Unfortunately #169 doesn't fix the problem. I've updated the codesandox https://codesandbox.io/s/n31p94m320 to use 0.9.2 and it's the same issue.

Like I and @VladShcherbin said, the issue is that initialValues are actually different objects in componentWillReceiveProps, because a value in that object has changed and resets the whole form. I don't want the whole form to be resetted because I will lose the dirty values like in the example.

The solution may be to have an option which prevents the form from refreshing dirty values on initialValues change like @VladShcherbin said in his issue #168

jaredpalmer:
I have a quick and dirty fix in #181

jaredpalmer:
Let’s move discussion there

bogdansoare:
cool, sounds good
[...]

PR #181, opened 2017-09-27
Add enableReinitialize option

### LINKED ISSUE #168: Add enableReinitialize option and set it to false by default
Currently, if you pass `initialValues`, you can overwrite them by passing another value - the form will update its values.

This brings a problem when your data is stored for example in redux or in another component. You get this data and pass into form. Once this values are changed (in redux / another component) - your form *magically* updates and removes all data you've changed.

In redux-form there is a nice [enableReinitialize](https://redux-form.com/7.0.4/docs/api/reduxform.md/#-code-enablereinitialize-boolean-code-optional-) option which is `false` by default. You can set it to `true` and your form will update itself.

I think, the same behavior and option should be added to formik so your data won't suddenly disappear.

---

Related issues - https://github.com/jaredpalmer/formik/issues/163, https://github.com/jaredpalmer/formik/issues/164

For the first related issue, there is [keepDirtyOnReinitialize](https://redux-form.com/7.0.4/docs/api/reduxform.md/#-code-keepdirtyonreinitialize-boolean-code-optional-) option which prevents form from refreshing dirty values on `initialValues` change.

jaredpalmer:
- Adds a `enableReinitialize`  option, default is `false`. If true, form will reset when `initialValues` (or the result of `mapPropsToValues` in HoC land) is different (using deep equality). (#168 , #188) 
- Adds readonly `initialValues` prop to the Formik bag (#161)

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -51,11 +51,13 @@ export interface FormikState<Values> {
 /**
  * Formik computed properties. These are read-only.
  */
-export interface FormikComputedProps {
+export interface FormikComputedProps<Values> {
   /** True if any input has been touched. False otherwise. */
   readonly dirty: boolean;
   /** Result of isInitiallyValid on mount, then whether true values pass validation. */
   readonly isValid: boolean;
+  /** initialValues */
+  readonly initialValues: Values;
 }
 
 /**
@@ -115,6 +117,10 @@ export interface FormikSharedConfig {
   validateOnBlur?: boolean;
   /** Tell Formik if initial form values are valid or not on first render */
   isInitialValid?: boolean | ((props: object) => boolean | undefined);
+  /** Reinitialize when new initial va */
+  enableReinitialize?: boolean;
+  /** Keep dirty values on reinitialize */
+  keepDirtyOnReinitalize?: boolean;
 }
 
 /**
@@ -165,7 +171,7 @@ export interface FormikConfig extends FormikSharedConfig {
 export type FormikProps<Values> = FormikState<Values> &
   FormikActions<Values> &
   FormikHandlers &
-  FormikComputedProps;
+  FormikComputedProps<Values>;
 
 const isEmptyChildren = (children: any) => React.Children.count(children) === 0;
 
@@ -176,6 +182,8 @@ export class Formik<
     validateOnChange: true,
     validateOnBlur: true,
     isInitialValid: false,
+    enableReinitialize: false,
+    keepDirtyOnReinitialize: false,
   };
 
   static propTypes = {
@@ -194,6 +202,8 @@ export class Formik<
   static childContextTypes = {
     formik: PropTypes.object,
   };
+
+  initialValues: any;
 
   getChildContext() {
     const dirty =
@@ -237,13 +247,34 @@ export class Formik<
       touched: {},
       isSubmitting: false,
     };
+
+    this.initialValues = props.initialValues || ({} as any);
   }
 
   componentWillReceiveProps(nextProps: Props) {
     // If the initialValues change, reset the form
-    if (!isEqual(nextProps.initialValues, this.props.initialValues)) {
-      this.resetForm(nextProps.initialValues);
+    if (
+      this.props.enableReinitialize &&
+      !isEqual(nextProps.initialValues, this.props.initialValues)
+    ) {
+      if (this.props.keepDirtyOnReinitalize) {
+        const values = keepKeys(
+          nextProps.initialValues,
+          this.state.values,
+          this.state.touched
+        );
+        this.initialValues = values;
+        this.setState({ values });
+      } else {
+        this.initialValues = nextProps.initialValues;
+        this.resetForm(nextProps.initialValues);
+      }
     }
+    // if (this.props.enableReinitialize) {
+    //   if (!isEqual(nextProps.initialValues, this.props.initialValues)) {
+    //     this.resetForm(nextProps.initialValues);
+    //   }
+    // }
   }
 
   componentWillMount() {
@@ -582,6 +613,7 @@ Formik cannot determine which value to update. For more info see https://github.
         : isInitialValid !== false && isFunction(isInitialValid)
           ? (isInitialValid as (props: Props) => boolean)(this.props)
           : isInitialValid as boolean,
+      initialValues: this.initialValues,
       handleSubmit: this.handleSubmit,
       handleChange: this.handleChange,
       handleBlur: this.handleBlur,
@@ -646,6 +678,20 @@ export function touchAllFields<T>(fields: T): FormikTouched {
     touched[k] = true;
   }
   return touched;
+}
+
+export function keepKeys(
+  fresh: any,
+  old: any,
+  keys: { [field: string]: boolean }
+) {
+  const override = Object.keys(keys).reduce((prev: any, curr: string) => {
+    if (keys[curr]) {
+      prev[curr] = old[curr];
+    }
+    return prev;
+  }, {});
+  return { ...fresh, ...override };
 }
 
 export * from './Field';
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -117,10 +117,8 @@ export interface FormikSharedConfig {
   validateOnBlur?: boolean;
   /** Tell Formik if initial form values are valid or not on first render */
   isInitialValid?: boolean | ((props: object) => boolean | undefined);
-  /** Reinitialize when new initial va */
+  /** Should Formik reset the form when new initialValues change */
   enableReinitialize?: boolean;
-  /** Keep dirty values on reinitialize */
-  keepDirtyOnReinitalize?: boolean;
 }
 
 /**
@@ -183,7 +181,6 @@ export class Formik<
     validateOnBlur: true,
     isInitialValid: false,
     enableReinitialize: false,
-    keepDirtyOnReinitialize: false,
   };
 
   static propTypes = {
@@ -197,6 +194,7 @@ export class Formik<
     component: PropTypes.func,
     render: PropTypes.func,
     children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
+    enableReinitialize: PropTypes.bool,
   };
 
   static childContextTypes = {
@@ -235,6 +233,7 @@ export class Formik<
         setSubmitting: this.setSubmitting,
         resetForm: this.resetForm,
         submitForm: this.submitForm,
+        initialValues: this.initialValues,
       },
     };
   }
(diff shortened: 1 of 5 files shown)


src/formik.tsx
```diff
@@ -648,6 +680,20 @@ export function touchAllFields<T>(fields: T): FormikTouched {
   return touched;
 }
 
+export function keepKeys(
```
jaredpalmer: This is a codesmell.

jaredpalmer: I don't even think this is required at all.

jaredpalmer:
To everyone in #163 and #168. Here is the solution...

Imagine you are using Apollo and you have a subscription to a real-time form. You only want the user to see updates on fields that they haven't already visited. Updates come through a GraphQL subscription (that uses a web socket).

```js
// App
export default graphql(`subscriptionQuery...`)(Wrapper)

const Wrapper = ({ data }) => {
  // only want to mount the form once we have data
   return data ? <Outer data={data} /> : null
} 

class Outer extends React.Component {
  constructor(props) {
    super(props)
    // stamp data on mount
    this.initalValues = Object.assign({}, props.data)
  }

 render() {
  return  (
    <Formik 
       initialValues={this.initialValues}
       render={props => <Form {...props} initialData={this.initialValues} streamingData={this.props.data} /> }
      />
   )
 }
}

class Form extends React.Component {
 componentWillReceiveProps(nextProps) {
   // whenever data arrives...
   if (this.props.streamingData !== nextProps.streamingData) {
    // decide how to update Formik on your own
    const nextValues = figureOutWhatValuesToKeep(
      this.props.values, 
      this.props.touched, 
      nextProps.streamingData
    )
     const nextErrors = figureOutWhatErrorsToKeep(
      this.props.values, 
      this.props.touched, 
      nextProps.streamingData
    )
     this.props.setValues(nextValues)
     this.props.setErrors(nextErrors)
   }
 }

 render() {
  return  (
    <Form>
      <Field name="thing" />
      <Field name="other" />
      {/** .... */}
     </Form>
   )
 }
}
```

One of the beautiful things about the render prop is that we don't need to reimplement stuff that's already possible with just vanilla React. This is the exact same approach you would take if you were going to do this without formik. Instead of calling setValues, you'd just use your own setState.

bogdansoare:
Thanks for providing this solution.
Honestly I'm not really a fan of this solution you need to define a Wrapper and Outer and only after that you Form, seems like a lot of boilerplate, also setting values & errors seems more on the imperative side.

Why did you drop having the 2 options enableReinitialize and keepDirtyOnReinitialize?

jaredpalmer:
A few reasons....

1. I asked the team from apollo and they had not even done a form like this. Generally speaking, the goal is to satisfy 80% of a users, or put differently, that formik can be used on 80% of your apps forms. This falls outside those bounds. 

2. You can do this with plain React... You can reduce boilerplate above by abstracting and composing Outer and MyForm and spreading props so you get something like`<RealTimeFormik render={} />`.

bogdansoare:
Understood, you are right that this scenario is kinda of special.
Thank you very much for explanation and patience.

Keep up the good work with Formik ❤️

VladShcherbin:
@jaredpalmer I also disagree with closing this. This is a good default form behavior that I expect from any form library. This is how redux-form works and I've been using it since v4.

I don't understand, why do users of formik need to reinvent this and bring this dirty hack to their projects. The slogan of formik is `Build forms in React, without the tears` - this topic is exactly what brings tears to developers (as you can see, issues with this are already rising).

Huge 👎 for closing this from me. I really hope you'll change your mind and merge this to formik.

jaredpalmer:
@VladShcherbin Agree the first solution was not very reusable so I refactored it so that the end result exposes a Component with an identical API to Formik but with the desired behavior. **The beauty of this is that it is not a hack at all**, just React. 

```js
// App
export default graphql(`subscriptionQuery...`)(App)

const App = ({ data }) => 
  <RealTimeFormik 
    initialValues={data}
    onSubmit={...}
    render={props => 
      <Form>
        <Field name="thing"/>
        <Field name="other" />>
      </Form>
    }
  />


class RealTimeFormik extends React.Component {
  render() {
    if (!this.props.initialValues) {
      return null
    }

    return (
      <FormikWithStickyValues 
        {...this.props} 
        render={props => 
          <KeepDirty streamingData={this.props.initialValues} {...props}>
            {this.props.render(props)}
          </KeepDirty> 
        }
      />
    )
  }
}

class FormikWithStickyValues extends React.Component {
  constructor(props) {
    super(props)
    // stamp data on mount
    this.initalValues = Object.assign({}, props.initialValues)
  }

 render() {
   // exclude initialValues from pass thru
  const { initialValues, ...props} = this.props
  return  (
    <Formik 
       initialValues={this.initialValues}
       {...props}
    />
   )
 }
}


class KeepDirty extends React.Component {
  componentWillReceiveProps(nextProps) {
    // whenever data arrives...
    if (this.props.streamingData !== nextProps.streamingData) {
     // decide how to update Formik on your own
     const nextValues = figureOutWhatValuesToKeep(
       this.props.values, 
       this.props.touched, 
       nextProps.streamingData
     )
      const nextErrors = figureOutWhatErrorsToKeep(
       this.props.errors, 
       this.props.touched, 
       nextProps.streamingData
     )
      this.props.setValues(nextValues)
      this.props.setErrors(nextErrors)
    }
  }
 
  render() {
   return this.props.children
  }
 }```

**Comment by VladShcherbin** [comment]
@jaredpalmer yes, I understand that we can create a custom form with such behavior.

The problem is that it's not convenient to bring it from project to project. For example, I used redux-form before and want my forms to work this way by default - so in every project that uses formik I need to bring this.

There are already other developers that want this - they also need to make such custom form. This brings extra pain when it could be solved with 2 flags and everyone would be happy.

**Comment by eonwhite** [comment]
The problem is that every flag, no matter how individually worthy, adds cumulative complexity to both the surface of the API and its internals. Keep adding too many "one more little flags", and eventually new users get dissuaded by a giant scary-looking interface that's hard to learn, and @jaredpalmer will be doing nothing but answering support requests.

That's not to say that this particular change may not be worth the cost, but it's a real cost that has to be weighed.

In this case the cost that looked especially painful here was not just the increase of the API surface with an extra flag, but also adding a magical `keepKeys` internal implementation that was almost certain to cause confusion and increase maintenance cost.

Could a compromise be to build a couple composed components for the most common usecases and ship them with Formik, but keep them out of the core Formik component itself?

**Comment by VladShcherbin** [comment]
Personally, I'd be happy with just `enableReinitialize` as mentioned in the [issue](https://github.com/jaredpalmer/formik/issues/168).

I think form's initial values should not be changed by default with props changes as it brings unnecessary surprises. I really don't want to add `componentWillReceiveProps` to forms for this.

All the complexity comes with `keepDirtyOnReinitialize`.

**Comment by cloud-walker** [comment]
Just to notify that, this was a breaking change (worth bumping from 0.10 instead of 0.9.3).

**Comment by Skillvendor** [comment]
I consider this to be a very common use case. Why would you guys say it's special in any way.

I have 2 input fields validated each. Because of how reinitialize works if I don't fix my validation on the first field and go to the second I lose the validation upon typing something in the 2nd field... This should be a feature, I am sorry, can't agree with your reasoning. A simple flag and from 80% you go to 98%. (but i really doubt 80% don't have this use case).

**Comment by oakgary** [comment]
we are currently looking for a way to keep dirty values on initialValues change aswell
this is definitely something formik is missing

**Comment by kylebebak** [comment]
Is there no official solution to this problem?

I'd like to migrate a few projects I have to Formik, because rxf is no longer maintained, but the fact that it's missing `keepDirtyOnReinitialize` is a real bummer.

If we take the comments in this thread as any indication, this is a common use case, and ought to part of Formik's core API. If rxf was able to do it and do it well, I don't see why Formik can't.

If, however, I'm confused, and it's just an issue of not understanding the API, I think this would be a great place to improve the documentation. Is there a reusable way to get behavior equivalent to `keepDirtyOnReinitialize` using plain React? If so, could we get a working example somewhere in the docs, or at least in this thread?

@vladshcherbin @jaredpalmer

**Comment by jaredpalmer** [comment]
@kylebebak just to clarify the specific desired outcome is to keep touched but reset the rest of form state?

**Comment by kylebebak** [comment]
@jaredpalmer 

Hey Jared, thanks for the response!

By `keepDirtyOnReinitialize`, I mean that if a user passes in new `initialValues` to a Formik instance, that the current form values are not overwritten, even if they differ from the values in `initialValues`.

However, any current form values that are different from the new values passed in `initialValues` are immediately marked as being `dirty`.

One use case that is well-solved by `keepDirtyOnReinitialize` is a form with auto-save. As the user changes a value in the form, e.g. by typing into a text input, the field is marked as dirty and a PUT request or whatever is sent to the API with the new form values.

If the request is successful, the form's `initialValues` can be updated to whatever the most recent successfully saved values are. If the user continued to type while the request was being sent, the value for the text field will still be marked as dirty when `initialValues` is updated, and another PUT request can be sent, **but it won't be truncated**.

**Comment by kylebebak** [comment]
@jaredpalmer 

What do you think... Is it worth adding this to the core API?

**Comment by jwmann** [comment]
+1 for `keepDirtyOnReinitialize`

I have to update my initial values because some of my form fields only "exist" on step 2 depending on fields in step 1.

So naturally, I want to provide an initial value for those fields but changing it reset the entire form.
I tried using `setValues` to force the dirty values to stay but updating my intialValues via a useState hook re-renders the page and either `setValues` does nothing or the `reinitialize` overwrites it

**Comment by dorklord23** [comment]
Another +1 for `keepDirtyOnReinitialize`. I'm having a hard time to accept that this is not a common case.

Thank you for the hard work @jaredpalmer

**Comment by johnrom** [comment]
I think adding a new bool prop would put too much power in the hands of the maintainers. For example, what if you want to keep dirty, but want only the newly updated values no longer touched? Now we need some weird new prop like `keepOldFieldsTouchedOnReinitialize={true}`. Instead, I think we could accept a callback to enableReinitialize. When `true`, the callback would do exactly what it does now, but one could instead do something like:

```
<Formik 
    initialValues={newValues} 
    enableReinitialize={(newProps, oldProps) => ({
        ...newProps,
        // or touched, or isDirty, or whatever you want to override
        values: {
            ...oldProps.values,
            ...newProps.initialValues,
        },
    })} 
/>
```

dorklord23:
On second thought, your idea sounds better and more scalable @johnrom

jwmann:
@johnrom has a way better idea and would give us way more fine control on how it reinitializes.

Either way, I'm just hoping for some kind of progress. I would even be happy with a boolean at this point haha

BadMask121:
@johnrom ideas better, currently facing this issue, I think @jaredpalmer this should be a feature, all my inputs are being reinitialized without dirty inputs and its beginning to feel like a bug. thanks

BinitaBharati:
So, I have to use enableReinitialize={true} so that my form field values get updated from redux. But, as being discussed ihere, all the other unrelated fields ( i.e the ones that never changed state on redux) also get reset - consequently, the errors is also reset, and other unrelated fields, which should have been showing error, do not show. I have to click on submit again, to see the same errors again.

ming1in:
any updates on this recently?

Marius-Adam:
Any updates on this?

gelembungsahabat:
any update on this ?

rgallison:
updates?

Issue #182, opened 2017-09-27
Field custom component cannot be an inline lambda


julienvincent:
I am wanting to use `<Field />` with the following pattern:

```js
<Field
  name="details.longName"
  component={({field, form}) => {
    return (
      <CustomInput
        {...field}
        onTextChange={value => form.setFieldValue(field.name, value)}
      />
    )
  }}
/>
```

However after each key-press within the input component, focus is lost. If I pull the wrapping `({field, form}) => { ... }` out and make it a constant then everything works as expected.

Note the `onTextChange` prop in my example might make it seem like this is a react native component, I am experiencing this on web.

julienvincent:
Taking a look at the source, it seems it can be easily remedied. I will submit a PR proposal.

PR #183, opened 2017-09-27
Add render props to <Field />

julienvincent:
This PR allows the usage pattern described in #182

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -8,7 +8,7 @@ import { FormikProps } from './formik';
  * context and wiring up forms.
  */
 export const Field: React.SFC<any> = (
-  { component = 'input', name, ...props },
+  { component = 'input', render, children, name, ...props },
   context
 ) => {
   const field = {
@@ -31,8 +31,11 @@ export const Field: React.SFC<any> = (
     ...props,
     ...bag,
   };
-  if (typeof component === 'function') {
-    return component(componentProps);
+  if (typeof render === 'function') {
+    return render(componentProps);
+  }
+  if (typeof children === 'function') {
+    return children(componentProps);
   }
   return React.createElement(component, componentProps);
 };
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -27,10 +27,14 @@ export const Field: React.SFC<any> = (
           field,
           form: context.formik,
         };
-  return React.createElement(component, {
+  const componentProps = {
     ...props,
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Is there really a need for this though? Just pass props if you need to write this inline.

julienvincent:
Passing props as apposed to using the context api adds a significant amount of ceremony, especially in complex forms spread across multiple nested components.

Using  the `<Field />` component api is very elegant, but unfortunately it does not conform to the normal React `onChange` + `value` props pattern. This means that integration with existing custom components is not as simple as just passing them to the `component` prop as they won't be compatible.

#### tl;dr
Adding support for inline lambdas will make compatibility with various custom components much simpler and will allow smoother and easier migration to Formik. Less tears :)

jaredpalmer:
By passing props i meant something like this. (this is the way it worked before context 0.9) 

```js
<Formik
   render={props => 
    <form onSubmit={props.handleSubmit}>
      <input 
         name="email"
         type="email"
         onChange={props.handleChange} 
         onBlur={props.handleBlur} 
         value={values.email} 
       />
    </form>}
/>

```

julienvincent:
Yea I'm with you on that. My use case is one in which my form input components are sitting in several nested components (complex capture forms) and so those Formik props would need to travel down through two or perhaps three container components before arriving at their respective input component.

Using the context api cleans this up significantly (at the cost perhaps of obscuring links between data and capture).

jaredpalmer:
I found myself wanting this the other day. Can you add render, component, and children ? should be similar to logic found in <Formik />

julienvincent:
You didn't finish writing the comment but I'm assuming you are referring to react-router? If so it might make sense to leave out the `children` prop according to [this react-router issue](https://github.com/ReactTraining/react-router/issues/5583)

jaredpalmer:
forgot to escape a character

julienvincent:
Ok I see now. I will make the changes tomorrow 👍

jaredpalmer:
Yeah just implement children for now to stay consistent. Just one line. Once React Router decides on children we'll follow suit.

julienvincent:
Alright I added a `render` and `children` as props. The implementation prioritizes render => children => component, not sure if you would like to change these priorities?

julienvincent:
Any updates here?

jaredpalmer:
Gonna merge and make the changes after.

PR #186, opened 2017-09-28
[Field] Improve typings

Grmiade:
Add missing `name` field in `InputProps`.
⚠️ BreakingChange: `FieldProps` has been renamed to `FieldComponentProps`. It's more coherent like that I think (`react-router` package uses this naming too)

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ const EditUserDialog = ({ user, updateUser, onClose }) => {
               touched.twitter &&
               <div>
                 {errors.twitter}
-              </div>}          
+              </div>}
             <button type="submit" disabled={isSubmitting}>
               Submit
             </button>
@@ -435,7 +435,7 @@ const EditUserDialog = ({ user, updateUser, onClose }) => {
               touched.twitter &&
               <div>
                 {errors.twitter}
-              </div>}          
+              </div>}
             <button type="submit" disabled={isSubmitting}>
               Submit
             </button>
@@ -722,9 +722,9 @@ Default is `true`. Use this option to tell Formik to run validations on `change`
 
 `<Field />` will automagically hook up inputs to Formik. It uses the `name` attribute to match up with Formik state. `<Field/>` will default to and `<input/>` element. To change the underlying element of `<Field/>`, specify a `component` prop. It can either be a string like `select` or another React component.
 
-```js
+```typescript
 import  React from 'react';
-import { Formik, Field } from 'formik';
+import { Field, FieldComponentProps, Formik } from 'formik';
 
 const Example = () => (
   <div>
@@ -737,7 +737,7 @@ const Example = () => (
           actions.setSubmitting(false)
         }, 1000);
       }}
-      render={(props: FormikProps<Values>) =>
+      render={props =>
         <form onSubmit={props.handleSubmit}>
           <Field type="email" name="email" placeholder="Email" />
           <Field component="select" name="color" >
@@ -752,7 +752,9 @@ const Example = () => (
   </div>
 );
 
-const CustomInputComponent: React.SFC<FormikProps<Values> & CustomInputProps> => ({
+type CustomInputProps = { customProp: any }
+
+const CustomInputComponent: React.SFC<CustomInputProps & FieldComponentProps> => ({
   field, // { name, value, onChange, onBlur }
   form: { touched, errors } // also values, setXXXX, handleXXXX, isDirty, isValid, status, etc.
   ...props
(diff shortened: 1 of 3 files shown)


src/Field.tsx
```diff
@@ -3,15 +3,33 @@ import * as React from 'react';
 
 import { FormikProps } from './formik';
 
+export interface FieldProps {
+  component: keyof React.ReactHTML | React.ComponentType<any>;
+  name: string;
+  type?: string;
```
jaredpalmer: What about the placeholder, autoComplete, etc?

src/Field.tsx
```diff
@@ -3,15 +3,33 @@ import * as React from 'react';
...
+  value?: any;
+}
+
+export interface InputProps {
+  /** Name of the input */
+  name: string;
+  /** Classic React change handler, keyed by input name */
+  onChange: (event: React.ChangeEvent<any>) => void;
```
jaredpalmer: ditto

README.md
```diff
@@ -752,7 +752,9 @@ const Example = () => (
   </div>
 );
 
-const CustomInputComponent: React.SFC<FormikProps<Values> & CustomInputProps> => ({
+type CustomInputProps = { customProp: any }
+
+const CustomInputComponent: React.SFC<CustomInputProps & FieldComponentProps> => ({
```
jaredpalmer: This makes sense

README.md
```diff
@@ -737,7 +737,7 @@ const Example = () => (
           actions.setSubmitting(false)
         }, 1000);
       }}
-      render={(props: FormikProps<Values>) =>
+      render={props =>
```
jaredpalmer: Why did `FormikProps<Values>` go away here?

src/Field.tsx
```diff
@@ -20,14 +38,16 @@ export const Field: React.SFC<any> = (
...
     typeof component === 'string'
       ? field
       : {
           field,
           form: context.formik,
         };
-  return React.createElement(component, {
+  // Cast component to React.ComponentClass to prevent typescript error
+  // https://github.com/Microsoft/TypeScript/issues/15019
+  return React.createElement(component as React.ComponentClass<any>, {
```
jaredpalmer: Nice

README.md
Grmiade: Here, types are inferred. No need to specify them ;)

src/Field.tsx
Grmiade: The `Field` component only need this props to work. 
The placeholder, autoComplete, ... definitions should be in the CustomInputProps, no ?

src/Field.tsx
Grmiade: We only accept these 4 props for [field](https://github.com/jaredpalmer/formik/pull/186/files#diff-bc78961df770c0df68f6cd758a6f2afdR32)

README.md
jaredpalmer: Yes, but they would be inferred as `FormikProps<any>`, unless I am mistaken (which i could be)

src/Field.tsx
jaredpalmer: Confused as to why we need these InputProps at all? This is never exposed to the end user.

jaredpalmer: Ahh I see this is used at the bottom of the file.

src/Field.tsx
```diff
@@ -37,38 +57,26 @@ Field.contextTypes = {
...
-export interface FieldProps {
-  field: {
-    /** Classic React change handler, keyed by input name */
-    onChange: (e: React.ChangeEvent<any>) => void;
-    /** Mark input as touched */
-    onBlur: (e: any) => void;
-    /** Value of the input */
-    value: any;
-  };
+export interface FieldComponentProps {
```
jaredpalmer: I wonder if we should give people the option to type this strictly? 

```tsx
interface FieldComponentProps<Values = any> {
   field: FieldProps,
   form: FormikProps<Values>
}
```

Grmiade: Idk if it will be useful but it's not a bad idea to provide to end users this flexibility :)
I will add this.

README.md
Grmiade: Indeed. But in this case we don't have `Values` type. And the `render` method doesn't use `props.values`. So I thought this explicit type was useless.

jaredpalmer:
If you change the breaking names, i can merge

Grmiade:
How I rename `FieldProps`? Just `Props`? :/

jaredpalmer:
see #225

PR #193, opened 2017-10-04
Rework validation

megazoll:
Rework validation to avoid some code duplication


jaredpalmer:
Thanks for this. Will take a look and comment later tonight.

jaredpalmer:
Can you refactor to avoid `SyncPromise` stuff?

megazoll:
`SyncPromise` is looks really strange, but it allows us to keep single interface for validators. Otherwise we will have to checking what type of result returns validator (Promise or something sync) every time we will call validator. Do you think it's a better way?

PR #196, opened 2017-10-05
Remove isDirty reference from README

slightlytyler:


jaredpalmer:
Thanks. Sorry about that!

PR #205, opened 2017-10-12
Add `transformValues` prop to allow auto-modification of form values

go1t:
For an ease of use, allow the user to pass `transformValues` prop into Formik, which will transform the values before they are validated and set to state. 

Example usage can be found in the test. Thank you in advance!


go1t:
will fix build issue and open the new one!

jaredpalmer:
Need to think about this. part of me thinks this is fine, another part of me thinks this will introduce pain and sadness and make input components less reusable because now there are multiple places where a value can be transformed.

go1t:
@jaredpalmer Sorry I don't fully understand why this would make input components less reusable, since `transformValues` will apply to all instances of `handleChange` and `setValues`, so it would affect all input components the same way. 

I opened a new PR at #206 (I thought I could delete this one but derp). Let's discuss more there :)

PR #206, opened 2017-10-12
Add `transformValues` prop to allow auto-modification of form values

go1t:
In the case where the user wants the form values to be modified before validating and updating, this prop can be helpful. The most recent use case I have encountered using this is the behavior of trimming input values and adding commas to multiple digit number. 

Obviously one may wrap the formik's `handleChange` prop and modify e.target.value, but that's tedious. I believe this prop would be useful :)

Thank you in advance for reviewing!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ I wrote Formik while building a large internal administrative dashboard with [Ia
 By now, you might be thinking, "Why didn't you just use [Redux-Form](https://github.com/erikras/redux-form)?" Good question.
 
  1. According to our prophet Dan Abramov, [**form state is inherently emphemeral and local**, so tracking it in Redux (or any kind of Flux library) is unnecessary](https://github.com/reactjs/redux/issues/1287#issuecomment-175351978)
- 2. Redux-Form calls your entire top-level Redux reducer multiple times ON EVERY SINGLE KEYSTROKE. This is fine for small apps, but as your Redux app grows, input latency will continue increase if you use Redux-Form. 
+ 2. Redux-Form calls your entire top-level Redux reducer multiple times ON EVERY SINGLE KEYSTROKE. This is fine for small apps, but as your Redux app grows, input latency will continue increase if you use Redux-Form.
  3. Redux-Form is 22.5 kB minified gzipped (Formik is 9.2 kB)  
 
 **My goal with Formik was to create a scalable, performant, form helper with a minimal API that does the really really annoying stuff, and leaves the rest up to you.**
@@ -256,6 +256,7 @@ npm install yup --save
     - [`enableReinitialize?: boolean`](#enablereinitialize-boolean)
     - [`isInitialValid?: boolean`](#isinitialvalid-boolean)
     - [`initialValues?: Values`](#initialvalues-values)
+    - [`transformValues?: Values`](#transformvalues-values-values--values)
     - [`onSubmit: (values: Values, formikBag: FormikBag) => void`](#onsubmit-values-values-formikbag-formikbag--void)
     - [`validate?: (values: Values) => FormikError<Values> | Promise<any>`](#validate-values-values--formikerrorvalues--promiseany)
     - [`validateOnBlur?: boolean`](#validateonblur-boolean)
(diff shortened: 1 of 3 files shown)


ezhlobo:
@jaredpalmer why did you close the issue? 

I saw your answer in the PR thread, so what do you think about adding a guide to a documentation of how to do that? Otherwise, I believe, people will create such issues in the future...

jaredpalmer:
So I've found that it's better to do this transformation in `setFieldValue` and have that be the single source of truth. The thinking here is that this colocates all the info you need about an input in one place. If we have a `transformValues` then there are 2 places that need to be updated when changes are made and 2 places that need to be refactored. Badness ensues.

jaredpalmer:
@ezhlobo Let's add this to the docs though.

sebald:
@jaredpalmer @ezhlobo An example would be nice. I am currently trying to wrap my head around this, but can't seem to get it to work. I used [this](https://gist.github.com/jaredpalmer/a8faaab12bc37e6a160a3c9549664f0b) to get started, but I can't unset a field error like you did in the gist.

I know this is not a support forum, so please just ignore me, if I am bothering you 🙂

This is what I currently have:

```ts
// types.ts
import { FieldProps } from 'formik';

export interface NumberFieldProps {
  name: string;
  label: React.ReactNode;
}

export interface NumberFieldInnerProps
  extends NumberFieldProps,
    FieldProps<any> {}
```

```ts
// NumerField.tsx

import { Field } from 'formik';
import { FormControl, FormHelperText } from 'material-ui/Form';
import TextField from 'material-ui/TextField';
import { NumberFieldInnerProps, NumberFieldProps } from './types';

// TODO: Move this to utils...
function isNumeric(val: any) {
  return !isNaN(val);
}

class NumberFieldInner extends React.Component<NumberFieldInnerProps> {
  transform = (value: string) => {
    const { form, field: { name } } = this.props;

    if (isNumeric(value)) {
      form.setFieldValue(name, value === '' ? value : +value);
      form.setFieldError(name, undefined as any);
    } else {
      form.setFieldError(name, 'Not a number');
    }
  };

  handleBlur = (e: React.ChangeEvent<any>) => {
    this.props.field.onBlur(e);
    this.transform(e.target.value);
  };

  handleChange = (e: React.ChangeEvent<any>) => {
    this.props.field.onChange(e);
    this.transform(e.target.value);
  };

  render() {
    const { label, field, form: { touched, errors } } = this.props;
    const error = Boolean(touched[field.name] && errors[field.name]);

    return (
      <FormControl error={error}>
        <TextField
          {...field}
          label={label}
          onBlur={this.handleBlur}
          onChange={this.handleChange}
          error={error}
          autoComplete="false"
        />
        {error && <FormHelperText>{errors[field.name]}</FormHelperText>}
      </FormControl>
    );
  }
}

export const NumberField: React.SFC<NumberFieldProps> = ({ name, label }) => (
  <Field name={name} label={label} component={NumberFieldInner} />
);
```

```ts
// usage example

import { Form, Formik, FormikProps } from 'formik';
import Button from 'material-ui/Button';

import { NumberField } from './';

const FormState: React.SFC<any> = props => (
  <div style={{ margin: '3rem 0' }}>
    <h3 style={{ fontFamily: 'monospace' }} />
    <pre
      style={{
        background: '#f6f8fa',
        fontSize: '1rem',
        color: 'black',
        padding: '.5rem',
      }}
    >
      <strong>props</strong> = {JSON.stringify(props, null, 2)}
    </pre>
  </div>
);

interface Values {
  num: number;
}

<Formik
  initialValues={{ num: 0 }}
    setTimeout(() => {
      action('onSubmit')(values);
      actions.setSubmitting(false);
    }, 2000);
  }}
  render={(props: FormikProps<Values>) => {
    const { isSubmitting, isValid } = props;
    return (
      <div>
        <Form>
          <NumberField name="num" label="Number" />
          <div style={{ margin: '30px 0' }} />
          <Button
            type="submit"
            raised
            color="primary"
            disabled={!isValid || isSubmitting}
          >
            Submit
          </Button>
        </Form>
        <FormState {...props} />
      </div>
    );
  }}
/>
```

Issue #202, opened 2017-10-11
Field name with object path


klis87:
Is it possible to achieve sth like this currently?
```javascript
<Formik
  initialValues={{ nested: { value: 'nested value' } }}
  validationSchema={schema}
  onSubmit={(values, { setSubmitting }) => {
    handleSubmit(values);
    setSubmitting(false);
  }}
>
  {() => (
    <Form>
      <Field name="nested.value" />
      <button type="submit">Save</button>
    </Form>
  )}
</Formik>
```
 I know I could use `mapPropsToValues` and `mapValuesToPayload`, but I find it really verbose and inconvenient for such a simple and common case. If I had deeply nested object, it would be even worse.

klis87:
I tried to solve it in my custom component:
```js
import React from 'react';
import { FormGroup, ControlLabel, FormControl, HelpBlock } from 'react-bootstrap';
import { get } from 'lodash';

const TextInput = ({ field, form: { touched, errors, values, setFieldValue }, ...props }) => (
  <FormGroup
    validationState={touched[field.name] && errors[field.name] ? 'error' : null}
  >
    <ControlLabel>{props.label}</ControlLabel>
    <FormControl
      type="text"
      {...field}
      {...props}
      value={get(values, field.name)}
      onChange={e => setFieldValue(field.name, e.target.value)}
    />
    {touched[field.name] && errors[field.name] && <HelpBlock>{errors[field.name]}</HelpBlock>}
  </FormGroup>
);
```
Please notice that I used `_.get` in `value` prop so reading is working for nested values. The problem is, that `setFieldValue('nested.value', value)` would add `nested.value` key to `values` object, instead of adjusting `values.nested.value`. The same problem I would have for `setFieldError`, `setFieldTouched` etc.

It seems that this library requires objects to be totally flat, but nested JSONs are really common. And flattening and "deflattening" in every form really kills productivity.

jaredpalmer:
Merged in #207

abelkbil:
@klis87 Thanks for sharing. How did you add bootstrap to formik. I dont even find any documentation online https://github.com/jaredpalmer/formik/issues/530

Can you please share. thanks you

klis87:
@abelkbil You can integrate Formik with whatever you desire, you need to create a component which you then should pass to `Field` `component` prop. See https://codesandbox.io/s/jRzE53pqR as an official example how to do it.

jswiton:
**First of all - thanks for your hard work!**

[...]

PR #207, opened 2017-10-12
Added nested values proposal prototype (#202)

klis87:
This is my quick suggestion to solve issue #202 . It is far from perfect (I don't use Typescript so I neglected types), also, I couldn't make lodash/fp working with Typescript, so I needed to use mutable `_.set`, and sadly I used it with `deepClone`. Of course this would need to be refactored, but this is just a prototype.

If you like this idea, we could work on it more, add missing tests etc.

The good news is, that all tests are passing so it is backward compatible change :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "gen-docs": "all-contributors generate && doctoc README.md"
   },
   "dependencies": {
-    "lodash": "4.17.4",
+    "dlv": "1.1.0",
     "lodash.isequal": "4.5.0",
     "prop-types": "^15.5.10",
     "warning": "^3.0.0"
@@ -46,7 +46,6 @@
   "devDependencies": {
     "@types/enzyme": "2.8.4",
     "@types/jest": "20.0.6",
-    "@types/lodash": "4.14.77",
     "@types/lodash.isequal": "4.5.2",
     "@types/node": "8.0.19",
     "@types/prop-types": "15.5.1",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "gen-docs": "all-contributors generate && doctoc README.md"
   },
   "dependencies": {
+    "lodash": "4.17.4",
     "lodash.isequal": "4.5.0",
     "prop-types": "^15.5.10",
     "warning": "^3.0.0"
@@ -45,6 +46,7 @@
   "devDependencies": {
     "@types/enzyme": "2.8.4",
     "@types/jest": "20.0.6",
+    "@types/lodash": "4.14.77",
     "@types/lodash.isequal": "4.5.2",
     "@types/node": "8.0.19",
     "@types/prop-types": "15.5.1",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "formik",
   "description": "Forms in React, without tears",
-  "version": "0.9.4",
+  "version": "0.10.5",
   "license": "MIT",
   "author": "Jared Palmer <jared@palmer.net>",
   "repository": "jaredpalmer/formik",
@@ -28,7 +28,7 @@
     "prebuild": "rimraf dist",
     "build": "tsc && cross-env NODE_ENV=production rollup -c && cross-env NODE_ENV=development rollup -c && rimraf compiled",
     "prepublish": "npm run build",
-    "format": "prettier --trailing-comma es5 --single-quote --write 'src/**/*' 'test/**/*'",
+    "format": "prettier --trailing-comma es5 --single-quote --write 'src/**/*' 'test/**/*' 'README.md'",
     "precommit": "lint-staged",
     "addc": "all-contributors add",
     "gen-docs": "all-contributors generate && doctoc README.md"
@@ -59,7 +59,7 @@
     "husky": "0.14.3",
     "jest": "20.0.4",
     "lint-staged": "4.0.2",
-    "prettier": "1.5.3",
+    "prettier": "^1.8.2",
     "react": "15.6.1",
     "react-dom": "15.6.1",
     "react-test-renderer": "15.6.1",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "gen-docs": "all-contributors generate && doctoc README.md"
   },
   "dependencies": {
-    "dlv": "1.1.0",
     "lodash.isequal": "4.5.0",
     "prop-types": "^15.5.10",
     "warning": "^3.0.0"
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -629,7 +629,9 @@ export function validateYupSchema<T>(data: T, schema: any): Promise<void> {
   return schema.validate(validateData, { abortEarly: false });
 }
 
-function setNestedObjectValues(object: any, value: any, response: any = {}) {
+function setNestedObjectValues(object: any, value: any, response: any = null) {
+  response = response === null ? {} : response;
+
   for (let k of Object.keys(object)) {
     const val = object[k];
     if (typeof val === 'object') {
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -1,14 +1,14 @@
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import isEqual from 'lodash.isequal';
-
 import {
   isFunction,
   isObject,
   isPromise,
   isReactNative,
+  isEmptyChildren,
   values,
-  setDeep,
+  setDeep
 } from './utils';
 
 import warning from 'warning';
@@ -22,17 +22,14 @@ export interface FormikValues {
 
 /**
  * An object containing error messages whose keys correspond to FormikValues.
+ * Should be always be and object of strings, but any is allowed to support i18n libraries.
  */
-export type FormikErrors = {
-  [field: string]: string;
-};
+export type FormikErrors<Values> = { [field in keyof Values]: any };
 
 /**
  * An object containing touched state of the form whose keys correspond to FormikValues.
  */
-export type FormikTouched = {
-  [field: string]: boolean;
-};
+export type FormikTouched<Values> = { [field in keyof Values]: boolean };
 
 /**
  * Formik state tree
@@ -40,15 +37,15 @@ export type FormikTouched = {
 export interface FormikState<Values> {
   /** Form values */
   values: Values;
-  /** 
-   * Top level error, in case you need it 
+  /**
+   * Top level error, in case you need it
    * @deprecated since 0.8.0
    */
   error?: any;
   /** map of field names to specific error for that field */
-  errors: FormikErrors;
+  errors: FormikErrors<Values>;
   /** map of field names to whether the field has been touched */
-  touched: FormikTouched;
+  touched: FormikTouched<Values>;
   /** whether the form is currently submitting */
   isSubmitting: boolean;
   /** Top level status state, in case you need it */
@@ -73,17 +70,17 @@ export interface FormikComputedProps<Values> {
 export interface FormikActions<Values> {
   /** Manually set top level status. */
   setStatus: (status?: any) => void;
-  /** 
-   * Manually set top level error 
+  /**
+   * Manually set top level error
    * @deprecated since 0.8.0
    */
   setError: (e: any) => void;
   /** Manually set errors object */
-  setErrors: (errors: FormikErrors) => void;
+  setErrors: (errors: FormikErrors<Values>) => void;
   /** Manually set isSubmitting */
   setSubmitting: (isSubmitting: boolean) => void;
   /** Manually set touched object */
-  setTouched: (touched: FormikTouched) => void;
+  setTouched: (touched: FormikTouched<Values>) => void;
   /** Manually set values object  */
   setValues: (values: Values) => void;
   /** Set value of form field directly */
@@ -99,7 +96,7 @@ export interface FormikActions<Values> {
 }
 
 /**
- * Formik form event handlers 
+ * Formik form event handlers
  */
 export interface FormikHandlers {
   /** Form submit handler */
@@ -131,42 +128,46 @@ export interface FormikSharedConfig {
 /**
  * <Formik /> props
  */
-export interface FormikConfig extends FormikSharedConfig {
-  /** 
+export interface FormikConfig<Values = object> extends FormikSharedConfig {
+  /**
    * Initial values of the form
    */
-  initialValues: object;
+  initialValues: Values;
 
-  /** 
-   * Submission handler 
+  /**
+   * Submission handler
    */
-  onSubmit: (values: object, formikActions: FormikActions<any>) => void;
+  onSubmit: (values: Values, formikActions: FormikActions<Values>) => void;
 
   /**
    * Form component to render
    */
-  component?: React.ComponentType<FormikProps<any> | void>;
+  component?: React.ComponentType<FormikProps<Values> | void>;
 
   /**
    * Render prop (works like React router's <Route render={props =>} />)
    */
-  render?: ((props: FormikProps<any>) => React.ReactNode);
+  render?: ((props: FormikProps<Values>) => React.ReactNode);
 
-  /** 
-   * A Yup Schema or a function that returns a Yup schema 
+  /**
+   * A Yup Schema or a function that returns a Yup schema
    */
   validationSchema?: any | (() => any);
 
-  /** 
-   * Validation function. Must return an error object or promise that 
+  /**
+   * Validation function. Must return an error object or promise that
    * throws an error object where that object keys map to corresponding value.
    */
-  validate?: ((values: any) => void | object | Promise<any>);
+  validate?: ((
+    values: Values
+  ) => void | object | Promise<FormikErrors<Values>>);
 
   /**
    * React children or child render callback
    */
-  children?: ((props: FormikProps<any>) => React.ReactNode) | React.ReactNode;
+  children?:
+    | ((props: FormikProps<Values>) => React.ReactNode)
+    | React.ReactNode;
 }
 
 /**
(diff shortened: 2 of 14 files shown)


jaredpalmer:
Here was my version ( very simiiliar). Mine uses developit's dlv utility, which does _.get in 128 Bytes. (see below) 

My concern with keeping this in `<Field>` and not in `<Formik>` is that `setFieldValue` would operate differently depending on where you call it. This is a code smell IMHO. 

```js
import React from 'react';
import PropTypes from 'prop-types';
import dlv from 'dlv';

function setDeep(path, value, obj) {
  let res = {};
  let i = 0;
  for (; i < path.length - 1; i++) {
    res = res[path[i]] || (res[path[i]] = (!i && obj[path[i]]) || {});
  }
  res[path[i]] = value;
  return res;
}

class Field extends React.Component {
  static contextTypes = {
    formik: PropTypes.object,
  };

  handleChange = e => this.setValue(this.props.name, e.target.value);

  handleBlur = e => this.setTouch(this.props.name, true);

  setValue = (key, value) => {
    let path = key.split('.');
    if (path.length > 1) {
      this.context.formik.setFieldValue(
        path[0],
        setDeep(path, value, this.context.formik.values)
      );
    } else {
      this.context.formik.setFieldValue(key, value);
    }
  };

  setTouch = (key, value) => {
    let path = key.split('.');
    if (path.length > 1) {
      this.context.formik.setFieldTouched(
        path[0],
        setDeep(path, value, this.context.formik.touched)
      );
    } else {
      this.context.formik.setFieldTouched(key, value);
    }
  };

  render() {
    const { component = 'input', name, ...props } = this.props;
    const field = {
      value:
        props.type === 'radio' || props.type === 'checkbox'
          ? props.value
          : dlv(this.context.formik.values, name),
      name,
      onChange: this.handleChange,
      onBlur: this.handleBlur,
    };
    const bag =
      typeof component === 'string'
        ? field
        : {
            field,
            form: {
              ...this.context.formik, 
              handleChange: this.handleChange,
              handleBlur: this.handleBlur,
              setFieldValue: this.setValue, 
              setFieldTouched: this.setTouch 
            }
          };
    return React.createElement(component, {
      ...props,
      ...bag,
    });
  }
}
```

klis87:
But I guess your approach is different, right? In my pull request `setFieldValue` itself is responsible for setting nested value, so `Field` component doesn't need to now about implementation detail. It is working in all cases: `setFieldValue('flatName', value)`, `setFieldValue('value.nested.nested2', value)` etc.

I agree that setters logic shouldn't be part of `Field` component, this should belong to `Formik` component, otherwise it won't be consistent. Not to mention that this won't cover all cases - like setting errors or setting dirty state for all inputs (in `touchAllFields`  function).  Also, people might want to implement own versions of `Field` component, or use some setter callback outside of it  - in my opinion `Field` should be treated like helper function/component, and nested path functionality should be really in the core of this library.

So to sum up, would you consider adding nested objects handling in this library? If yes, what are steps required to merge it? I can think of the following:
- adding missing tests to prove that setters understand path concept
- making sure that it would work with arrays
- optimizing setters performance
- updating docs of course

Btw, this form library is really perfect, I think that nested object is the only

package.json
```diff
@@ -34,6 +34,7 @@
     "gen-docs": "all-contributors generate && doctoc README.md"
   },
   "dependencies": {
+    "lodash": "4.17.4",
```
jaredpalmer: please use dlv as described. no need to import all of lodash 😄  !

package.json
```diff
@@ -45,6 +46,7 @@
   "devDependencies": {
     "@types/enzyme": "2.8.4",
     "@types/jest": "20.0.6",
+    "@types/lodash": "4.14.77",
```
jaredpalmer: remove

jaredpalmer:
To make things easier. Remove dlv and inline it along with setDeep:

```
// utils.ts
/** 
 * @private Deeply get a value from an object via it's dot path. 
 * See https://github.com/developit/dlv/blob/master/index.js
 */
export function dlv(obj: any, key: string | string[], def: any, p: number = 0) {
  key = (key as any).split ? (key as any).split('.') : key;
  while (obj && p < key.length) {
    obj = obj[key[p++]];
  }
  return obj === undefined ? def : obj;
}

/** 
 * @private Deeply set a value from in object via it's dot path. 
 * See https://github.com/developit/linkstate
 */
export function setDeep(obj: any, path: string[], value: any) {
  let res: any = {};
  let i = 0;
  for (; i < path.length - 1; i++) {
    res = res[path[i]] || (res[path[i]] = (!i && obj[path[i]]) || {});
  }
  res[path[i]] = value;
  return res;
}```

**Comment by klis87** [comment]
I removed lodash dependency. For deepSet I used your suggestion, but I needed to adjust it slightly as it should return the whole object, not only nested part of it. Also, path can contain brackets to set array items, so 'list.0' and 'list[0]' it the same - I needed to do it because `yupError.inner` paths are in this format. I admit it is a little messy and I think the best would be to cherry pick `lodash/fp/set`.

I tested nested values including array and now sth like `<Field name="list.0" />` is possible - there won't be any need to flatten arrays anymore - its is working with values, errors and touched.

**Comment by jaredpalmer** [comment]
Problem I see is that we are fully cloning and replacing all of values 2x on each keystroke now. I need to run this on my benchmarking suite to see perf implications.

**Comment by klis87** [comment]
Regarding your comment, like I mentioned above, please see my adjusted `setDeep`. Your version has 2 problems:
- it expects path as array and I believe splitting path like `x.y` into `['x', 'y']` should be really done inside, otherwise splitting will need to be repeated before every call of this function
- it returns incorrect value for nested paths - only last portion of object

Regarding perf implications, I am waiting for the results

**Comment by klis87** [comment]
I believe it should be fine, as we do only shallow clone by `{ ...obj, ...res }` in `setDeep`

**Comment by klis87** [comment]
I inlined `dlv`. Also, I modified slightly our `seetDeep` method - I replaced `(resVal[pathArray[i]] = (!i && obj[pathArray[i]]) || {});` with `(resVal[pathArray[i]] = (!i && { ...obj[pathArray[i]] }) || {});` to avoid obj mutations in some cases. Also, I added tests for seetDeep to be sure it returns correctly updated object and doesn't mutate input object.

**Comment by klis87** [comment]
I troubled me too much and now our setDeep supports arrays, yup doesn't complain anymore.

**Comment by gangstaJS** [comment]
When do you merge this pull request? I need this feature)

**Comment by gangstaJS** [comment]
Looks like `enableReinitialize` do not works correct with type of field `this.props.setFieldValue(`phone[0].value`, text)`. I will try create code example for this case

**Comment by klis87** [comment]
@gangstaJS  try `phone.0.value`. `dlv` getter doesn't anticipate brackets in path, only dots. `deepSet` is compatible with brackets only, because yup uses brackets for error paths. In my opinion, we should add brackets support do dlv as well - this is how _.get is working and what people generally expect.

**Comment by gangstaJS** [comment]
@klis87 As I see it replaced to dots `var pathArray = path.replace(/\]/g, '').split(/\.|\[/);` at `setDeep`. So `[]` does not matter

**Comment by klis87** [comment]
It is probably because you are pushing and removing array items. I tested arrays and it is working for me, but my array had a fixed length. I didn't try to add new items or remove them. I will play with this use case today and I will give you my results. If necessary, I will try to fix it.

**Comment by gangstaJS** [comment]
I have fixed problem. See my bullshit code)
```
if (this.props.enableReinitialize && !index$5(nextProps.initialValues, this.state.values)) {
            this.initialValues = __assign({}, this.state.values, nextProps.initialValues);
            this.resetForm(__assign({}, this.state.values, nextProps.initialValues));
        }
```

**Comment by klis87** [comment]
So it was bug in your app? Or we need to add something to Formik to handle this case?

**Comment by gangstaJS** [comment]
I think this is aa bug of formik. I have fixed it at the source code of formik

**Comment by klis87** [comment]
Ok, so again, I will try to prepare some demo for this use case later today.

**Comment by gangstaJS** [comment]
I am not very familiar with TS so I just edit complied code, sorry

**Comment by klis87** [comment]
No problem. I fell that this is not formik issue anyway, but they way you add items to your array. The demo will show us what to do

**Comment by klis87** [comment]
You mutate phone array - you made only shallow copy of current object.

Try this:
```js
case ADD_NEW_PHONE_ROW:
  return { 
    ...state,
    current: {
      ...state.current,
     phone: [...state.current.phone, {value: '', type: 'work'} ]
   }
};
```

gangstaJS:
@klis87 Adding is always working fine^ I have issue with removing

```js
case REMOVE_PHONE_ROW:
      state.current.phone.splice(action.index, 1);

      return {
        ...state,
        current: {
          ...state.current,
          phone: [...state.current.phone]
        }
      };
```

klis87:
I doubt it is the cause, but you should create phone clone before doing splice - otherwise you can cause unexpected behaviour, as you are mutating previous state:
```js
case REMOVE_PHONE_ROW:
      const phone = [...state.current.phone];
      phone.splice(action.index, 1);

      return {
        ...state,
        current: {
          ...state.current,
          phone
        }
      };
```

klis87:
@gangstaJS  So I assume everything is working now?

Btw. I created my prototype to make sure that everything is working:
```js
<Formik
  initialValues={{ list: ['a', 'b'] }}
  validationSchema={schema}
  onSubmit={(values) => {
    console.log(values);
  }}
>
  {({ values, setFieldValue }) => (
    <Form>
      {values.list.map((v, i) => (
        <div key={i}>
          <Field
            name={`list.${i}`}
            label={i}
          />
          <button
            type="button"
            onClick={() => {
              const n = [...values.list];
              n.splice(i, 1);
              setFieldValue('list', n);
              // we should update touched and potentially status state here
            }}
          >
            Remove {i}
          </button>
        </div>
      ))}
      <button
        type="button"
        onClick={() => {
          setFieldValue('list', [...values.list, '']);
        }}
      >
        add
      </button>
      <button type="submit" className="pull-right">Save</button>
    </Form>
  )}
</Formik>
```

Everything is working great, apart from touched state in some cases. For example, we have 2 items initially, we click add so we have 3, we focus and blur 3rd item so it is touched, we remove it, then add it again - item 3 will be set as touched immediately because when we removed a value, we didn't update touched object in the same time. We have 3 options:
1) Ignore this as it is not the end of the world
2) handle this in onClick in remove buttons
3) we could potentially think of a new Formik actions for pushing and removing items from array. It would reduce boilerplate needed to be done on app level in exchange for slightly bigger API.

What do you think @jaredpalmer ?

gangstaJS:
@klis87 yes it was my bad, ty

jaredpalmer:
Just an update. 

I really want to make a decision on this, but I need to setup a benchmark harness so we can perf test this programatically. Have an arm chair feeling that adding in a memoization flag may be required. I am also concerned about the size of the bundle increasing tbh.

stunaz:
What could we do to speed up the merge of this pull request?

jaredpalmer:
@stunaz i pushed up the benchmarking branch, you fork and copy and paste this in and begin perf testing.

stunaz:
> you fork and copy and paste this in 
You mean, to fork the benchmark branch and merge this pull request in it?

What are the steps/command to perf testing?

jaredpalmer:
`cd benchmark`
`yarn start` 
open another window
`yarn test`

This will start puppeteer (Chrome headless) and create a profile.json which you can open in devtools 
 performance tab to examine

Right now it's not very automated, you still need to swap out the component you want to benchmark in `src/index.js`

danielriley06:
@jaredpalmer Hey Jared! Really liking formik so far, only complaint is lack of the aforementioned nesting. We are in the middle of switching out all of our forms now, but wanted to check on if the merge above would be added and pushed soon or you're holding off. Would significantly impact how we are implementing.

Thanks!

mattleff:
@jaredpalmer Here's what I get for the performance test:

**master / 0.9.4**

image
image
image

**#207**

image
image
image

-----

Bundle sizes:

**master / 0.9.4**

JS: 102938
ES6: 102748
UMD: 103039
UMD minified: 28979

**#207**

JS: 103870 **+0.905%**
ES6: 103680 **+0.907%**
UMD: 103971 **+0.905%**
UMD minified: 29371 **+1.353%**

I'd humbly submit that the interest in this pull justifies a ~1% bundle increase.

VladShcherbin:
Had to move to redux-form just because of this feature yesterday. I'd be happy if this will be merged soon so I can switch back.

rzane:
FYI, if you're using the current version of formik, you can use this `<Field />` component to get nested values to work: https://gist.github.com/rzane/f98a8b961572e8dbb7a91e2132bfaf4e.

It's implemented in a different way than this PR solves the problem, but, the API should be the same.  Once this PR gets merged, you can just delete this file from your app and just use Formik's default `<Field />` component.

Magellol:
@VladShcherbin There are definitely ways to get nested data working with Formik with a small added cost.

VladShcherbin:
@Magellol in my project I need deeply nested dynamic forms with arrays. It's probably possible to make it work now, but the cost is too high.

While I like `formik` and don't like some parts of `redux-form` which I always try to avoid, I always come to `redux-form`. The api is extremely well thought and is pleasant to work with. The features it provides are great and, sadly, with such pace of `formik` development I don't expect to see them here anytime soon. 😥

src/formik.tsx
```diff
@@ -421,42 +420,23 @@ Formik cannot determine which value to update. For more info see https://github.
...
+      touched: setDeep(field, true, prevState.touched),
     }));
 
-    this.runValidationSchema(
-      {
-        ...this.state.values as object,
-        [field]: value,
-      } as object
-    );
+    this.runValidationSchema(setDeep(field, value, this.state.values));
```
ianks: setDeep should only be called once, and it could be used here

klis87:
@ianks Good catch. I also spotted this during refactoring to `setDeep`, but I left it as `handleChangeValue` is deprecated method, so it will be removed anyway.

klis87:
I couldn't wait for this any longer. If anyone needs nesting functionality, you can use [formik-nested](https://www.npmjs.com/package/formik-nested) . The downside is this is based on this PR - so on formik version 0.9.4, so please keep it in mind.

stunaz:
Same here, went back to redux-form

gangstaJS:
@klis87 Already did it

sexualmaniac:
After all hopes and excitement about this library we also had to move back to redux-form. Personally, I wish it had another author, the library has great potential, quality and community.

stunaz:
well this was released today as another alternative https://medium.com/@erikras/final-form-the-road-to-the-checkered-flag-cd9b75c25fe

sexualmaniac:
@stunaz thanks, will definitely take a look. It's even from @erikras, the creator of the best react forms library 🎉

sexualmaniac:
@jaredpalmer wow, you even removed the messages with [mentioned library](https://github.com/erikras/final-form) which will definitely be better than formik in all aspects. I wish I knew what person you are before even trying formik.

Such nice library with such an author, what a shame. Go write a rant tweet about it, that's all you can do.

jaredpalmer:
I want you all to understand that I have been thinking about this deeply over the last month. This is pain point in React, Redux, and MobX, not just Formik.

I have asked for advice from several "thought leaders" and yet there is no clear consensus. On the one hand, it's "easy" and magical. On the other, nested state should generally be avoided at all costs. Normalization is far more performant and easier to debug. Furthermore, the path update pattern isn't very "React-y." But to push Formik forward to 1.0, **I think a formal array solution is required.** So let's merge this on the `next` branch and move forward.

jaredpalmer:
@sexualmaniac attacking me personally is not cool. While I understand you are frustrated, you have violated the [Code of Conduct](https://github.com/jaredpalmer/formik/blob/master/CODE_OF_CONDUCT.md).

goodmind:
But you can't write types for something like this api, no?
```
<Field name={`list.${i}`} />
```

jaredpalmer:
dlv is the get method. I think we might need to adjust it

klis87:
@goodmind it should work. I used current version of Formik with dynamic arrays and it was fine (as long as you have an item with a given index in your list.

jaredpalmer:
@goodmind to get an index shouldn't it be: `<Field name={`list.thing[${i}]`} />`

klis87:
@jaredpalmer I think `dlv` doesnt respect brackets but I might be wrong. But it definitely works with dots.

jaredpalmer:
@klis87 thanks for this PR btw. I think we should add bracket support to dlv?

klis87:
@jaredpalmer Thanks for considering and merging this :) I think we should add this. Especially I needed to add bracket support for our nested setter, because `yup` creates paths for array items using brackets, not dots. Also, `lodash.get` supports brackets, and this is what people generally would expect working anyway.

klis87:
Btw, would you consider adding sth like [this from redux form](https://redux-form.com/7.2.0/docs/api/fieldarray.md/) ? Because without it is is hard to add/remove items from array - you need to do it like i did in my snippet, instead we could expose alternative actions to `setFieldValue`, like `push`, `pop`, `insert` etc.

jaredpalmer:
@klis87 I agree and think lodash's path syntax is the goal. As for array helpers, my plan is see how far I can get with a `<FieldArray render={({ sort, insert, pop, push, delete }) => }>` component to start.

jaredpalmer:
@klis Let's make a new issue for that as this has been merged

goodmind:
I meant TypeScript types

jaredpalmer:
Sorry, misread that. It's still a `string`, but typing Field properly is basically impossible anyways because there is no such thing as a JSX generic at the moment. So we punted as just have good old PropTypes at the moment.

zachlatta:
Hi, I've spent the past few hours trying to figure out how to use the changes from this pull request.

Can someone provide an example of how to create a form with nested values using the changes made here?

bs1180:
@zachlatta Have you installed the `next` branch? That caught me out - you need to run `yarn add formik@next`, and then it should work out of the box. The [example](https://github.com/jaredpalmer/formik/tree/next#basics) is also updated.

zachlatta:
Oh wow, thanks for that. I feel like a release should be cut if this issue is marked as closed.

Few things more confusing than running into an issue with a library, googling for it, finding a PR that got merged, and then trying the example from it and having it not work.

mrchief:
I tried `@next` but doing `setFieldValue(`${bag}.${id}`, { id, name }, false)` causes Chrome tab to crash.

mrchief:
Ok, this is not a Formik issue. My `id` is a number and it seems redux chokes on that. Without dev tools open, it works but is sluggish. With devtools open, it goes out of memory. Happens with redux-form too so it's definitely not Formik.

Skand17:
> ```js
>     return React.createElement(component, {
> ```

Thanks @jaredpalmer . This look good

PR #245, opened 2017-11-10
Fix #234. More expressive types on onSubmit

### LINKED ISSUE #234: onSubmit handler typing issue
Hey there 👋 

I've been using `formik` for quite some time but recently I've been using it with Typescript for the first time and I've came across some issue (I think) regarding the `onSubmit` handler typing.

https://github.com/jaredpalmer/formik/blob/b50e2ce19dcbe38bffcd12bfe8d5407be98e3022/src/formik.tsx#L138-L141

Because of `value: object`, I'm not able to know what's going to be the shape of my value object in TS. I've tried abstracting _a_ type  such as:

```typescript
type FormikSubmitHandler<V> = (value: V, actions: FormikActions<V>) => void;
```

Although, once done that, I get a signature mismatch when I pass the function to formik. I'm not sure what's the right fix atm but to get TS (and autocomplete) not complaining anymore I had a workaround by doing a type assertion. Here's a complete example that illustrates the issue:

```typescript
type FormikSubmitHandler<V> = (value: object, actions: FormikActions<V>) => void;

interface FormValues {
  foo: string;
  bar: number
}

// At this point, I don't know what's in `value` since it's type is `object`.
const myHandler: FormikSubmitHandler<FormValues> = (value, actions) => {
  const typedValue = value as FormValues;
  value.foo; // string
  value.bar; // number
}

render() {
  return <Formik onSubmit={myHandler} />
}
```

I'd be happy to provide a PR but I'm not sure what's to do to get it fixed (if needs be).

Thanks.

jaredpalmer:
- Formik will now work with TS strict mode
- and whenever DefinitelyTyped React/ReactDOM gets its shit together on TS 2.6.1

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -129,26 +129,26 @@ export interface FormikSharedConfig {
 /**
  * <Formik /> props
  */
-export interface FormikConfig extends FormikSharedConfig {
+export interface FormikConfig<Values = object> extends FormikSharedConfig {
   /**
    * Initial values of the form
    */
-  initialValues: object;
+  initialValues: Values;
 
   /**
    * Submission handler
    */
-  onSubmit: (values: object, formikActions: FormikActions<any>) => void;
+  onSubmit: (values: Values, formikActions: FormikActions<Values>) => void;
 
   /**
    * Form component to render
    */
-  component?: React.ComponentType<FormikProps<any> | void>;
+  component?: React.ComponentType<FormikProps<Values> | void>;
 
   /**
    * Render prop (works like React router's <Route render={props =>} />)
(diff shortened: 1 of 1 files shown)


src/formik.tsx
```diff
@@ -129,26 +129,26 @@ export interface FormikSharedConfig {
 /**
  * <Formik /> props
  */
-export interface FormikConfig extends FormikSharedConfig {
+export interface FormikConfig<Values = object> extends FormikSharedConfig {
```
jaredpalmer: I think this can just be `object`, might need to be `any`.

PR #256, opened 2017-11-15
Fix #255. Remove invalid warning in Field

### LINKED ISSUE #255: Warning: You should not use a non-string <Field component> and <Field children> in the same <Field> component; <Field component> will be ignored.
Using a custom Select field gives me the warning ` Warning: You should not use a non-string <Field component> and <Field children> in the same <Field> component; <Field component> will be ignored.`.

I dont' see how I can use a custom Select component on a Field, not be a string and also have children, I think this warning isn't valid.

This is the code that generates the warning.
```jsx
<Field
  component={FormikSelectField}
  name="availability"
  floatingLabelText="Availability"
  fullWidth
>
  {Object.keys(USER_AVAILABILITY_LABELS).map(availability => (
    <MenuItem
      key={availability}
      value={availability}
      primaryText={USER_AVAILABILITY_LABELS[availability]}
    />
  ))}
</Field>
```

```jsx
// FormikSelectField.js
import React from 'react';
import SelectField from 'material-ui/SelectField';

const FormikSelectField = ({
  field: { name, value },
  form: { touched, errors, setFieldValue, setFieldTouched },
  onChange,
  ...otherProps
}) => {
  const handleChange = (event, key, data) => {
    setFieldTouched(name, true);
    setFieldValue(name, data);

    if (onChange) {
      onChange(event, key, data);
    }
  };

  return (
    <SelectField
      errorText={touched[name] && errors[name]}
      value={value}
      onChange={handleChange}
      {...otherProps}
    />
  );
};

export default FormikSelectField;
```

bogdansoare:
Removes invalid warning on the `<Field>` in cases when you have a custom component and some children (Selects)

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -105,15 +105,6 @@ export class Field<Props extends FieldAttributes = any> extends React.Component<
 
     warning(
       !(
-        typeof this.props.component !== 'string' &&
-        this.props.children &&
-        !isEmptyChildren(this.props.children)
-      ),
-      'You should not use a non-string <Field component> and <Field children> in the same <Field> component; <Field component> will be ignored.'
-    );
-
-    warning(
-      !(
         this.props.render &&
         this.props.children &&
         !isEmptyChildren(this.props.children)


prichodko:
I think warning should be modified to rather check if `children` is not a function. Because then the component would be ignored.

prichodko:
Also when making changes to the code, don't forget to update tests 👍 https://github.com/jaredpalmer/formik/blob/master/test/Field.test.tsx#L171

prichodko:
@jaredpalmer do you agree that warning should be modified? If yes @bogdansoare are you still up for this PR? Otherwise I can fix this now, since it could confuse other people.

jaredpalmer:
Yes

bogdansoare:
@prichodko thanks for noticing it, your solution would be better like @jaredpalmer said.
Sorry for the delayed response.

PR #260, opened 2017-11-17
Added typescript usage example

kingdaro:
Usage with TypeScript wasn't immediately obvious to me, so I figured it'd be helpful to include an example of how to do so while getting proper typechecks when rendering the component. Credits to [this blog post](https://blog.mojotech.com/typescript-generic-react-components/), from which I learned about the `class extends` bit.


stunaz:
would be great to add an example using **typescript** and the **hoc** `withFormik`

kingdaro:
@stunaz Done! It's adapted from one of the usages in the docs, but slimmed down a little. Still seems a little lengthy, though. Maybe it could be shortened a little more?

jaredpalmer:
FYI I disagree with that post, you don’t want to actually extend React.Conponent

kingdaro:
I'd normally agree, but I don't know of a better way to create react classes with custom generics. TypeScript doesn't seem to be able to infer the Formik generics just from usage within JSX, so it seems like a good enough approach for now.

Havret:
I tried to use your example but it does not compile. I'm using Typescript 2.6.1.

I've got such error: 
`Error:(9, 32) TS2344: Type 'FormikConfig<LoginFormValues>' does not satisfy the constraint 'FormikConfig<object>'.
  Types of property 'onSubmit' are incompatible.
    Type '(values: LoginFormValues, formikActions: FormikActions<LoginFormValues>) => void' is not assignable to type '(values: object, formikActions: FormikActions<object>) => void'.
      Types of parameters 'values' and 'values' are incompatible.
        Type 'object' is not assignable to type 'LoginFormValues'.`

PR #261, opened 2017-11-18
Improve Field warnings

### LINKED ISSUE #255: Warning: You should not use a non-string <Field component> and <Field children> in the same <Field> component; <Field component> will be ignored.

prichodko:
I tried to improve warnings to cover all invalid cases.

First of all I removed `component` default prop and moved it to the `render()` as default argument. This way can check for `component`'s existence rather if it isn't a string.

This passed checks before:

```javascript
<Field
  name="text"
  component="select"
  render={() => <input type="text" />}
/>
```

Now it warns user that `You should not use <Field component> and <Field render> in the same <Field> component; <Field component> will be ignored`. It of course still warns user, if React component is passed and it works if nothing is specified.

Same condition is also applied for the check with `children`, but here I also changed condition to check if children is a function to fix #255.

**PROPOSAL**
`warning` logic is inconsistent between `Field` and [formik](https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L271). Maybe we should make it so?

**Qs**
In React documentation, `createElement`'s states to pass children as a third argument - https://reactjs.org/docs/react-api.html#createelement, does it make any difference?

When running `test` script a get bunch of `Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs`, is it happening to you too?

Fixes #255 
Closes #256

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -2,7 +2,7 @@ import * as PropTypes from 'prop-types';
 import * as React from 'react';
 
 import { FormikProps } from './formik';
-import { isEmptyChildren } from './utils';
+import { isFunction, isEmptyChildren } from './utils';
 import warning from 'warning';
 
 export type GenericFieldHTMLAttributes =
@@ -86,10 +86,6 @@ export class Field<Props extends FieldAttributes = any> extends React.Component<
     formik: PropTypes.object,
   };
 
-  static defaultProps = {
-    component: 'input',
-  };
-
   static propTypes = {
     name: PropTypes.string.isRequired,
     component: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Smart!

prichodko:
Thanks! What do you think about unifying logic I wrote in proposal?

jackkainov:
Hi guys, this change was release as a patch version in 0.10.5. Was there a reason to do so? It affects snapshots if you use jest or any snapshot generator. Would be at lease a minor version better?
Thanks

PR #276, opened 2017-11-28
Next

jaredpalmer:
- [x] Sync field-level validations + Docs
- [x] Async field-level validations + Docs . 
- [x] Document nested path syntax
- [x] Nested path syntax for deep updates (from #207) 
- [x] `<FieldArray>` (for now this is bundled with core), might move to `formik-array` pkg
   - push
   - pop
   - remove
   - insert
   - move
   - swap
   - unshift
- [x] Document `<FieldArray>`
- [x] Upgrade to TS 2.6.2
- [x] Upgrade to React 16, Jest 21, Enzyme 3
- [x] Typechecking in test files during development


LinusU:
I think field level validations seems awesome, I'm going to try it out on a form which adds fields dynamically which have been a huge pain currently for me 🍻

src/Field.tsx
```diff
@@ -113,19 +119,58 @@ export class Field<Props extends FieldAttributes = any> extends React.Component<
...
+    const maybePromise = (validate as any)(value);
+    // Check if validate it returns a Promise
+    if (isPromise(maybePromise)) {
+      (maybePromise as Promise<any>).then(
+        () => setFieldError(name, undefined),
+        error => setFieldError(name, error)
+      );
+    } else {
+      // Otherwise set the error
+      setFieldError(maybePromise);
```
LinusU: Shouldn't this be `setFieldError(name, maybePromise)`? I think that this is causing an error for me...

LinusU: Yeah, adding `name` here fixes this for me. The debugger also told me that `setDeep` was being called with my error message as the path so this seems to be the culprit. I opened #290 to fix it

jaredpalmer:
Will push this when tests pass

LinusU:
Thank you!

(tests seems to be passing now 😃)

LinusU:
So, one thing to think about, which I'm not sure really how to solve, is that setting a validator on field will not rerun that validation if `setFieldValue` is used somewhere else 🤔

jaredpalmer:
True true. Hmmmmm

brikou:
@jaredpalmer this looks great!

brikou:
@jaredpalmer it looks like when dealing nested array the required name should look like `my_array.0.my_obj.my_prop`. Yup validation uses instead `my_array[0].my_obj.my_prop` in default errors. I you follow this convention then yup validation would work out of the box, WDYT?

jaredpalmer:
Yeah we need to add bracket support to dlv

src/FieldArray.tsx
```diff
@@ -0,0 +1,103 @@
...
+      pop: this.pop,
+      swap: this.swap,
+      move: this.move,
+      insert: this.insert,
+      unshift: this.unshift,
+      remove: this.remove,
+    };
+    const { render } = this.props;
+    return render
+      ? (render as any)({ ...arrayHelpers, form: this.context.formik })
```
brikou: form is missing in ts definition

jaredpalmer: ahh yes good looks

brikou: As you fix missing def, can you release a new alpha, as I'm living on the edge :D? Thank you so much!

brikou: @jaredpalmer thanks :+1:

src/formik.tsx
```diff
@@ -639,7 +619,7 @@ function warnAboutMissingIdentifier({
  * Transform Yup ValidationError to a more usable object
  */
 export function yupToFormErrors<Values>(yupError: any): FormikErrors<Values> {
```
brikou: yup is now definitively typed :) ... take a look https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yup/index.d.ts

jaredpalmer: Yah will do

renchap:
I just implemented an custom password input field with built-in strenght check and validation using the field-level validation and it works amazingly well, great work!

Have you considered the interaction between those field-level validations and the `validationSchema` prop? Currently you can not use both at the same time as errors are reset in `validateYupSchema` (https://github.com/jaredpalmer/formik/blob/next/src/formik.tsx#L343).
I think it would be better to merge errors from the fields and the one from the schema, but I am not sure on how this could work. Any ideas?

renchap:
To add more details to my previous comment:
I have a `validationSchema` to ensure that `email` is required and a field-level validation on the `password` field.

When I add calls to `console.log` in my form `render` function, I see that Formik's `errors` is changed twice when I make a change to a field:
- First update: `{email: "email is a required field", password: "Password too weak"}`
- Second update (immediatly after): `{email: "email is a required field"}`

After some debugging, it seems related to `runValidationSchema`. When there are no schema errors, the errors are reset: https://github.com/jaredpalmer/formik/blob/560a5e3ba87f5cd181fab0ab5063f2944e4f5272/src/formik.tsx#L341-L343

When there are schema errors, the errors are replaced by the one from the schema validation:
https://github.com/jaredpalmer/formik/blob/560a5e3ba87f5cd181fab0ab5063f2944e4f5272/src/formik.tsx#L348-L350

They should be merged with errors coming from the fields (and eventually overriding them, depending on which one you want to prioritize).

renchap:
And a related error: 
Starting from a `state.errors = { password: "invalid" }`, calling `setFieldError('password', undefined)` results in `state.errors = { password: undefined }`.
A way to fix it would be to call `delete` on the property in `setDeep` (after cloning the object to not mutate the argument).

Please tell me if you prefer separate issues about these reports.

bogdansoare:
Merging the errors from field-level validations and validationSchema would be really useful

renchap:
I digged a bit more into this, and the real issue is merging validation results when some (or all) of them are async (either from `validationSchema`, a validation function, or field-level validations).
I do not really know how to fix this properly, as you need to know when to merge (the new validation results that come in needs to be applied on top of the existing ones). Probably something like:
- when we need to validate the form, reset the validation results, generate a "validation pass id"
- when an async validation resolves, if its for the current "validation pass id", merge it into the existing results. Otherwise ignore it.

This is needed as we can have a new validation pass happening while some async results did not resolve yet (like if coming from a slow server-side process). As there is no way to cancel promises, we need to check if the results are current when it resolves.

I might have missed something here, I am still trying to grasp the whole codebase.

jaredpalmer:
Okay folks, going to merge and cut a release.

LinusU:
Great work 🎂

PR #290, opened 2017-12-05
Add missing name in call to setFieldError

LinusU:
(if this could be released as 0.11.0-alpha.2 asap I would be sooooo happy ❤️)


jaredpalmer:
thanks

PR #291, opened 2017-12-05
Add warning if validationSchema is missing form fields

davorbadrov:
Hello,

I really like your project and while using it I've run into a bug when using Yup with Formik.

TL;DR:

This PR adds a warning when some fields are specified through props.initialValues but are missing from the validationSchema.

In long:

An uncaught error is thrown when using Yup's object validation.

**Versions:**

formik@0.10.5
yup@0.23.0

If you don't specify all the form fields/initial values inside of the Yup object and
use Yup.when then a uncaught error is thrown and it crashes your app.

To recreate the bug all that's needed is:

```js
// validation schema
Yup.object().shape({
    // email: Yup.string().email(), // this should be commented out for the bug to surface
    username: Yup.string().when('email', {
        is: 'test@email.com',
        then: Yup.string().min(2),
        otherwise: Yup.string().min(6)
    })
})

// initial fields / values
{
    username: '',
    email: ''
}
```

You can see the bug on [Codesandbox example](https://codesandbox.io/s/rypmz212rp)

It took me a few hours to track down the bug, the error it throws "Uncaught (in promise) TypeError: Cannot read property 'length' of undefined" doesn't give much info.

I was thinking of adding a warning if you have some fields in the validation schema missing but they are present in the initialValues. This will probably save someone the headache I already had.

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -8,7 +8,8 @@ import {
   isReactNative,
   isEmptyChildren,
   values,
-  setDeep
+  setDeep,
+  warnIfValidationSchemaHasMissingFields,
 } from './utils';
 
 import warning from 'warning';
@@ -255,6 +256,7 @@ export class Formik<
     };
 
     this.initialValues = props.initialValues || ({} as any);
+    warnIfValidationSchemaHasMissingFields(this.props);
   }
 
   componentWillReceiveProps(nextProps: Props) {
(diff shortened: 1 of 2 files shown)


jaredpalmer:
So we have a LOT of forms on which we only validate part of the values for various reasons. Need to think about this.

jaredpalmer:
Can I close this?

davorbadrov:
Sure.

PR #334, opened 2018-01-09
Allow bracket path support

### LINKED ISSUE #307: [Next] Add bracket path support
## Current Behavior

`name="social.0.facebook"`

## Expected Behavior

`name="social[0]facebook"`

## Relevant code

`setDeep` and `dlv` in `src/utils.ts` probably need to be updated

- lodash's `stringToPath`: https://github.com/lodash/lodash/blob/b2ea6b1cd251796dcb5f9700c4911a7b6223920b/.internal/stringToPath.js

skvale:
Fixes #307

### DIFF
--- a/README.md
+++ b/README.md
@@ -1283,7 +1283,7 @@ slightly relaxed and allow you to return a `Function` (e.g. `i18n('invalid')`).
 
 ### `<FieldArray/>`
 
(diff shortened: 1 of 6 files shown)


jaredpalmer:
Nice I can review when I get back from CES in a few hours.

jaredpalmer:
@skvale do you think fieldarray should be a separate package?

skvale:
I don't think I fully understand why fieldarray would or wouldn't be a separate package and how it relates.

PR #337, opened 2018-01-09
Fix unshift in FieldArray

existentialism:
Fixed a small bug with FieldArray's unshift (it was returning the original array), and added tests.


jaredpalmer:
Good looks

PR #362, opened 2018-01-22
do not touch array on FieldArray.changeValue

timc13:
Touching the array field changes the structure of `touched` to not match `values` anymore. I ran into this bug when adding new objects into my array.

Given `values`:
```
{
  items: [
    { foo: 'foo0', bar: 'bar0' },
    { foo: 'foo1', bar: 'bar1' },
  ]
}
```
I expect `touched` to look something like:
```
{
  items: [
    { foo: false, bar: false },
    { foo: false, bar: false },
  ]
}
```

Instead, by touching the array field on every `FieldArray.changeValue`, the current behavior turns touched into this: 
```
{
  items: true
}
```

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -38,11 +38,10 @@ export class FieldArray extends React.Component<
   };
 
   changeValue = (fn: Function) => {
-    const { setFieldValue, setFieldTouched, values } = this.context.formik;
+    const { setFieldValue, values } = this.context.formik;
     const { name } = this.props;
     const val = fn(dlv(values, name));
     setFieldValue(name, val);
-    setFieldTouched(name, true);
   };
 
   push = (value: any) => this.changeValue((array: any[]) => [...array, value]);


timc13:
I see the problem with this PR.  Once an array field has changed, validations are run immediately - and for that to be clear, we need to touch the fields.

I actually think running ArrayHelper actions should NOT trigger validations.  What's the argument for validating upon changing the array items?

jaredpalmer:
Imagine a form with validation where an array needs to be at least n items long.

timc13:
I see your point, but I think for consistency, we should validate & touch arrays deeply in the same way that we do for nested objects.  I don't think mutating a nested objects sets their touched value to true, does it? - or maybe there is no way to add/remove keys to the object in a "form" way.

It seems that for MVP purposes, it's simply more usable to control the length/min/max of an array via access to the array helper methods.  Currently, `FieldArray` is useless because anytime you touch an array item field, `dlv` crashes.

I do agree though, that we probably want a way to track/validate the array itself, as well as track/validate the items inside.

timc13:
done in https://github.com/jaredpalmer/formik/commit/b32ebfd3f294996f88634bdaa9574f158274bb49

PR #370, opened 2018-01-24
Change layout of generics in main Formik class to allow better inference

### LINKED ISSUE #314: FormikConfig has a generic type of an empty object
I'm running into the following issue using Formik 0.11.0-beta.1, TypeScript 2.6.2, and `strict: true` in `tsconfig.json`. I can't see why `FormikConfig` has a generic type of `{}`.

```ts
interface Values {
  subject: string;
}

<Formik
  initialValues={{ subject: '' }}
  onSubmit={(values: Values) => alert(JSON.stringify(values))}
  // ...
/>
```

```bash
Type '{ initialValues: { subject: string; }; onSubmit: (values: Values) => void; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Formik<FormikConfig<{}>, FormikValues>> & Readonly...'.
  Type '{ initialValues: { subject: string; }; onSubmit: (values: Values) => void; }' is not assignable to type 'Readonly<FormikConfig<{}>>'.
    Types of property 'onSubmit' are incompatible.
      Type '(values: Values) => void' is not assignable to type '(values: {}) => void'.
        Types of parameters 'values' and 'values' are incompatible.
          Type '{}' is not assignable to type 'Values'.
            Property 'subject' is missing in type '{}'.
```

weswigham:
The generics as used before were not inferrable, so provided no security for users outside the class. (meaning they could annotate conflicting types for `initialValues` and an `onChange` handler, and not much would change).

To elaborate, previously formik's constructor had a signature similar to 
```ts
new <Props extends FormikConfig<Values> = FormikConfig<Values>, Values = object>(props: Props): React.Component<Props> & { initialValues: Values, onChange?: (props: Values) => void }
```
When TS does inference on such a signature (as we will now do for JSX as of TS 2.7; it's been missing from JSX for awhile), and you pass in an argument type like `{ initialValues: { x: 0 }, onChange: p => undefined }` inference works in this way: We infer the positional arguments between the callsite and the original signature pairwise, so we infer from `{ initialValues: { x: 0 }, onChange: p => undefined }` to `Props`. `Props` is a type parameter, so we assign it the source type as the inference and move on to other parameters. There are no other paramaters, and `Values` has yet to be assigned an inference, so it is assigned its default, `object`. The lack of a real inference on `Values` causes us not to establish a link between the contextual type on the argument position of the callback and the value of the `initialValues` member of the argument type, so it just gets the type of the default in the end, too.

By changing the signature to (effectively) 
```ts
new <ExtraProps = {}, Values = object>(props: FormikConfig<Values> & ExtraProps): React.Component<FormikConfig<Values> & ExtraProps> & { initialValues: Values, onChange?: (props: Values) => void }
```
we restructure the parameter such that it has type has positions from which to infer both `Props` (by redefining it as the combination of `FormikConfig<Values>` and some set of extra props) and `Values` (through its' presence in the type `FormikConfig<Values>`).

This should fix #314 when paired with [this fix](https://github.com/Microsoft/TypeScript/pull/21383) on the TS side (which is fixing a bug relating to us doing something incorrectly with respect to assigning contextual types on jsx attributes).

The changes to the `console.error` lines are because `prettier` reformatted them on commit. I assume that means that the change on them is desirable, if unrelated.

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -179,10 +179,10 @@ export type FormikProps<Values> = FormikState<Values> &
   FormikHandlers &
   FormikComputedProps<Values>;
 
-export class Formik<
-  Props extends FormikConfig<Values> = FormikConfig<Values>,
-  Values = object
-> extends React.Component<Props, FormikState<any>> {
+export class Formik<ExtraProps = {}, Values = object> extends React.Component<
+  FormikConfig<Values> & ExtraProps,
+  FormikState<any>
+> {
   static defaultProps = {
     validateOnChange: true,
     validateOnBlur: true,
@@ -221,7 +221,7 @@ export class Formik<
           ? this.state.errors && Object.keys(this.state.errors).length === 0
           : this.props.isInitialValid !== false &&
             isFunction(this.props.isInitialValid)
-            ? (this.props.isInitialValid as (props: Props) => boolean)(
+            ? (this.props.isInitialValid as (props: this['props']) => boolean)(
                 this.props
               )
             : (this.props.isInitialValid as boolean),
@@ -245,7 +245,7 @@ export class Formik<
     };
   }
 
-  constructor(props: Props) {
+  constructor(props: FormikConfig<Values> & ExtraProps) {
     super(props);
     this.state = {
       values: props.initialValues || ({} as any),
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Thank you for this. Is their any way to issue a patch that will make Formik work with 2.6.2?

jaredpalmer:
Need to revert. Will add this change back to next branch shortly

PR #373, opened 2018-01-24
Fix setDeep mutations with nested values

### LINKED ISSUE #364: setDeep changes two-level deep objects in-place
## Bug

React state is changed in-place if you change deeply (more than 2 levels) nested values.

## Current Behavior

A call like this:

    const obj = { x: 'y', twofoldly: { nested: { value: 'a' } } };
    setDeep('twofoldly.nested.value', 'b', obj)`

changes the object at `obj.twofoldly.nested` in-place. This means the react state is changed in-place if you have a field with a name like `twofoldly.nested.value`.

In my case I have an equality check in `componentWillReceiveProps` of the form component with `isEqual(nextProps.values, this.props.values)` which now always returns `true`.

Add a test case to `test/utils.test.tsx` like this

    it('updates twofoldly nested value', () => {
      const obj = { x: 'y', twofoldly: { nested: { value: 'a' } } };
      const newObj = setDeep('twofoldly.nested.value', 'b', obj);
      expect(obj.twofoldly.nested === newObj.twofoldly.nested).toEqual(false); // fails, same object still
      expect(obj).toEqual({ x: 'y', twofoldly: { nested: { value: 'a' } } }); // fails, it's b here, too
      expect(newObj).toEqual({ x: 'y', twofoldly: { nested: { value: 'b' } } }); // works ofc
    });

to see the problem.

(CCing @klis87, thank you for adding the nesting feature ☺️🎉  it is really helpful)

## Desired Behavior

There should be a new object in the nested state instead of the old nested object with a changed value. The above unit test should pass.

## Suggested Solutions

Create a proper copy of the objects.

## Info
 
- Formik Version: v0.11.0-beta.1 (`setDeep` was added in 3a8e620dc56f0ac14dd9796795fd5da17853fb40 / #207)
- OS: macOS 10.13.2
- Node Version: v8.1.4
- Package Manager and version: yarn 1.3.2

ecoleman:
Fixes #364

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "gen-docs": "all-contributors generate && doctoc README.md"
   },
   "dependencies": {
+    "lodash.clonedeep": "^4.5.0",
     "lodash.isequal": "4.5.0",
     "lodash.topath": "4.5.2",
     "prop-types": "^15.5.10",
@@ -47,6 +48,7 @@
     "@types/enzyme": "3.1.5",
     "@types/enzyme-adapter-react-16": "1.0.1",
     "@types/jest": "20.0.6",
+    "@types/lodash.clonedeep": "^4.5.3",
     "@types/lodash.isequal": "4.5.2",
     "@types/lodash.topath": "4.5.3",
     "@types/node": "8.0.19",
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import toPath from 'lodash.topath';
+import cloneDeep from 'lodash.clonedeep';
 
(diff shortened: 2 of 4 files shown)


jaredpalmer:
How big is the umd bundle on this?

ecoleman:
```
fix-set-deep-mutation: Bundle size: 45.01 KB, Gzipped size: 12.66 KB
next: Bundle size: 36.7 KB, Gzipped size: 11.04 KB
```

ecoleman:
If you'd prefer, I can submit an alternative solution extending the original functionality. 

I just noticed the couple lodash dependencies, and assumed it was acceptable.  :)

PR #379, opened 2018-01-26
Added react native types, fixed handleSubmit in rn/typescript

### LINKED ISSUE #376: TypeError when calling handleSubmit in react-native
## Current Behavior
The recommended usage of formik in react native is: `<Button onPress={props.handleSubmit} />`. However, when using typescript, the type of event that `handleSubmit` accepts is different than what react-native's `onPress` generates.

## Desired Behavior
Formik's `handleSubmit` should take a `GestureResponderEvent` in addition to `React.FormEvent<HTMLFormElement>`.

## Suggested Solutions
This branch fixes this issue: https://github.com/Shurelia/formik/tree/handleSubmit-fix
Unfortunately, there is a typings conflict between `@types/node` and `@types/react-native` which prevents the build. Since the only type used from node was `process`, I've manually declared that as any and removed `@types/node`. Not an elegant solution, and I'd be open to working on it more if the maintainers are okay with removing `@types/node` from the project (and perhaps manually declaring `process`?).

## Info
- Formik Version: 0.10.5
- OS: Windows 10
- Node Version: 9.3.0
- Package Manager and version: yarn 1.3.2
- Typescript: 2.6.2
- react: 16.2.0
- react-native: 0.51.0

Shurelia:
Fixes #376


jaredpalmer:
Going to merge and fix on master

Issue #361, opened 2018-01-22
Re-run validations on `resetForm`


rpowelll:
## Bug, Feature, or Question?

When `resetForm` is called, it should be possible to also re-run the form's validations. Whether this should be an optional argument to `resetForm`, or some automatic behaviour tied to `isInitialValid` I'm not sure.

## Current Behavior

Upon resetting my form, there's no way to run validations against the new values./

## Desired Behavior

I can run validations against the form's new values.

## Suggested Solutions

There's a few possible solutions:

- Allow `resetForm` to take an optional second argument to indicate whether to run validations
- Expose a `performValidation` function or something similar as an escape hatch
- If `isInitialValid` is false, resetting the form should should also cause validations to be re-run

## Info
 

- Formik Version: 0.11.0.beta.1
- OS: macOS 10.13.2
- Node Version: 9.3.0
- Package Manager and version: npm 5.6.0

jaredpalmer:
I think this make sense. Fn should be called validate() and exposed via props.

jaredpalmer:
This has been shipped in v0.11. You can now call `validateForm()`

phuongthuan:
@jaredpalmer 
I'm caring about this issue.
Could you please give me some example? 

I want to reset validation after form was submitted.

```
onSubmit={async (values, { setSubmitting, setStatus, setErrors, resetForm }) => {
            const { title, achievement, plan, description, comment } = values;

            let issues = [];
            if (isEmpty(values.issues)) {
              issues = []
            } else {
              issues = values.issues.map(issue => ({ id: issue }));
            }

            try {
              await createDailyReport({
                variables: {
                  title,
                  achievement,
                  issues,
                  plan,
                  description,
                  comment
                }
              });

              resetForm({});
              setStatus({success: true})
            } catch (error) {
              setStatus({success: false});
              setSubmitting(false);
              setErrors({submit: error.message})
            }
          }}
```

phuongthuan:
@jaredpalmer This is exactly the issue that I'm facing.
My code like this:

```

         onSubmit={async (values, { setSubmitting, setStatus, setErrors, resetForm }) => {
            const { title, achievement, plan, description, comment } = values;

            let issues = [];
            if (isEmpty(values.issues)) {
              issues = []
            } else {
              issues = values.issues.map(issue => ({ id: issue }));
            }

            try {
              await createDailyReport({
                variables: {
                  title,
                  achievement,
                  issues,
                  plan,
                  description,
[...]

PR #384, opened 2018-01-29
Validate helper

rpowelll:
As discussed in #361, this PR adds a `validateForm` function to the props that Formik provides to wrapped components. Note that I've named it `validateForm` rather than just `validate` for consistency with `resetForm`.


jaredpalmer:
Let's close and reopen on the next branch. I am locking down master until we move to next.

jaredpalmer:
I missed this PR when trying to get 11 shipped. Will get you on another. My bad.

rpowelll:
All good, thanks for getting it merged in :tada

Issue #243, opened 2017-11-09
Yup schema.validate() options, show every error of a field at the same time


acezard:
Hi, sorry I didn't find another issue related to this problem. And since yup has first class support I thought it would be justified to ask about it.

So I'm trying to set an option, mainly here the abortEarly option. My goal is to have a password field with multiple errors displayed and it seems it was the only way. Couldn't find how to do that with validateSchema()

So I did the following instead:

```
  validate: (values) => {
    const schema = Yup.object().shape({
      email: Yup.string()
        .matches(/georges.abitbol@gmail.com/, 'cant change email'),
      providerName: Yup.string()
        .required('type your name'),
      password: Yup.string()
        .min(8, 'at least 8 chars')
        .matches(/[a-z]/, 'at least one lowercase char')
        .matches(/[A-Z]/, 'at least one uppercase char')
        .matches(/[a-zA-Z]+[^a-zA-Z\s]+/, 'at least 1 number or special char (@,!,#, etc).'),
      passwordConfirm: Yup.string()
        .equalTo(Yup.ref('password'), 'passwords don't match')
    })

    return schema.validate(values, { abortEarly: false })
      .then(() => {})
      .catch((err) => {
        throw err
      })
  }
```

This way I can get every error of a single field and map them to my components to display every error of a field at the same time. Isn't there a cleaner way ? Also I think it could be a good use case to showcase ? Thanks

clentfort:
@jaredpalmer I'm also facing this problem. Would you be open to a `validationSchemaOptions`-property on the `Formik`-component? If so Id' be willing to provide a PR.

AnastasiyaYelets:
have the same problem. it ll be great to add this in validateSchema

umpc:
:/

Woile:
I'm also interested in this, I would like to know what are the proposed solutions.
I'd like to have something like:
```js
errors = {
  name: [
    "Too short",
    "Cannot contain numbers",
  ]
}
```

Is it possible?

umpc:
Comments here keep getting upvotes, so I'm going to post this and hopefully keep this issue open for addressing it in the near future.

egemon:
@jaredpalmer could u help us please?

DrJacobHolden:
Would love to see this become an option.

mbrowne:
New PR for the new version of Formik (which is quite different since it now uses hooks):
https://github.com/jaredpalmer/formik/pull/1573

[...]

PR #387, opened 2018-01-29
Return all errors from `validationSchema`

clentfort:
**Summary:**

Currently only the first error reported by yup is picked up by Formik,
even though `abortEarly` is set to `false`.


clentfort:
I propose this changes to solve #243. If we can agree on an API I will add some unit-tests.

mbrowne:
I created a new PR that essentially does the same thing, based on the current `next` branch of this repo:
https://github.com/jaredpalmer/formik/pull/1573

Issue #378, opened 2018-01-25
[Next] Nested FieldArrays and setTouched not working together


rovansteen:
## Bug, Feature, or Question?
Bug

## Current Behavior
When you have nested `FieldArrays`, let's say the following path: `questions.0.values.0.label` where `questions` and `values` are the arrays and `label` is a regular input field. When `onBlur` is invoked on label, it wants to touch the path. But at that point (due to changing the `values` array) the current touched state is:
```js
questions: 0: {form_question_type_id: true, values: true}
```

which causes the following error in `setDeep`:
```
Uncaught TypeError: Cannot create property '0' on boolean 'true'
```
which happens on the last line of this method:
https://github.com/jaredpalmer/formik/blob/next/src/utils.ts#L69

## Desired Behavior
It should update the touched state to a deep nested state like this:
```js
questions: 0: {form_question_type_id: true, values: { label: true }}
```

## Suggested Solutions
I'm not sure I understand the current implementation of `setDeep`. Why not just clone the object and use something like https://github.com/kalmbach/bury?

## Info
 

- Formik Version: 0.11.0-beta.1 
- OS: Mac OS High Sierra
- Node Version: v9.4.0
- Package Manager and version: Yarn v1.3.2

davidhenley:
I opened this same issue a while ago. Still no response. My one problem with Formik@next right now.

jaredpalmer:
Damn wish I knew about bury earlier. Will have to check this out more detail

timc13:
why not stick with lodash and use [set](https://lodash.com/docs/4.17.4#set)  which works for bracket notation AND dot notation...

timc13:
i believe this is a result of using dot notation to access array items.

i am using bracket notation, and the array field gets completely replaced with a boolean.
See https://github.com/jaredpalmer/formik/pull/362

timc13:
awww, what was wrong with my PR? =X

jaredpalmer:
Awww I'm sorry man. I was on a grind and just missed this. Agh.

a-tarasyuk:
@jaredpalmer Hi. 

I noticed that `torched` and `errors` has a different data structure and after remove an item from `values` in `touched` property is not deleted. There is [`commit`](https://github.com/jaredpalmer/formik/commit/b32ebfd3f294996f88634bdaa9574f158274bb49#diff-7d037a4310f1022dee96554929bd14b3L64) where was removed `setFieldTouched`.

```
"values": {
[...]

PR #396, opened 2018-02-01
#378 Alter touched and error if FA.remove and FA.pop

jaredpalmer:
This solves #378 .

tl;dr
When either `pop()` or `remove()` is called, the related partial object within `errors` and `touched` are also removed. I also added proper generics (which somehow no one noticed) 🤷‍♀️


a-tarasyuk:
@jaredpalmer Does it mean that when we [`remove/push item`](https://github.com/jaredpalmer/formik/blob/master/src/FieldArray.tsx#L80), `setState` will be called 3 times(_also `render` fires 3 times_)? Maybe would be better to have one method which will update 3 fields and call `setState` only once? What do you think? Does it make sense?

Issue #367, opened 2018-01-23
[Next] setFormikState


jaredpalmer:
This will make it easier to handle a few edge cases and prevent double renders in certain situations. It will also act as the ultimate escape hatch.

Thoughts ?

Issue #51, opened 2017-07-12
validate values method


jonstuebe:
Is it possible to expose a validate fields/values method that can be run within the hoc? Right now it looks like validation is run on all fields when a value changes or on submit and it would be nice to run it whenever.

jaredpalmer:
Undocumented, but Formik also exports some helpers...

- [validateFormData: (values: any, validationSchema: any) => Promise<void>](https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L31) (it will throw if invalid).
- [yupToFormErrors: (yupError: any) => FormikErrors](https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L8) (will transform the caught Yup error in the shape of `props.errors`).

jaredpalmer:
See #61

PR #398, opened 2018-02-02
Add setFormikState and validateForm

jaredpalmer:
- #367 Adds an undocumented `setFormikState` to FormikActions / Bag so that `FieldArray` and `FormikPersist` can make atomic state updates. This may be renamed to `unstable_setFormikState` as it really should not be used outside of library authors. 
- #51 Adds `validateForm: (values?: any) => void` to FormikActions / Bag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1004,9 +1004,9 @@ Your form's values. Will have the shape of the result of [`mapPropsToValues`]
 (if specified) or all props that are not functions passed to your wrapped
 component.
 
-##### `validateForm: (values?: Values) => void`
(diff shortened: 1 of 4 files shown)


lardissone:
@jaredpalmer I think it should be a good idea to add `setFormikState` to the documentation. I've been long time to figure out how to make a simple change to the state.

jaredpalmer:
setFormikStste is fundamentally unstable to use as it limits backwards compat because of possible future state shape changes. Any documentation must come with a strong disclaimer.

Issue #401, opened 2018-02-03
RFC: Form Reducer / Middleware


jaredpalmer:
##  Feature

Allow users to hook into and reduce state changes. 

## Current Behavior

N/A

## Desired Behavior

```js

const MY_CUSTOM_FORMIK_ACTION = 'MY_CUSTOM_FORMIK_ACTION' 

<Formik 
  reducer={(prevState, nextState, action) =>{
      if (action.type === FormikActions.SET_VALUES) {
        // do something
         return nextState 
      }
      
     if (action.type === MY_CUSTOM_FORMIK_ACTION) {
        // since this action does not exist inside of formik's internal reducer,
        // nextState === prevState. 
        //  
         return {...prevState, warning: action.payload }
      }

      return prevState;
   })

  render=(({ dispatch, warning }) => 
    <Form>

      <button onClick={() => dispatch({ type: MY_CUSTOM_FORMIK_ACTION, payload: 'covfefe' }>
         Set a warning
      </button>
      {warning && warning}
       [/** .. */}
    </Form>
   }/>
 />
```

This would run on every state update. This approach would allow user's to add custom reducers and middleware on top of formik. For example, here's some pseudo code of a logger and quick and dirty formik-persist feature:

```js

const logger = (prevState, nextState, action) => {
   console.log(`--- FORMIK ${action.type} ---`)
   console.log('previous form state', prevState)
   console.log('action', action.type, action)
   console.log('next form state', nextState)
   console.log(`-----------------------------`)
   return nextState;
}

const persist = (prevState, nextState) => {
   // would probably debounce this. but this is core idea.
   window.localStorage.setItem('state', JSON.stringify(nextState));
   return nextState;
}

// Usage
<Formik 
  onSubmit={...}
  initialValues={JSON.parse(window.localStorage.getItem('state')) || { ... }}
  reducer={compose(persist, logger)}
  render={props => ... }
 />
```

The reason this is better than just using regular react state, is that the reducers should be highly reusable across your forms. Furthermore, if you don't like a specific feature of Formik, you can now safely extend or modify Formik. 

### Use cases:

- Transforming values. 
- Adding state (like warnings, submit counts, soft submits, etc.)
- Testing!

## Suggested Solutions

We would need to rewrite Formik's internals to use a "reducer" pattern. This is quite simple actually.

```js
export const FormikActions = {
  SET_VALUES: 'SET_VALUES',
  SET_ERRORS: 'SET_ERRORS',
[...]

PR #402, opened 2018-02-04
Added onChange callback & validation in componentDidMount

morgs32:
Also added storybook.
image


morgs32:
Can do. Also I didn't mean for storybook to serve as docs. It's been really handy as a development tool, something you would spin up locally. When you're working on this, and experimenting with the UI, are you doing a `yarn link` and always using the compiled files? I was looking for an easy way to run `formik` locally, being able to drop breakpoints and debug it via dev tools, etc. If you would prefer not to use storybook, perhaps a standalone app to run locally? Thoughts?

jaredpalmer:
I try not to rely too hard on UI while writing new stuff. This forces me to think through tests. If I need to link, I throw it in a local branch of React-parcel-example with typescript. The docs site also has runnable examples like React Router docs, so storybook is somewhat redundant

jaredpalmer:
componentWillUpdate is being deprecated. However, I’m def open to a separate storybook PR

jaredpalmer:
onChange encourages ppl to sync state in 2 places which usually never works out. I’m going to push up my work on the state reducer / middleware and would love to get your input on that though.

morgs32:
@jaredpalmer I see this RFC you have... https://github.com/jaredpalmer/formik/issues/401
Let me know if that's on a branch, I'm interested...

PR #404, opened 2018-02-05
better match withFormik validationSchema types

chodorowicz:
Currently when using withForm HoC `validationSchema` property it is incorrectly matched always to `any`. We're always loosing props' type (when using function for this property). With following change TS correctly discerns function form and simple Yup schema argument.

Before
`[ts] Parameter 'props' implicitly has an 'any' type.`

image

After
`parameter) props: Props`
image


jaredpalmer:
LGTM. 

cc @weswigham

jaredpalmer:
@chodorowicz should we actually use Yup's recently released TS types for this?

chodorowicz:
Yeah, that would make typing perfect. I'm not 100% sure how to handle this cross usage of types. It seems `@types/yup` should end up in main `dependencies` of formik 🤔 

I've done it this in another branch in my fork as a reference. 
https://github.com/chodorowicz/formik/pull/2/files
Tests and type checking passes without problem 🥇

PR #406, opened 2018-02-06
Upgrade rollup

jaredpalmer:
Bumping rollup and grabbing some uglify config from developit's microbundle.


jaredpalmer:
In a perfect world, we would move over to microbundle.

PR #409, opened 2018-02-07
Fix typings for FieldConfig['component']

jaredpetker:
The current typings weren't working for me at all, but this seems to work fine.


jaredpalmer:
interesting. not sure why this is. but yes can confirm this is a solid fix.

Issue #413, opened 2018-02-08
Update "form.touched" to suport nested fields


dccarmo:
Related to https://github.com/jaredpalmer/formik/issues/368

## Bug, Feature, or Question?

Bug

## Current Behavior

`FormikTouched` only allows one level of nesting. TypeScript complains if it's nested further:

`[ts] Property 'name' does not exist on type 'boolean'.`

## Desired Behavior

With the "Deep State" update I'd think that nested fields would also work.

## Suggested Solutions

Not sure if this is an implementation issue or a TypeScript definition issue.

## Info
 

- Formik Version: 0.11.7
- OS: macOS
- Node Version: 8.4.0
- Package Manager and version: npm 5.6.0

jaredpalmer:
This is a typescript issue.

jaredpalmer:
Can you paste in your code that is through the bug?

dccarmo:
Sure.

Here is the `FormFields` interface:

```
interface FormFields {
  business: {
    name: string;
  };
}
```

And then, in my `Field` component:

```
<Field
  name="business.name"
  render={({ field, form }: FieldProps<FormFields>) => (
    <div className="field">
      <div className="control">
        <input
          {...field}
          type="text"
          placeholder="Business Name"
          className={
            form.touched.business.name && form.errors.business.name
              ? 'input is-danger'
              : 'input'
          }
        />
      </div>
    </div>
  )}
/>
```

The TS error (`[ts] Property 'name' does not exist on type 'boolean'.`) is given on `form.touched.business.name`.

Looking at the type definition, it seems that `form.errors.business.name` won't work as well, because it has a `any`type. 

I tried searching for a way to get keys of nested objets the same way you're currently doing with `keyof` but didn't find anything, maybe it's not possible right now.

jaredpalmer:
Correct way to handle this: 

```js
<Formik
      initialValues={{ business: { name: '' } }}
      onSubmit={values => console.log(values)}
      render={({
        values,
        handleChange,
        handleBlur,
      }: FormikProps<{ business: { name: string } }>) => (
        <Form>
          <Field
            name="business.name"
            render={({
              field,
              form,
            }: FieldProps<{ business: { name: string } }>) => (
              <div className="field">
                <div className="control">
                  <input
                    {...field}
                    type="text"
                    placeholder="Business Name"
                    className={
                      form.touched.business &&
                      form.touched.business.name &&
                      form.errors.business &&
                      form.errors.business.name
[...]

PR #414, opened 2018-02-08
Fix #368. Relax types for FormikTouched 

### LINKED ISSUE #368: allow dot path & bracket path for setFieldValue, setFieldTouched, & setFieldError
## Bug
The exposed interface for `setFieldValue`, `setFieldTouched`, & `setFieldError` do not allow dot path names
## Current Behavior
only first level field names can be set.
## Desired Behavior
allow field names that contain dot path as well as bracket path.
## Suggested Solutions
change the `keyof Values` type for the field argument to `string`
## Info
 

- Formik Version: 0.11.0-beta.1
- OS: MacOS High Sierra 10.13.2
- Node Version: 8.9.4
- Package Manager and version: npm 5.6.0

jaredpalmer:
#368 #413 Adds TypeScript support for nth nested paths in FormikTouched

@weswigham Is there a better way to do this?

We would need some sort of recursive mapped type (i think)

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "tsc-watch": "1.0.7",
     "tslint": "5.5.0",
     "tslint-react": "3.2.0",
-    "typescript": "2.6.2",
+    "typescript": "^2.7.1",
     "yup": "0.21.3"
   },
   "lint-staged": {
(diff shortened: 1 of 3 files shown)


weswigham:
It seems like these should be clones of the `Values` object but with non-object leaf nodes replaced with "boolean". Pretty much a textbook example of what conditional types in ts2.8 are going to be for. In the meantime, I think you can probably get most of what you want (though it's a bit hacky, so I'd see if it works for you) with a definition like `+export type FormikTouched<Values> = { [field in keyof Values]?: boolean & FormikTouched<Values[field]> }`.

jaredpalmer:
Awesome. Yeah I had a feeling that conditional types were going to solve this. Can’t wait for 2.8

PR #419, opened 2018-02-08
#417 Tweak validation within FieldArray, rename dlv to getIn

### LINKED ISSUE #417: Exception when validating FieldArray
## Bug?

Trying out FieldArray (in 0.11) but running in to issues with validation. The form does not revalidate when adding/removing objects in array, and it also throws exception in a certain scenario

https://codesandbox.io/s/ykr57mqm61

Validation sort of works, but an exception occurs when doning theese repro-steps:
1 Remove, 2. Remove, 3. Submit, 4. Add, 5. Remove

jaredpalmer:
Fix #417

### New
- Add safer calls to array methods in FieldArray 
- Add new prop `validateOnChange?: boolean` to `FieldArray`, default is `true` to trigger validation function after array manipulations
- Add docs about validation gotchas when using FieldArray
- Export `setIn` and `getIn` functions out of utils (undocumented) for power users.

### Internals
- rename `dlv` to `getIn` 
- rename `setDeep` to `setIn` and swap argument order
- update test names
- add react-dom types to dev


davidwandar:
👍

PR #422, opened 2018-02-10
Add <FastField />

jaredpalmer:
kapture 2018-02-09 at 22 10 49

This adds a new `<FastField>` component to Formik. It renders fields independently while keeping an identical API to `<Field />`.

### Implementation notes

- FastField keeps local field state `value`, `error`, and `touched`. (Although i think we can drop `touched` because it is redudant), it passes _local state_ to the `field` prop in render and to the input.
- In FastField's handleChange: updates local state instead of the parent Formik's state. 
- In FastField's handleBlur: sends local state up to Formik's which causes a re-render (which is desired in this case).
- Syncing Formik reset behavior: in order to make FastField's state appear to be sync'd up to the parent Formik's, FastField registers itself with the parent Formik's `this.fields: { [fieldName: string]: () => void }` object. This object's keys are just the dotpaths/names of the fastfields and the values are local reset functions. During `resetForm()`, Formik iterates through `this.fields`'s keys, and calls the reset functions. Whenever a FastField unmounts, it also unregisters itself. There might be better / safer ways to implement this.
- FastField only validates on change if passed a validate={} prop and validateOnChange is true.
- I coded this on a train, so there may be some gotchas / edge cases i didn't consider

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -14,7 +14,8 @@ export interface FastFieldState {
   touched: boolean;
 }
 
-function diffObjectsExceptForKey(a: any, b: any, path: string) {
+/** @private Returns whether two objects are deeply equal **excluding** a key / dot path */
+function isEqualExceptForKey(a: any, b: any, path: string) {
   return isEqual(setIn(a, path, undefined), setIn(b, path, undefined));
 }
 
@@ -87,6 +88,7 @@ export class FastField<
       values,
       validationSchema,
       errors,
+      setFormikState,
     } = this.context.formik;
     const { type, value, checked } = e.target;
     const val = /number|range/.test(type)
@@ -95,7 +97,6 @@ export class FastField<
     if (validateOnChange) {
       // Field-level validation
       if (this.props.validate) {
-        console.log('field-validate');
         const maybePromise = (this.props.validate as any)(value);
         if (isPromise(maybePromise)) {
           this.setState({ value: val });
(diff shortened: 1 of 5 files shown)


slightlytyler:
I see the benefit in this but what are the downsides to using `FastField` vs a regular one?

jaredpalmer:
Fast Field can only run field-level validation onChange

jaredpalmer:
Let me add that.

src/FastField.tsx
```diff
@@ -78,20 +75,77 @@ export class FastField<
...
+          this.setState({ value: val });
+          (maybePromise as any).then(
+            () => this.setState({ error: undefined }),
+            (error: any) =>
+              this.setState({ error: getIn(error, this.props.name) })
+          );
+        } else {
+          this.setState({
+            value: val,
+            error: getIn(maybePromise, this.props.name),
```
jaredpalmer: We could compare the error object returned by top-level validation / validationSchema to see if error object is different from context.errors. If it is, we could update top-level parent here too. With that added wrinkle, this could actually fully replace Field without breaking changes (I think)

src/FastField.tsx
```diff
@@ -78,20 +75,77 @@ export class FastField<
...
+        }
+      } else if (validationSchema) {
+        // Top-level validationsSchema
+        const schema = isFunction(validationSchema)
+          ? validationSchema()
+          : validationSchema;
+        const mergedValues = setIn(values, this.props.name, val);
+        // try to validate with yup synchronously if possible...saves a render.
+        try {
+          validateYupSchema(mergedValues, schema, true);
```
jaredpalmer: This tries to validate synchronously with Yup. Yup distinguishes validation errors vs. other errors by error.name

src/FastField.tsx
```diff
@@ -111,25 +112,44 @@ export class FastField<
...
           (maybePromise as any).then(
             () => this.setState({ error: undefined }),
-            (error: any) =>
-              this.setState({ error: getIn(error, this.props.name) })
+            (error: any) => {
+              // Here we diff the errors object relative to Formik parents except for
+              // the Field's key. If they are equal, the field's validation function is
+              // has no inter-field side-effects and we only need to update local state
+              // otherwise we need to lift up the update to the parent (causing a full form render)
+              if (isEqualExceptForKey(maybePromise, errors, this.props.name)) {
```
jaredpalmer: This diffs the errors object between the parent's current errors and the one returned by validate. If it differs, then it means that the validation of the field has side effects. For example...

```js
const validate = (values: any) => {
  let errors: any = {};
  if (values.friends[0].name === values.email) {
    errors.email = 'cant be the same';
    errors.friends = [{ name: 'cant be the same' }];
  }
  return errors;
};```

In this situation, FastField will work act like the old Field and update the parent's Formik's state instead of the localized one--causing rerenders where necessary. This diff trick is not needed during onBlur / handleBlur because FastField _always_ updates the parent.

goodmind:
@jaredpalmer is this example in GIF available as code?

vmittal26:
I want to change one field on basis of another , with Field this scenario works 
but  typing is too slow on input fields , since my form is large , please let me know the solution !

PR #423, opened 2018-02-10
Attempt to move to create-react-context

jaredpalmer:
This PR attempts to removes the direct use of prop-types and instead uses create-react-context. 

Writing typed render prop components that take a generic is super awkward now. Ugh. 

To make tests pass, we will need to either use enzyme mount / or figure out something else

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ coverage
 .DS_Store
 next.d.ts
 legacy.d.ts
+.rpt2_cache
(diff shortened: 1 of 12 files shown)


jaredpalmer:
Will make a college try soon. My team is also upgrading to 16.3 this week

jaredpalmer:
See #594 .

jaredpalmer:
Any help finishing the move would be greatly appreciated.

PR #426, opened 2018-02-12
Field.children can be a function or a node

timc13:
As defined in the propType.


timc13:
Could we get this merged?  Currently, Typescript complains when I want to do this:
```
<Field type="select">
  <option>1</option>
  <option>2</option>
  <option>3</option>
</Field>
```

Issue #258, opened 2017-11-16
handleChange bug for empty number values: parseFloat('') is null


cloud-walker:
Input type number will be cast to `null` when cleared, because of this:

https://github.com/jaredpalmer/formik/blob/61a0869c121840fe7b69d9aba7442bdea409e425/src/formik.tsx#L389

Demo with the latest version of formik: https://codesandbox.io/s/m7268r7xoy

jaredpalmer:
True true.

jaredpalmer:
Should this be `parseFloat(val) !== null ? parseFloat(val) : ''`?

DrewBe121212:
Related issue: 284. 

Also, for checking the value, what about this?

    Number.isNaN(Number.parseFloat(val)) ? '' : Number.parseFloat(val);

jaredpalmer:
Submit a PR and with this change and a test if you get a chance

victordidenko:
Didn't make test though...

PR #437, opened 2018-02-14
Fix #284 and #258 - Empty input type=number results in NaN

### LINKED ISSUE #284: Empty input type=number results in NaN
I'm using Formik with Yup Schema validation.

I have an input field of type "number". If I enter a value and empty the field this results in a `NaN`. This behavior might be as expected. For that I added a transform function to my schema like that:

```
yup
.number()
.transform(value => {
  console.log(value, Number.isNaN(value) ? 0 : value 
  return Number.isNaN(value) ? 0 : value
})
```

Should it respect the transform function? Or is the schema actually only be used for validation, not for "cleaning" the data? If so, is there a way how to solve that? Maybe I'm missing something.
Of course, I can check the value before attaching it to the input field. But the value in the global values object stays the same.

victordidenko:
Fixes issues #284 and #258, when emptying input with type="number" causes error "Received NaN for the `value` attribute. If this is expected, cast the value to a string."

### DIFF
--- a/src/formik.tsx
+++ b/src/formik.tsx
@@ -420,8 +420,9 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
     e.persist();
     const { type, name, id, value, checked, outerHTML } = e.target;
     const field = name ? name : id;
+    let parsed;
     const val = /number|range/.test(type)
-      ? parseFloat(value)
+      ? ((parsed = parseFloat(value)), Number.isNaN(parsed) ? '' : parsed)
       : /checkbox/.test(type) ? checked : value;
 
     if (!field && process.env.NODE_ENV !== 'production') {


victordidenko:
@jaredpalmer Thanks for quick response! Do you have plans to draft new release any time soon?

jaredpalmer:
Yeah will cut one this morning

--
Jared
________________________________

strange-developer:
Great library. I've just run into the same issue as well. Do you have a rough estimate on when this fix will be published to NPM?

victordidenko:
🎉 as I see, new release has been drafted about three hours ago ^_^

jaredpalmer:
Try 11.11.

innerdaze:
I'm running 11.11 and still have this issue

jaredpalmer:
hrm. so perhaps the fix did not work?

innerdaze:
oh hang on... I'm using the `component` prop. if i leave that out then it works fine

innerdaze:
Ok I fixed it my end - might be worth documenting that if you use a custom component then you need to add a special case for number fields

PR #440, opened 2018-02-15
Fix <Field type=checkbox /> initial value

victordidenko:
When you set initial value in `mapPropsToValues` for checkbox and use Formik's Field component – it doesn't set `checked` property

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -37,7 +37,7 @@ export interface FieldProps<V = any> {
     /** Mark input as touched */
     onBlur: (e: any) => void;
     /** Value of the input */
-    value: any;
+    [x: string]: any; // 'value' or 'checked' field
     /* name of the input */
     name: string;
   };
@@ -167,10 +167,9 @@ export class Field<Props extends FieldAttributes = any> extends React.Component<
 
     const { formik } = this.context;
     const field = {
-      value:
-        props.type === 'radio' || props.type === 'checkbox'
-          ? props.value // React uses checked={} for these inputs
-          : getIn(formik.values, name),
+      [props.type === 'radio' || props.type === 'checkbox'
+        ? 'checked'
+        : 'value']: getIn(formik.values, name),
       name,
       onChange: validate ? this.handleChange : formik.handleChange,
       onBlur: validate ? this.handleBlur : formik.handleBlur,


prichodko:
I am against changing the structure of `field` prop based on a type. Because:

1. It could lead into unexpected behaviour when using custom Input components (it has to be documented)
2. It's a potential breaking change, imagine if someone has an input:

```js
<Field
 render={({ field }) => <input type="checkbox" checked={field.value} />} // and some other props
/>
```

What, I believe, is a better fix is to pass adjusted props only when it's a string component [here](https://github.com/victordidenko/formik/blob/304906db5992235b626ea4f487cbe946ca6bbe21/src/Field.tsx#L188).

prichodko:
I am not sure if we should change the structure of the `field` prop completely, shouldn't this be applied only to `string` components?

victordidenko:
@prichodko did you remove your comment? I received it to email, but I don't see it here...

You wrote an example
```
<Field
 render={({ field }) => <input type="checkbox" checked={field.value} />} // and some other props
/>
```
in that case there wont be problem, because `type` is undefined, and `field.value` will be filled.
Problem will be in case someone will write something like that:
```
<Field
 type='checkbox'
 render={({ field }) => <input type="checkbox" checked={field.value} />} // and some other props
/>
```

But I agree, that I didn't see that coming, may be you are right.
Should I change my PR? (I don't know though should I create new PR or this PR will change status to "open" after I commit something to my repo, or any new changes here will be automatically approved...)

prichodko:
Yeah I deleted my comment, because I needed more time to think about the use cases. 😄

How I use Formik at one of my codebases:
```js
const Checkbox = ({ name, ...rest }) => (
  <Field
    name={name}
    render={({
      field: { value, onChange },
      form: { errors, touched, setFieldTouched },
    }) => {
      const isTouched = touched[name]
      const hasError = errors[name]

      return (
        <Label>
          <Input
            type="checkbox"
            name={name}
            checked={value}
            onChange={e => {
              setFieldTouched(name, true)
              onChange(e)
            }}
            {...rest}
          />
          {isTouched && hasError && <Error>{error}</Error>}
        </Label>
      )
    }}
  />
```

When using `Field`  in conjunction with custom components I rely on the same `props` structure independently on the input type. After thinking, I think it should be kept this way. However last word is behind @jaredpalmer.

But still, there is a bug, when using `string` components.

jaredpalmer:
I'm not sure why we would change the default React behavior here.

prichodko:
What do you mean?

hrgui:
Value for checkboxes and radio are supplied by the dev, not coming from the initial form state itself.

For example, in radio:
`{"ice_cream": "strawberry"}`, we'd expect there is a radio with label strawberry to be selected, where value is strawberry.

Here's an example, of how it'd be used with formik
https://codesandbox.io/s/0xl776v8mv

The requested change in this PR would actually do the following: `field.checked = 'strawberry'`


For checkboxes:

1. If we want checkboxes as true/false, then value is not needed, we'd use checked (which this PR would work out)
Example: https://codesandbox.io/s/wonyrv91mw

2. If we want checkboxes as the value being other than true/false, then we'd probably need to do something about handleChange, as we can't use that since it sends in props.checked to formik:
Example: https://codesandbox.io/s/7kz1vjzq70

victordidenko:
@hrgui Those are very good points. I need to think about it.
Maybe use input's `value` in conjunction with `checked` field, to cover such cases...

victordidenko:
Oh, completely forgot about this PR >_<
I guess I'll close it for now, because I won't have time in a few days to make appropriate change, that would fit anyone :)

PR #444, opened 2018-02-17
resetForm should reinitialize this.initialValues even if nextValues is not provided

### LINKED ISSUE #441: resetForm should reinitialize this.initialValues even if nextValues is not provided
## Bug, Feature, or Question?

IMO Bug ;-)

## Current Behavior

Calling `resetForm` without `nextValues` causing that `dirty` stay equal to `true`, because `this.initialValues` is not updated.

## Desired Behavior

`resetForm` should also reset `this.initialValues` so `dirty` flag will be equal to `false` after reset. 

## Suggested Solutions

change this 3 lines https://github.com/jaredpalmer/formik/blob/master/src/formik.tsx#L574-L576
```
    if (nextValues) {
      this.initialValues = nextValues;
    }

```
on this 
```javascript
this.initialValues = nextValues ? nextValues : this.props.initialValues;
```

istarkov:
closes #441

### DIFF
--- a/test/formik.test.tsx
+++ b/test/formik.test.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { Formik, FormikProps } from '../src/formik';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { sleep, noop } from './testHelpers';
 
 interface Values {
@@ -16,6 +16,7 @@ const Form: React.SFC<FormikProps<Values>> = ({
   handleBlur,
   status,
   errors,
+  dirty,
   isSubmitting,
 }) => {
   return (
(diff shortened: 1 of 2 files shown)


test/formik.test.tsx
```diff
@@ -963,5 +983,37 @@ describe('<Formik>', () => {
...
+      tree.setState({ data: { name: 'Jared' } });
+
+      await tree
+        .find(Form)
+        .props()
+        .handleReset();
+
+      expect(
+        tree
+          .update()
```
istarkov: IMO should work without update call here, but it doesn't. Cant get why.

istarkov: because https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#for-mount-updates-are-sometimes-required-when-they-werent-before

PR #447, opened 2018-02-21
Fix typo in FieldArray example

newswim:
Surely no one would just copy a code sample and paste it into their production application.......

BUT just in case, this should fix a syntax error that they might encounter. :see_no_evil:


jaredpalmer:
😂

PR #451, opened 2018-02-22
Allow handleSubmit to be called without preventDefault

jeffbski:
In some use cases handleSubmit could be called without having access to a ReactEvent, for example when being invoked from a custom component or ReactNative.  Currently handleSubmit expected to be called with an object containing a preventDefault method and it would throw an error if called without one.

Added a test and modified the code to simply check for existence before attempting to call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,7 +326,7 @@ npm install yup --save
       - [`handleBlur: (e: any) => void`](#handleblur-e-any--void)
       - [`handleChange: (e: React.ChangeEvent<any>) => void`](#handlechange-e-reactchangeeventany--void)
       - [`handleReset: () => void`](#handlereset---void)
-      - [`handleSubmit: (e: React.FormEvent<HTMLFormEvent>) => void`](#handlesubmit-e-reactformeventhtmlformevent--void)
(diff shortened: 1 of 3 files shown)


jaredpalmer:
LGTM. Thanks for writing the test

jeffbski:
I'm not sure why the travis build failed. Test works locally. Any ideas?

jaredpalmer:
Not sure need to figure that out.

jeffbski:
looks like it might be failing in FastField, missing import?

jeffbski:
It is a case sensitive issue that started here https://github.com/jaredpalmer/formik/commit/6c3e0627e9e1463b3ef5629eff913a2b9b68878b

I'll create a PR to attempt to fix it.

jaredpalmer:
I think you are right.

jeffbski:
I'll rebase with master to pick up the rename fix

jeffbski:
yay! That worked.

jeffbski:
Thanks for the quick review and merge! I am loving Formik, so thanks for all the work you put into it.

PR #452, opened 2018-02-22
rename file formik to Formik

jeffbski:
On file systems that are case sensitive, typescript will fail
to find the file if it does not match the the name being
imported.

Commit
https://github.com/jaredpalmer/formik/commit/6c3e0627e9e1463b3ef5629eff913a2b9b68878b
switched to using Format in imports


jaredpalmer:
Weird that VSCode (what i use for Git) did not pick this up.

PR #454, opened 2018-02-23
Added type safety to both field and value of setFieldValue

### LINKED ISSUE #453: Overloaded methods break type safety
## Bug, Feature, or Question?
Bug

## Current Behavior
Due to the overloaded function definitions for `setFieldValue` [here](https://github.com/jaredpalmer/formik/blob/fde20a6ef323f789a6cb591f081ef00cd91273bc/src/Formik.tsx#L100), when a strict match is not found it falls back to the unsafe overloaded definition, which will take any field string and any value. This can lead to difficult to debug logic errors.

## Desired Behavior
If the `field` passed to `setFieldValue` is not a key on `Values`, an error should be thrown. If the data type passed as the value does not match the data type defined for that key, an error should be thrown.

## Suggested Solutions
PR: https://github.com/jaredpalmer/formik/pull/454

- Formik Version: 0.11.11
- React Version: 16.0.0
- TypeScript Version:  2.7.2
- CodeSandbox Link: [https://codesandbox.io/s/3yx632on3q](https://codesandbox.io/s/3yx632on3q)
- OS: Win10
- Node Version: 8.9.4
- Package Manager and Version: yarn 1.3.2

Shurelia:
Fixes #453. Removed overloaded function definitions. `setFieldValue` takes a type parameter which will be inferred based on the `field` passed to the method, which will then be used to look up the type to check against the passed `value`.

Note, I have not tested these changes using the <Formik /> syntax, only the withFormik<>()() syntax.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -89,15 +89,13 @@ export interface FormikActions<Values> {
   /** Manually set values object  */
   setValues(values: Values): void;
   /** Set value of form field directly */
-  setFieldValue(
-    field: keyof Values,
-    value: any,
+  setFieldValue<K extends keyof Values>(
+    field: K,
+    value: Values[K],
     shouldValidate?: boolean
   ): void;
-  setFieldValue(field: string, value: any, shouldValidate?: boolean): void;
   /** Set error message of a form field directly */
   setFieldError(field: keyof Values, message: string): void;
-  setFieldError(field: string, message: string): void;
   /** Set whether field has been touched directly */
   setFieldTouched(
     field: keyof Values,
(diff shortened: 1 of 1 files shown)


jaredpalmer:
This breaks everyone’s custom inputs tho

jaredpalmer:
I rather wait until TS 2.8 next month and we can use conditional types and avoid the generic all together

Shurelia:
@jaredpalmer Apologies, I'm not quite sure I follow. Why would this break everyone's custom inputs? There's no actual need to pass the type parameter, it's automatically inferred. The syntax is the same, it will now just throw errors if the string passed as `field` doesn't exist on the `Values` interface, or if the type of `value` passed doesn't match the value on that key.

Shurelia:
This breaks when used with the array[index] notation (e.g. field is something like `'data[2]'`, and the intent is to set index 2 on the data array, not values.data.2). I can't think of a way to type this at the moment, but this should be closed for now. I'll reopen it if I can think of anything that doesn't cause breaking changes and does improves type safety.

jaredpalmer:
In 2.8 we'll be able to use conditional types to alter the typing of FormikValue, FormikTouched, FormikError to account for dot-path notation

egmanoj:
> In 2.8 we'll be able to use conditional types to alter the typing of FormikValue, FormikTouched, FormikError to account for dot-path notation

@jaredpalmer Has this been fixed? I'm trying to do 

```ts
setFieldValue(`something[${index}]field`, "value", false);
```

and am getting a type error:

```
Argument of type 'string' is not assignable to parameter of type '"something" | "other" | "top_level" | "fields".
```

sentient-drone:
What's the status of this issue? We're upto typescript 3.3 so I was wondering if any further progress has been made.

Or is there a preferred alternative to dot notation property selectors?

PR #458, opened 2018-02-23
Blur active element before submitting form

STenorio99:
Blurs the active element so it's blur event can be ran then submits the form.

Currently, if a user types in a `<FastField/>` then submits a form without blurring the field, the updated value doesn't get passed into `formik`'s state. So the submitted values don't include whatever the user recently typed. Validation for the field also fails to update since it receives the old value.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -494,7 +494,16 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
     if (e && e.preventDefault) {
       e.preventDefault();
     }
-    this.submitForm();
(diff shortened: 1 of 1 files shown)


STenorio99:
Found that this is an error with Atlaskit's mk2 Button. Closing to bother them instead.

jeffbski:
@jaredpalmer could we get another version published to npm that picks up this recent set of changes? Thanks!

jaredpalmer:
Yup. Will do.

jaredpalmer:
Done.

PR #466, opened 2018-02-27
Add handler for submit with invalid form data

bramschulting:
This PR introduces a new config option which allows you to handle the case where a user tries to submit a form with invalid data.

Our use case is tracking what errors our users make so we can improve our forms, but I'm sure there are other interesting use cases for this feature.

I'm curious about your opinion one something. I'm not sure which function signature I prefer:

*Option A*
```typescript
onSubmitWithInvalidData?: (
  values: Values,
  formikActions: FormikActions<Values>,
  errors: FormikErrors<Values>
) => void;
```

or

*Option B*
```typescript
onSubmitWithInvalidData?: (
  errors: FormikErrors<Values>,
  values: Values,
  formikActions: FormikActions<Values>
) => void;
```

I went with option A for now, as it is consistent with `onSubmit`. But the most interesting thing in this callback are probably the errors. What do you think?

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -429,6 +429,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
       if (isValid) {
         this.executeSubmit();
       } else {
+        this.executeSubmitWithInvalidData(combinedErrors);
         this.setState({ isSubmitting: false });
       }
     });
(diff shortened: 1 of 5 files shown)


bramschulting:
@jaredpalmer any chance you could take a look at this?

jaredpalmer:
Sorry I never responded. I am thinking about this API. Need some time.

bramschulting:
@jaredpalmer Have you been able to think about the API? We're using a fork right now, but I'd love to be able to this version.

bramschulting:
Why is this PR closed?

jaredpalmer:
bot debacle. I reopened.

bramschulting:
Ah, I see. Bots these days! Thanks for reopening.

gabrielfnogueira:
redux-form has `onSubmitFail`, since it's pretty similar wouldn't it be better if the name were the same? Anyway, I'm migrating to formik from redux-form and this config would be great :grinning:

bramschulting:
I'm not familiar with `redux-form`, but I think `onSubmitFail` suggests that an actual submit was tried and failed. That's not what's happening in this PR. `onSubmitWithInvalidData` is called _before_ submitting.

ashikodes:
Here you are. Surfing the doc all day, expected this out of the box.

matths:
Anything new on this feature request from 2018?

matths:
FYI: I'll got with a workaround based on a validation check after the submitForm promise resolved.
(source: https://github.com/jaredpalmer/formik/issues/1580#issuecomment-770409235)

Issue #346, opened 2018-01-14
Pass field as prop to FieldArray


rovansteen:
## Bug, Feature, or Question?
Question / Feature

## Current Behavior
Currently the `<FieldArray />` gets only the `arrayHelpers` and `form` passed in as prop. Is there any reason it does not get the field property passed in? Now I have to get the values 'manually' from the form prop. 

## Desired Behavior
It would be nice to get at least the `value` from the field (based on the field name).

## Suggested Solutions
Passing a similar `field` prop as the `<Field />` component receives, except for the `onChange` and `onBlur`.
 
- Formik Version:  0.11.0-beta.1
- OS: Mac OS High Sierra
- Node Version: 9.3.0
- Package Manager and version: 1.3.2

charlax:
+1

It would be nice also for extra props to be passed down the component:

See https://codesandbox.io/s/6y8q0ql34z

```jsx
<FieldArray
            name="categoryIds"
            shouldBePassed="hello"
            component={props => {
              console.log(props)
              return null
    }
/>
```

In the current state, `shouldBePassed` will not be passed down to the rendered component.

dmlukichev:
Also got confused by the fact that rest props are not passed to `<FieldArray/>` component. It would be nice to allow it.

At the same time workaround will look like this:
```
<FieldArray
    name="categoryIds"
    render={props => (
        <Component 
            {...props} 
            shouldBePassed="hello"
        />
    )
/>
```

jaredpalmer:
Update: plan is to likely deprecated FieldArray and move these manipulations into core

Tsudhishnair:
How can I pass custom props to the child component in fieldarray now?

maddhruv:
Closing as `stale`

Issue #343, opened 2018-01-12
Automatic error and touched extraction inside <Form> component


Dremora:
## Current Behavior

`<Field>` component, when using with a custom component or render prop, passes down the following structure to the component being rendered: `{ field: { name, onBlur, onChange, value }}`. Extraction of `touched` and `error` for a particular field has to be done manually. This becomes quite painful when dealing with deeply nested fields. For instance, when accessing `errors.foo.bar`, one needs first to check the existence of `errors.foo`: `errors.foo && errors.foo.bar`.

## Desired Behavior

Pass down the following structure: `{ field: { error, name, onBlur, onChange, touched, value }}`.

## Info
 
There is at least one caveat: `<input>` and other HTML elements should not receive those extra properties. It might make sense to only implement this for render prop.

jaredpalmer:
My plan is to add a `meta` prop that holds these values so that ppl can still spread `field` over an input

Dremora:
Not using formik anymore, and this was actually one of the reasons for switching. But I believe this is still on the roadmap. Any chance of doing this as part of #828?

jaredpalmer:
Yes for sure.

prichodko:
https://github.com/jaredpalmer/formik/pull/579

jaredpalmer:
Pin

lourenci:
What about `dirty` by field?

I'd like to validate my field against my schema before the field is touched.

IronSean:
In the documentation here: (https://jaredpalmer.com/formik/docs/api/fieldarray) it says as of

>  Formik v0.12 / 1.0, a new meta prop may be added to Field and FieldArray

However, according to this that field is instead earmarked for v2. Is this already available somewhere or is the documentation incorrect?

jaredpalmer:
It's already in v2.

GoPro16:
Can we not also spread the meta for the field onto the component here: 
https://github.com/jaredpalmer/formik/blob/next/src/Field.tsx#L189

I find it that I have to use the `useField` hook in order to do everything I want rather than just passing custom Input tag into the `as` prop of `Field` because i need access to the meta for additional css className rendering.

jaredpalmer:
Yeah I have noticed this myself during v2 usage.

bschnelle:
@jaredpalmer can you help me understand why `meta` isn't passed into `component`?

jaredpalmer:
Backwards compat

bschnelle:
[...]

PR #167, opened 2017-09-20
Update Field structure, add meta props

VladShcherbin:
Proposal for updated `Field` props structure:

```js
const field = {
  input: {
    name,
    value,
    onChange,
    onBlur
  },
  meta: {
    dirty,
    error,
    initialValue,
    pristine,
    touched
  }
}
```

Very useful in custom components:

1. Show error

```js
// before
{form.touched[field.name] && form.errors[field.name] && (
  <p>{form.errors[field.name]</p>
)}

// after
{field.meta.touched && field.meta.error && (
  <p>{field.meta.error}</p>
})

// after with FieldError component
<FieldError state={field.meta} />
```

2. Reuse meta props in component functions, components

```js
// before
const fieldState = {
  touched: form.touched[field.name],
  error: form.errors[field.name]
}

const fieldClassName = generateFieldClassName(fieldState)

<input className={fieldClassName} {...field} />
<FieldError state={fieldState} />

// after
const fieldClassName = generateFieldClassName(field) // or pass `field.meta`

<input className={fieldClassName} {...field.input} />
<FieldError state={field} /> // or pass `field.meta`
```

TODO:

- [x] Add `initialValue` meta prop
- [x] Add `pristine` / `dirty` meta prop
- [x] Update typings
- [x] Update docs
- [x] Add example

PR #476, opened 2018-03-02
Add Field, FastField, ArrayField meta data props

prichodko:
There was a PR https://github.com/jaredpalmer/formik/pull/167, which unfortunately didn't land. However I would really love to see this landed, so I am taking this over. 

PR adds `meta` object to `Field` bag to ease getting of different states. Currently `meta` object consists of:

- error
- touched
- initial value

### Q
Is there anything we should add to it?

### WIP
- [ ] Documentation
- [x] Tests
- [x] Add meta to ArrayField

### Related issues
Resolves https://github.com/jaredpalmer/formik/issues/343
Resolves https://github.com/jaredpalmer/formik/issues/346

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -263,6 +263,7 @@ export class FastField<
     } = this.props as FieldConfig;
 
     const { formik } = this.context;
+
     const field = {
       value:
         props.type === 'radio' || props.type === 'checkbox'
@@ -272,18 +273,21 @@ export class FastField<
       onChange: this.handleChange,
       onBlur: this.handleBlur,
     };
+
+    const meta = {
+      touched: getIn(formik.touched, name),
+      error: this.state.error,
+      initialValue: getIn(formik.initialValues, name),
+    };
+
     const bag = {
       field,
       form: formik,
-      meta: { touched: getIn(formik.touched, name), error: this.state.error },
+      meta,
     };
 
     if (render) {
-      return (render as (
-        props: FieldProps<any> & {
-          meta: { error?: string; touched?: boolean };
-        }
-      ) => React.ReactNode)(bag);
+      return (render as (props: FieldProps<any>) => React.ReactNode)(bag);
     }
 
     if (isFunction(children)) {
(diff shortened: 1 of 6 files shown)


jaredpalmer:
Let’s add tests

prichodko:
@jaredpalmer tests added. 

I extended `FieldArray` meta props to contain `touched`, `errors`, `initialValues` and `isEmpty`. Those values also default to empty array if they are undefined. So user can always just use `touched[index]` without the need to check if it exists.

I also added a prop which will be named after the `name` and will contain to values for the FieldArray (which also defaults to empty array). So for example:

```js
<FieldArray
 name="friends"
 render={({ friends }) => (
  {friends.map(f => ...)}
 )}
```

Lastly, the `FieldArray` rendering logic is updated so it's same as in `Field` & `FastField`.

jaredpalmer:
This is a big breaking change. Not opposed to it though. Need to review

prichodko:
@jaredpalmer, sure. If necessary I can break it in two separate PR's. `Field` & `FastField` meta `prop` and `FieldArray` enhancements.

kheyse-oqton:
I came here looking for a solution for the FieldArray values problem, which could be solved with the proposal above:
```
<FieldArray
  name="friends"
  render={({ friends }) => (
    {friends.map(f => ...)}
  )}
/>
```

Personally, I'd choose not to name the prop after the name passed to FieldArray, because the name can contain full paths, e.g. `fb.friends`. Why not leave it in field.value? 
```
<FieldArray
    name="fb.friends"
    render={({ field, form }) => (field.value.map(friend => <Textarea value={friend} />)}
/>
```
This would be exactly the same as for the Field component.
```
<Field
     name="fb.friend"
     render={({ field, form }) => (<Textarea value={field.value} />)}
/>
```

prichodko:
@jaredpalmer I will separate this PR into two parts, adding a `meta` prop for a regular `<Field />` seems like a no-brainer to me, we can have discussion about `<FieldArray />` in another PR.

jaredpalmer:
Lets separate into 2 different PR's

prichodko:
Closing in favor of: https://github.com/jaredpalmer/formik/pull/579.

Issue #375, opened 2018-01-25
Add a submitAttemptCount value on form


henrykuzmick:
This would be useful for displaying errors after invalid submission whether it's a local validation or async.
This way a field could display an error when `form.errors[field.name] && (form.touched[field.name] | form.triedSubmit)`

jaredpalmer:
What about if we tracked `submitAttemptCount` (Or maybe call it just `submitCount`) as a number?

henrykuzmick:
Nice! that sounds better!

PR #477, opened 2018-03-02
Add submitCount to Formik

prichodko:
Adds a property to state, which tracks number of times user tried to submit the form.

If we agree on the implementation, I will add related tests & documentation 🤖.

### WIP
- [x] Tests
- [x] Documentation

### Related issues
Resolves https://github.com/jaredpalmer/formik/issues/375

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,19 +326,20 @@ npm install yup --save
       - [`handleBlur: (e: any) => void`](#handleblur-e-any--void)
       - [`handleChange: (e: React.ChangeEvent<any>) => void`](#handlechange-e-reactchangeeventany--void)
       - [`handleReset: () => void`](#handlereset---void)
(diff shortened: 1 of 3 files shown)


prichodko:
@jaredpalmer I added necessary tests and also renamed test file, so it matches with this change - https://github.com/jaredpalmer/formik/pull/452.

However, @jaredpalmer could you elaborate how doctoc works? When I run it, I don't see any changes made.

jaredpalmer:
you need to run `doctoc readme.md`

jaredpalmer:
What happens when resetForm is called?

prichodko:
Great catch @jaredpalmer, thanks. Fixed and test added.

jaredpalmer:
LGTM

PR #479, opened 2018-03-02
Fix FormikTouched type

### LINKED ISSUE #439: setTouched type error with TypeScript
## Bug, Feature, or Question?
Bug

## Current Behavior
```ts
interface FormValues {
    amount: number
}

export class MyForm extends Component<Props & FormikProps<FormValues>, State> {

	componentWillMount() {
		this.props.setTouched({amount: true})
	}
}
```

TS Error:
```
Argument of type '{ amount: true; }' is not assignable to parameter of type 'FormikTouched<FormValues>'.
  Types of property 'amount' are incompatible.
    Type 'true' is not assignable to type '(true & number) | (false & number) | undefined'.
```

## Desired Behavior
```ts
this.props.setTouched({amount: true})
```
Should be valid

## Suggested Solutions

Seems like FormikTouched type should be:
```ts
export declare type FormikTouched<Values> = {
    [field in keyof Values]?: boolean | FormikTouched<Values[field]>;
};
```
or simply
```ts
export declare type FormikTouched<Values> = {
    [field in keyof Values]?: boolean
};
```

instead of:
```ts
export declare type FormikTouched<Values> = {
    [field in keyof Values]?: boolean & FormikTouched<Values[field]>;
};
```
It's impossible to be 'true' _and_ 'number' at the same time 😄 .

This simple change fixes the compilation error.
I'll be using 
```ts
this.props.setTouched({amount: true} as any)
````

for now.

BTW, thanks for such an amazing library 🙌 . We're slowly migrating away from redux-form v2 to Formik and loving it (especially the built in TS support!)

## Info
- Formik Version: 0.11.10
- OS: MacOS 10.13.2
- Node Version: v8.9.4
- Package Manager and version: yarn v1.3.2

### PR DESCRIPTION
**Pull request by wfortin**
Fixes #439 

`FormikTouched` should only accept boolean values.

I don't think the `& FormikTouched<Values[field]>` is necessary. Another solution would be to set FormikTouched as:
```ts
export declare type FormikTouched<Values> = {
    [field in keyof Values]?: boolean | FormikTouched<Values[field]>;
};
```

Let me know what you think.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -32,9 +32,7 @@ export type FormikErrors<Values> = { [field in keyof Values]?: any };
  *
  * @todo Remove any in TypeScript 2.8
  */
-export type FormikTouched<Values> = {
-  [field in keyof Values]?: boolean & FormikTouched<Values[field]>
-};
+export type FormikTouched<Values> = { [field in keyof Values]?: boolean };
 
 /**
  * Formik state tree


jaredpalmer:
This won’t work with dotpath strings

jaredpalmer:
Let’s go with the alternative union type

piotrwitek:
Basically we cannot do much better right now, but soon with TS 2.8 that will change dramatically:
Here is a working implementation with typescript@next I tested today.

@jaredpalmer could you please confirm if the resulting type is align with how the API work (I do not have a full knowledge of all possible cases/shapes of `touched` object)?

```ts
type DeepFormikTouched<T> = {
  [K in keyof T]?: T[K] extends object ?
    DeepFormikTouched<T[K]> : boolean;
};

type Values = {
  id: string,
  social: {
    facebook: string,
  },
};
declare const touched: FormikTouchedObject<Values>;
// type touched = {
//    id?: boolean | undefined;
//    social?: {
//        facebook?: boolean | undefined;
//    } | undefined;
// }
```

jaredpalmer:
Yeah in 2.8 we’ll be able to change up some existing type issues. When’s 2.8 supposed to drop?

jaredpalmer:
Any updates?

piotrwitek:
TS 2.8 was published today, I could create PR with DeepFormikTouched tomorrow

jaredpalmer:
You rock! Looking forward to it

jaredpalmer:
Closing this for #543 (TS 2.8 upgrade)

PR #480, opened 2018-03-03
Pass though props to wrapped component

sepehr500:
I have run into a problem on my current project while using Formik where we want to pass props to the component Formik is wrapping, while keeping the component pure for easy testing. Currently our solution has been to do the following.

```
const renderComp = otherProps => formikProps => (<div> ...</div>)

<Formik render(renderComp({test: 'test'})) />

```

This is something we do often on the project, but is not as clean as Formik doing this itself. I have added `otherProps` to the Formik bag. It is possible to pass the props though normally, although I am no sure how intuitive that is.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -170,6 +170,8 @@ export interface FormikSharedConfig {
   isInitialValid?: boolean | ((props: object) => boolean | undefined);
   /** Should Formik reset the form when new initialValues change */
   enableReinitialize?: boolean;
+
+  otherProps?: object;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


prichodko:
Maybe I don't have enough context, but in the way how you use `Formik`, you should rather use `withFormik` HOC, which passes all props (https://github.com/jaredpalmer/formik/blob/master/src/withFormik.tsx#L144).

When using render prop, you don't really need to pass anything down, because you can do it yourself. As follows:

```js
const MyForm = props => (
    <Formik
        render={formikProps => (
             <form>...</form>
             )}
    />
)

```

jaredpalmer:
You already have access to all props and state in the component. It is redundant to pass them through. Just use React and wrap the component like you have been.

PR #482, opened 2018-03-04
Fix example for Field#validate in README.md

stefanwille:
The example for Field#validate in the README.md has a little error, as it doesn't reflect the API that it demos.  

This MR fixes that.


piotrwitek:
I just got hit by this issue, it looks good and could be safely merged @jaredpalmer

PR #496, opened 2018-03-07
Provide error and touched to component wrapped by Field

emckay:
Sorry for the unsolicited PR!

I often find myself using the render prop in `Field` like this:

```
<Field
  name="fieldName"
  render={({field, form}) => (
    <CustomInputProp
      {...field}
      error={getIn(form.errors, field.name)}
      touched={getIn(form.touched, field.name)}
      label="Field Label"
      customProp={123}
    />
  )}
/>
```

I have to type out the `error` and `touched` props each time, which gets a bit repetitive. This is something that could be easily computed by formik's `Field` component and passed down to the render prop function the same way that `value` is passed down.

Is there a reason this isn't already done? Please let me know if I'm missing something or if there's any thing to change in the PR. This only took a few minutes to throw together, so feel free to close this if the PR is way off base :)

### DIFF
--- a/test/Field.test.tsx
+++ b/test/Field.test.tsx
@@ -213,6 +213,10 @@ describe('A <Field />', () => {
                 const { handleBlur, handleChange } = formikProps;
                 expect(field.name).toBe('name');
                 expect(field.value).toBe('jared');
+                expect(field.hasOwnProperty('error')).toBe(true);
+                expect(field.hasOwnProperty('touched')).toBe(true);
+                expect(field.error).toBe(undefined);
+                expect(field.touched).toBe(undefined);
                 expect(field.onChange).toBe(handleChange);
                 expect(field.onBlur).toBe(handleBlur);
                 expect(form).toEqual(formikProps);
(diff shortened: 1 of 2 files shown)


test/Field.test.tsx
```diff
@@ -213,6 +213,10 @@ describe('A <Field />', () => {
                 const { handleBlur, handleChange } = formikProps;
                 expect(field.name).toBe('name');
                 expect(field.value).toBe('jared');
+                expect(field.hasOwnProperty('error')).toBe(true);
+                expect(field.hasOwnProperty('touched')).toBe(true);
+                expect(field.error).toBe(undefined);
+                expect(field.touched).toBe(undefined);
```
emckay: Is there a better way to test this? I didn't see any test data with values for `error` and `touched`

prichodko:
Thank you for your work, however it's a good practice to open an issue before working on a PR. It's already being implemented here - https://github.com/jaredpalmer/formik/pull/476.

emckay:
No worries! I figured it was faster to show my feature request in the code than to figure out how to communicate in words. Thanks for the work you've done on Formik!

Issue #486, opened 2018-03-05
async submitForm


rhyek:
## Bug, Feature, or Question?
Feature

## Current Behavior
`formikBag.submitForm` does not consider submit handlers that return promises.

## Desired Behavior
`formikBag.submitForm` should return a Promise (ideally the submit handler's own Promise) so the developer can handle server API rejections accordingly in cases where it is not possible / desirable to handle them in the original submit handler.

## Suggested Solutions
Would require changes in `src/Formik.tsx`. In it, methods `submitForm` and `executeSubmit`.

## Additional Information
Commented here #42, but created a new issue for visibility as I think I have a valid use-case. To reiterate, I want to create a custom reusable form component that handles server API validation errors in a generic way.

Small example of what I want to do:
```jsx
class LoginForm extends React.Component {
  async handleSubmit(values) {
    await axios.post('login', values) // no try/catch here
  }

  render() {
    return (
      <Formik
        onSubmit={this.handleSubmit}
        render={() => (
          <CustomGenericForm>
            ...stuff
          </CustomGenericForm>
        )}
      />
    )
  }
}

class CustomGenericForm extends React.Component {
  handleSubmit = async event => {
    event.preventDefault()
    const { formik } = this.context
    try {
      await formik.submitForm()
    } catch (error) {
      if (error.response && error.response.data && error.response.data.validationErrors) {
        formik.setErrors(error.response.data.validationErrors)
      } else {
        throw error
      }
    } finally {
      formik.setSubmitting(false)
    }
  }

  render() {
    const { children } = this.props
    return <form onSubmit={this.handleSubmit}>{children}</form>
  }
}

CustomGenericForm.contextTypes = {
  formik: PropTypes.object,
}
```

Or is there a way to achieve what I want with the current version?

---

- Formik Version:1.0.0-alpha.2
- React Version: 16.2

rhyek:
There is still a PR open for this issue.

Filson14:
+1 on that feature.
I'm trying to define handleSubmit as a function returning promise. Would like to get benefit from it when triggered submitForm() inside form component. Currently, submitForm returns "a promise", but it is not that one I expected. Triggering handleSubmit directly (as said in the docs, it should also start submit process) it returns nothing.

I'm not sure if i missed something but I guess it's a good issue to comment. Any suggestions (even for workaround) are welcome :wink:

I saw that there is something developed currently. #500 
Any updates about the work status?Long time since the issue has been reported.

[...]

PR #500, opened 2018-03-09
Async submitForm

rhyek:
Related #486. Wanted to get feedback on this. If the work is desired I can add some unit tests.


rhyek:
@jaredpalmer any thoughts on this? I'm using this in a project, currently, and it's working nicely.

kevinmook:
For what it's worth, this would be a very helpful change for us!

jaredpalmer:
This would be a major breaking change as it is currently implemented. However, this is possible to do. The trick would be to simply check if the function is a promise or not (just like validation) and keep functionality as is, or if it is a promise, call setSubmitting accordingly. They’ll be some gotchas (I think) related to error handling but overall I think this is a good feature.

rhyek:
@jaredpalmer, thanks. What breaking changes do you expect to stem from this change? This was one of my concerns as well, honestly. Currently all tests are passing and I have a mix of several forms some using this functionality and others not. Everything is working normally in all configurations.

rhyek:
Also, I'm sure you know this, but the way I did it, I'm generalizing most of the behavior with `Promise.resolve()` which will work regardless of whether the parameter is a "thenable" or not. For example, `Promise.resolve(2).then(n => // n === 2)`.

jaredpalmer:
I missed that you always wrapped in a promise. I need to think about this and try it out on our codebases. On the one hand it makes testing consistent `await submitForm`. However it would also cause an unnecessary render if submit is truly synchronous.

mgyarmathy:
@jaredpalmer @rhyek any chance this will be merged anytime soon?

slorber:
Hi,

Also need this, because when submitting forms, I use the returned promise to display a spinner inside the submit button.

See https://github.com/slorber/react-nested-loader

dclark27:
This would also be a great benefit for my team. We want to be able to do something on valid form submission.

tomshane:
I also find this feature very handy and i could use it immediately. Any chances this could be merged any time soon? Thanks.

quolpr:
Any updates on that?

rhyek:
Closed. Awesome!

jaredpalmer:
We are working on it internally.

jaredpalmer:
Will have something up this weekend or early next week

jaredpalmer:
We can reopen this. Sorry I meant to keep this one open.

jaredpalmer:
Was able to reopen. Phew. Sorry about that!

Filson14:
Hi @jaredpalmer. Is there any chance to include that feature? It would be really helpful. Any news?

src/Formik.tsx
```diff
@@ -411,10 +411,14 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
       }
     }
 
-    this.submitForm();
+    this.submitForm().catch(error => {
+      if (error.name !== 'ValidationError') {
+        throw error;
```
jaredpalmer: Why do we want to throw this? Why not just set it to `apiError` or something?

rhyek: We could do that, but the error does not necessarily have to be from an api. It could be a network error, for example. Also, an api might be responding with specific http status codes that the user might need to just catch and handle accordingly. I do not see the benefit of hiding the exception.

rhyek: https://en.m.wikipedia.org/wiki/Error_hiding

jaredpalmer: How would the user catch this? People who do SSR cannot use ErrorBoundary / componentDidCatch?

rhyek: You can see an example in #486.

src/Formik.tsx
```diff
@@ -428,7 +432,7 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
     return this.runValidations().then(combinedErrors => {
       const isValid = Object.keys(combinedErrors).length === 0;
       if (isValid) {
-        this.executeSubmit();
+        return this.executeSubmit().catch();
```
slorber: I'm not sure why we catch here. I would expect form submission to reject for cases where form in not valid and submission fails

src/Formik.tsx
slorber: @rhyek according to my understanding, ValidationError is something related to "yup" only, and this.submitForm() does not throw validations "combinedErrors" when there are errors (in yup or not), so this catch should basically do nothing. Note: runValidations currently does not reject on errors

I suggest making "submitForm" throw" on any error (validation + async calls + others), and catching here all "expected exceptions" (ie validation/network errors) and rethrow unexpected ones so that global error handler can log them (ie report to sentry or something). Maybe this strategy can be pluggable so that user may customize it to his own needs.

rhyek: I am not sure, to be honest. Looking at the code quickly it seems there's some new things I hadn't seen before. This is PR is kind of old.

lashiec9:
Has anyone been able to work around this issue while waiting for the PR to go through?

jehartzog:
@lashiec9 I was able to for our use case, but it was incredibly ugly and I don't recommend anybody use this same method 😆 

Basically we for all the forms we wanted to track to make sure submission was complete, when we created their unique id to an array. We'd then pass down a prop to each form `registerSubmitCompete(id)` that the form would call at the very end of it's `onSubmit` handler.

The parent could then trigger all children forms to submit, and would then poll until all the child forms called their `registerSubmitCompete()`. Here is some of the code:

## In parent
```js
  registerSubmitComplete = id => {
    this.submittingForms = this.submittingForms.filter(form => form.id !== id);
  };

  submitChildForms = async () => {
    // This would work if Formik submitForm returned the promise from the submit handler
    // https://github.com/jaredpalmer/formik/pull/500
    // const allSubmitFunctions = this.submitForms.map(form => form.submitForm);
    // // Executes all the submit functions, and returns a promise that resolves once they all resolve
    // await Promise.all(allSubmitFunctions.map(func => func()));

    this.setState({ isSubmitting: true });

    // Instead we just call it sync, and use a callback method to track when complete
    this.submitForms.forEach(form => form.submitForm());

    // Make a copy of all forms that need to be submitted
    this.submittingForms = [...this.submitForms];

    // Wait until the callback comes that each submit is done
    await new Promise(resolve => {
      const timerCheckSubmitted = setInterval(() => {
        if (this.submittingForms.length <= 0) {
          resolve();
          clearInterval(timerCheckSubmitted);
        }
      }, 100);
    });

    this.setState({ isSubmitting: false });
  };
```

Again I wouldn't really recommend anyone use this, but for us we had already built up enough forms assuming we could manage child form submissions that this little hack was quicker than a better refactor.

randycoulman:
@lashiec9 Here's our work-around for `handleSubmit` not being async: https://gist.github.com/randycoulman/608271914f29f580f07f44d38197742d.  This may not work for everyone, but it seems to do the trick for us. Note that we use higher-order components (including `withFormik`) and [recompose](https://github.com/acdlite/recompose) in our system, so that's what we're using.

jaredpalmer:
Async submit has just been shipped in v2 canary. `npm i formik@canary`

PR #501, opened 2018-03-09
Update README.md

browniefed:


browniefed:
Not sure the verbiage. Need to use `submitForm` for React Native but didn't see it documented.

PR #513, opened 2018-03-14
Stop bundling external dependencies for cjs and esm bundles

vlad-zhukov:
Currently `formik` bundles contain all external dependencies, which add unnecessary KBs to final app bundles.

Before: ~190KB
After: ~43KB

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -4,6 +4,7 @@ import replace from 'rollup-plugin-replace';
 import resolve from 'rollup-plugin-node-resolve';
 import sourceMaps from 'rollup-plugin-sourcemaps';
 import uglify from 'rollup-plugin-uglify';
+import pkg from './package.json';
 
 const shared = {
   input: `compiled/index.js`,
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Maybe i don't understand, but at this point we might as well move all deps to peerDeps yes?

Andarist:
> Maybe i don't understand, but at this point we might as well move all deps to peerDeps yes?

Not really, deps will be installed by consumers automatically while peerDeps won't - which is obvious, so I guess I might have misunderstood your intent.

Anyway - the point of this PR is to keep deps as deps and let npm/yarn dedupe those dependencies when installing multiple packages, by bundling them into the source code you effectively prevent that.

jaredpalmer:
The ts types were whack and I couldn't figure them out initially, was easier to write in TS, when I last checked like 6 months ago

jaredpalmer:
@Andarist what's need to solve the microbundle x prop-types issue?

Andarist:
I'm not yet sure - need to investigate why it happens (how rollup-plugin-commonjs works) and maybe we could fix it upstream in the plugin - if not we can always hardcode "the fix" in microbundle's code.

PR #521, opened 2018-03-15
Improve RN support, allowing curried handleChange to accept a string

jaredpalmer:
This PR does 2 things. 

1. Only calls e.persist if it exists in handleChange and handleBlur. 
2. Allow curried handleChange to accept a string as the first argument to the returned function thus allowing `onChangeText={handleChange('firstName')}` to work in React Native and React Native Web.

cc @browniefed

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -9,7 +9,6 @@ import {
   isEmptyChildren,
   setIn,
   setNestedObjectValues,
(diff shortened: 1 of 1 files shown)


browniefed:
Sweet, should make my forms a little cleaner no more `onChangeText={value => setFieldValue(paths.mobile, value)}`

PR #522, opened 2018-03-16
Pass Formik props into validateYupSchema() as context

### LINKED ISSUE #503: Pass component props as context for yup validation
## Feature
Yup validation can accept the context which can be used for validation purposes, but which is not a part of schema/values. Can we pass props as context? 

## Current Behavior
I see that [validateYupSchema](https://github.com/jaredpalmer/formik/blob/12c100bc8894e42206729e71f1cdf3765ad7c33c/src/Formik.tsx#L743) can accept context argument but it's not used anywhere.

## Desired Behavior
Be able to use component's props and validation rules.

## Suggested Solutions
Add component's props as context for `validateYupSchema()`

berdyshev:
closes #503


jaredpalmer:
Does this have any impact on existing schemas? Cc @jquense

jaredpalmer:
Otherwise LGTM

berdyshev:
No, it doesn't

jquense:
you might want to scope the prop to one (e.g. context), folks will start passing props just to pass them through which will limit Formik's ability optimize there.

jaredpalmer:
Can we back up a sec? Why do we need this? Whenever validate and validationSchema are called you have access props and values.

berdyshev:
validationSchema doesn't provide access to props inside the schema. Or I've missed something? There are could be some state parameters which can influence to the validation process, but they aren't part of values.

For example, in my case I have to same form for creating/editing of an entity, but the validation process is a bit different for creating/editing: some fields became non-required, some of them are skipped at all, etc. But this "mode" is not a part of the values — it's a prop of the Form Component.

jaredpalmer:
As described in the docs `validationSchema` prop in `<Formik>` can be a function that returns a schema. In `withFormik()`, `validationSchema` can be a function where the first argument are `props` as well. 

https://github.com/jaredpalmer/formik#validationschema-schema----schema

berdyshev:
As I understand, this doesn't matter, because `context` could be passed into schema only with some methods: `cast` and `validate`

piotrwitek:
I think that providing a way to pass a validation context might be really usefull, but I don't like the proposal to pass props directly as the context value.

@berdyshev right now we can achieve the same goal by using `validate` with yup like in below sample (but it is cumbersome and you need to know intricaties of Formik internals to know how to use throw with yupToFormErrors helper)
```jsx
const MyEnhancedForm = withFormik({
...
  validate: (values, props) => {
    const context = props.context || {};

    const schema = yup.object({
      email: yup
        .string()
        .when(
          "$emailRequired",
          (isRequired, schema) => (isRequired ? schema.required() : schema)
        )
    });

    return schema
      .validate(values, { abortEarly: false, context })
      .catch(err => {
        throw yupToFormErrors(err);
      });
  },
...
```

@jaredpalmer with context we could very easily make dynamic changes to schema based on values or props changes as in the example above, maybe adding a new method would be a better approach than passing props directly, something like this:
`validationSchemaContext?: Context | ((values, props) => Context)`

That method could be used when calling `validateYupSchema` function to provide static or dynamic (when used as a function) Context object.

---

I have created an example playground to test the `validate` workaround solution (might be usefull for people looking for a workaround to linked issue): https://codesandbox.io/s/oqvp75vxxz

jaredpalmer:
@piotrwitek That's awesome. Thank you for taking the time to make the sandbox.

While I am always hesitant whenever we expand the API surface, I think `validationSchemaContext` is the right API.

@berdyshev Go ahead and update so we can merge!

piotrwitek:
@jaredpalmer no problem :) you can ping me if you need a code review and testing as I'm already quite familiar with the code responsible for schema validation

jaredpalmer:
@piotrwitek If you have free bandwith, would be great to codify this thread into an "Advanced Validation" guide. Want to take a stab?

piotrwitek:
@jaredpalmer yea sure I might try. What would be the outline, should we continue with a separate issue for discussing the details?

jaredpalmer:
This needs to be added to docs. Then I'm fine with merging

leesiongchan:
@berdyshev can we make `FastField` supports this change as well?

leesiongchan:
@jaredpalmer LGTMerge?

leesiongchan:
ping @jaredpalmer

leesiongchan:
ping @berdyshev @jaredpalmer

timc13:
eagerly awaiting this!

timc13:
ping @berdyshev

AmenZhou:
I need this PR to support my forms, how can we get it merged?

zarv1k:
How can we get it merged?

mikkelwf:
...

andercard:
Ping

mastacheata:
> How can we get it merged?
@berdyshev forgot to adjust the tests for the new property.
Also it seems there are now merge conflicts because part of this was implemented already (for some strange reason formik will now pass along context to yup, but there's no way for the formik user to pass context to formik yet)

probablyup:
Happy to say this finally landed in 2.3

PR #527, opened 2018-03-20
checkbox and radio value is set using 'checked' property

kheyse-oqton:
It seems like the (initial) value of a checkbox and radio was not set correctly.

```<Field type="checkbox"  name="a.friend" />```

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -164,8 +164,12 @@ export class Field<Props extends FieldAttributes = any> extends React.Component<
     const field = {
       value:
         props.type === 'radio' || props.type === 'checkbox'
-          ? props.value // React uses checked={} for these inputs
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by prichodko** [comment]
This would broke the behaviour for users using `render` prop. See this PR: https://github.com/jaredpalmer/formik/pull/440.

**Comment by kheyse-oqton** [comment]
You mean when you use the render prop in combination with type="checkbox"?

Do you have an other fix for this bug since the checkbox type is now broken?
I use this workaround for now:
```
<Field
        name={`a.friend`}
        render={({ field, form }) => (
          <input
          type="checkbox"
          name={field.name}
          checked={field.value}
          onChange={form.handleChange}
          />)
/>
```

paresh-salunke:
Do we have an example for radio button ? Is it going to be with radio button?

PR #528, opened 2018-03-20
Add size-limit to prevent accidental bloat

prichodko:
After this https://github.com/jaredpalmer/formik/pull/513 great catch and huge decrease of bundle size, I decided to add [size-limit](https://github.com/ai/size-limit) to prevent from accidentally bundle something what would drastically increase it.

_Note:_ I noticed `prettier` on `formatOnSave` in VSCode formatted also `package.json`, but I decided to leave it, because it's being used in a library and `JSON` is supported format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "typings": "dist/index.d.ts",
   "files": ["dist"],
   "scripts": {
-    "test": "jest --env=jsdom",
+    "test": "jest --env=jsdom && npm run size",
     "test:watch": "npm run test -- --watch",
     "start":
       "cross-env NODE_ENV=development tsc-watch --project tsconfig.base.json  --onSuccess \"rollup -c\"",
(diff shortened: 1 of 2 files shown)


vlad-zhukov:
This is good idea but wow, `size-limit` adds lots of dependencies. It looks like it uses `webpack` under the hood but we use rollup instead. Perhaps [`bundlesize`](https://github.com/siddharthkp/bundlesize) is a better candidate?

prichodko:
@vlad-zhukov I was thinking about `bundlesize`. However there is a big difference between those two 
 - `bundlesize` outputs just a file sizes, `size-limit` shows the size of lib + it's dependencies (prevents unexpected huge dependency).

More info [here](https://github.com/ai/size-limit#comparison-with-bundlesize). What do you think?

jaredpalmer:
What should we do here?

prichodko:
Just merge, if you agree it's a good thing to have. Let me know and I will rebase the branch.

jaredpalmer:
Lets rebase and get it going

prichodko:
@jaredpalmer `next` or `master? 😄

jaredpalmer:
next

jaredpalmer:
Wait woah. Sorry I am out of it. Formik next is not what I want. I got confused because of how I am working on Razzle. Sorry about this. All against master. oof.

PR #531, opened 2018-03-23
onChange prop

rhyek:
This adds an `onChange` prop to the `Formik` component that, when defined, replaces the `handleChange` function returned from `getFormikBag`. This means this new `handleChange` will be available on both the render props and any child components obtaining the formik bag from its context.

My use-case for this is I have a series of forms across several "steps" where Formik's validations are run when I hit the *next* button on each step. The values for all these forms are being sent to Redux where a persisting middleware is saving everything to the `sessionStorage`. I was previously persisting to Redux during my submit handler (next button), but I need to persist on every change even *before* hitting *next* for a better user experience. I do not mind persisting invalid data since I can't advance to the next step until everything is valid and I don't want to lose any values when reloading or hitting *back* (which doesn't do validation).

Anyways, this is especially useful to me since I have a custom `Field` component that is grabbing `handleChange` from the context and I want it to be consistent with the one provided via render props.

This is how I am using it:

```js
render() {
  return (
    <Formik
      initialValues={form}
      onChange={(event, { setFieldValue }) => {
        const { name, value } = event.target
        setFieldValue(name, value)
        this.props.update({ [name]: value }) // redux
      }}
      onSubmit={this.handleSubmit}
      render={({
        handleChange, // from above
        handleSubmit
      }) => ...}
    />
  )
}
```

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -693,8 +693,16 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
       handleBlur: this.handleBlur,
       handleChange: !this.props.onChange
         ? this.handleChange
-        : (e: React.ChangeEvent<any>) => {
-            this.props.onChange!(e, this.getFormikActions());
+        : (eventOrString: any) => {
+            const result = this.handleChange(eventOrString);
+            if (typeof result === 'function') {
+              return (e: React.ChangeEvent<any>) => {
+                result(e);
+                this.props.onChange!(e, this.getFormikActions());
+              };
+            }
+            this.props.onChange!(eventOrString, this.getFormikActions());
+            return;
           },
       handleReset: this.handleReset,
       handleSubmit: this.handleSubmit,
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -182,6 +182,14 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
   initialValues: Values;
 
   /**
+   * Change handler
+   */
+  onChange?: (
+    e: React.ChangeEvent<any>,
+    formikActions: FormikActions<Values>
+  ) => void;
+
+  /**
    * Reset handler
    */
   onReset?: (values: Values, formikActions: FormikActions<Values>) => void;
@@ -247,6 +255,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
     validateOnBlur: PropTypes.bool,
     isInitialValid: PropTypes.oneOfType([PropTypes.func, PropTypes.bool]),
     initialValues: PropTypes.object,
+    onChange: PropTypes.func,
     onReset: PropTypes.func,
     onSubmit: PropTypes.func.isRequired,
     validationSchema: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
@@ -682,7 +691,11 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
       registerField: this.registerField,
       unregisterField: this.unregisterField,
       handleBlur: this.handleBlur,
-      handleChange: this.handleChange,
+      handleChange: !this.props.onChange
+        ? this.handleChange
+        : (e: React.ChangeEvent<any>) => {
+            this.props.onChange!(e, this.getFormikActions());
+          },
       handleReset: this.handleReset,
       handleSubmit: this.handleSubmit,
       validateOnChange: this.props.validateOnChange,


rhyek:
I was also considering not having the provided function **replace** Formik's internal `handleChange`, but rather have the latter call the former after it finished its own processing. Sort of like a callback. That way it would also be similar to how the `onSubmit` handler works.

I could easily make that change.

rhyek:
I went ahead and made that change. So now I use it like this:

```js
render() {
  return (
    <Formik
      initialValues={form}
      onChange={(event, formikBag) => { // formikBag still provided even though I'm not using it here
        const { name, value } = event.target
        this.props.update({ [name]: value }) // redux
      }}
      onSubmit={this.handleSubmit}
      render={({
        handleChange, // from above
        handleSubmit
      }) => ...}
    />
  )
}
```

jaredpalmer:
Hey thanks for this. Couple of notes....

1. There are no event.target in React Native
2. Formik is an uncontrolled component. The problem with onChange is that you are effectively encouraging ppl to store the same state in 2 places like you’ve shown in your example. This is never a good idea in React. That being said, I understand the desire. If you absolutely must sync state with Redux, you can do so in a component that’s accepts Formik state and the update action as a prop and then implement componentDidUpdate or componentWillReceiveProps. Again though, this is discouraged because you are attempting to sync state in 2 places, which usually can’t be done. Ultimately, Formik should hold your forms state and this may mean lifting it up much higher into your component tree than you feel comfortable with. However, that’s just how React works right now. If there was an onChange prop it would either have to be like the state reducer/middleware RFC. I am working on this locally and hope to push up my work soon.

rhyek:
@jaredpalmer I have several forms in my project and am only using this pattern in a couple of places. I agree that state should only be held in one place, normally. 

Your solution with a component taking the state as props and it doing the redux work sounds interesting, but it might not me the most elegant approach. My appreciation for Formik comes from it's simple yet effective API, and my solution sort of adheres to that style.

Your concerns are valid though, but I'm wondering if it's something that could be handled by just documenting it as having a different purpose than doing dual state storage? There might be other use-cases for this functionality.

jaredpalmer:
@rhyek if this were to be implemented, I would opt for a more global concept of onChange and not just intercepting handleChange. Along those lines, the code in this PR is too naive in that it does not consider changes to values through setFieldValue, setValues, reset, and reinitialize. This will cause confusion. In regards to solving your problem, by creating a component like the one I described, you can abstract your sideeffect into a declarative component which you can move around as needed. In my experience this turns out to be better down the road...


For example...
```js
<Formik 
  render={({ values }) => 
    <form>
      <MyDeclarativeFormSideEffect values={values} />
       {/* ...  */}
    </form>
  }
/>
```

where `<MyDeclarativeFormSideEffect values={values} />` is something like:

```js
class MyDeclarativeFormSideEffect extends React.Component {
  componentDidUpdate(prevProps) {
     if (prevProps.values !== this.props.values) {
        this.props.update(/ ** whatevs */)
     }
  }
  render() {
    // doesn't actually render anything
    return null
  }
 }

// You could connect directly to redux if you need to here too.
export default connect(s => ({ thing: s.thing }) , { update: updateThing })(MyDeclarativeFormSideEffect)
```

rhyek:
> @rhyek if this were to be implemented, I would opt for a more global concept of onChange and not just intercepting handleChange. Along those lines, the code in this PR is too naive in that it does not consider changes to values through setFieldValue, setValues, reset, and reinitialize.

This seems reasonable to me and I agree.

Your component idea is ok. I saw that you did something similar with formik-persist. That is a pattern than I have not seen any where else. I'll admit I am somewhat new to React, but tbh, I think a component like that belongs as a parent and not a child of Formik. If I modified my `onChange` prop's behavior and made the callback receive all changed values and not the event, I think It could all be tied up nicely as:

For formik-persist:
```js
render() {
  return (
    <FormikPersist
      name="some-form"
      render={(persistedValues, persist) => (
        <Formik
          initialValues={persistedValues || form}
          onChange={(changedValues, formikBag) => { // formikBag still provided
            persist(changedValues)
          }}
          onSubmit={this.handleSubmit}
          render={...}
        />
      )}
    />
  )
}
```

For my use-case:
```js
render() {
  return (
    <Formik
      initialValues={form}
      onChange={(changedValues, formikBag) => { // formikBag still provided
        this.props.update(changedValues) // redux
      }}
      onSubmit={this.handleSubmit}
      render={...}
    />
  )
}
```

Anyways.

>  If there was an onChange prop it would either have to be like the state reducer/middleware RFC. I am working on this locally and hope to push up my work soon.

That sounds interesting and I'm curious to see what you have in... _store_. Heh. Love the project, btw. Keep up the good work.

PR #543, opened 2018-03-28
Feature: Upgrade typescript v2.8 + FormikTouched upgrade to be recursive

piotrwitek:
Hello @jaredpalmer 
Here comes the upgrade to TypeScript and the type improvements for touched. 
Below all the necessary changes:
- upgraded TS to v2.8.1 - necessary for conditional types (there was 2 small errors coming from upgrade I needed to fix -> 80ceaef)
- upgraded prettier - because it was showing errors related to a new conditional types syntax
- updated FormikTouched type to become recursive for nested objects
- added new test file containing type tests (I included some test example for new FormikTouched, but it should be the base to add other Generic Types tests as well to prevent regression in types)

Please tell me what you think. I'll be glad to refine some more types after this is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "rollup-plugin-sourcemaps": "0.4.2",
     "rollup-plugin-uglify": "^3.0.0",
     "ts-jest": "^21.2.4",
-    "tsc-watch": "1.0.7",
+    "tsc-watch": "1.0.17",
     "tslint": "5.5.0",
     "tslint-react": "3.2.0",
     "typescript": "2.8.1",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "tsc-watch": "1.0.7",
     "tslint": "5.5.0",
     "tslint-react": "3.2.0",
-    "typescript": "^2.7.1",
+    "typescript": "2.8.1",
     "yup": "0.21.3"
   },
   "lint-staged": {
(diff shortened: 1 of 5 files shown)


jaredpalmer:
I will have time to review tomorrow morning

jaredpalmer:
Would be awesome to get someone from the TS team to review this. I’m not all up on what’s new yet in 2.8

jaredpalmer:
cc @weswigham

piotrwitek:
Thanks @weswigham that's a good suggestion!

@jaredpalmer
1) there was a small issue with `tsc-watch` which had duplicated typescript dependency (2.6.2) in yarn.lock, so I reinstalled and updated to clean it, works great now. (was breaking on conditional types)
2) Tomorrow I'll take a look at https://github.com/jaredpalmer/formik/pull/543/files#diff-18d1f112a814aca08509a74bded25fbfR25
I think it should be handled the same way as `touched` property type.

PS: Regarding the "function overload" issue you had mentioned if you have (or could create) a issue for it with some background I could take a look after this one is merged.

piotrwitek:
Hi @jaredpalmer,
I updated FormikErrors to conditional types the same as FormikTouched and added an additional set of type tests to prevent regression.

What do you think, ready to be merged?

jaredpalmer:
LGTM.

PR #550, opened 2018-04-01
fix Arrays example

saadq:
Old Version: https://codesandbox.io/s/jjlj9w5v2w
Fixed Version: https://codesandbox.io/s/0yj34y8pvn

---

The current code in the `Arrays.js` example currently fails to work. You can reproduce the error by doing one of the following:

* Remove a friend and then add a friend
* Add a friend and then remove a friend

This PR should fix this (and also get rid of the error about a missing `key`).

### DIFF
--- a/examples/Arrays.js
+++ b/examples/Arrays.js
@@ -27,7 +27,7 @@ const SignIn = () => (
               <div>
                 {values.friends.length > 0 &&
                   values.friends.map((friend, index) => (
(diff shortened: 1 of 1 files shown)


saadq:
The issue seems to go away without the need for also adding in a check for:

```js
touched.friends[index] &&
```

But let me know if you'd like me to include that as well.

PR #554, opened 2018-04-03
improved validateYupSchema types

piotrwitek:
Hi @jaredpalmer 
I have updated types for `validateYupSchema` function, especially removed `any` annotations and added proper Generic constrains.


jaredpalmer:
Nice!

Issue #548, opened 2018-03-29
FieldArray - arrayHelpers - add removeItem


elado:
Feature request

`FieldArray` `arrayHelpers.remove` gets an index. I'm suggesting it will also expose `removeItem/removeValue` that gets a value, and will remove all instances of that value in the array.

This helps when working with checkbox list, for example. Currently the only way is to grab the `indexOf(value)` and use `remove`.

https://codesandbox.io/s/o5pw1vx916

---

- Formik Version: 0.11.11

aashishnagpal:
@martavasconcelos @jaredpalmer

I guess somehow the code merge was missed. The related issue and PRs are closed but the code is not there anymore.

https://github.com/jaredpalmer/formik/blob/master/src/FieldArray.tsx - Doesn't contain removeValue and handleRemoveValue methods are missing entirely.

CC: @elado

PR #563, opened 2018-04-05
RemoveValue function added

martavasconcelos:
Based on issue 548 I added a function to remove an item from an array based on its value.

#548


browniefed:
Would be nice if this was removed based on index not by value.

jaredpalmer:
@browniefed that’s what `remove<T>(index: number): T` does right meow 🐱.

aashishnagpal:
@martavasconcelos @jaredpalmer 

I guess somehow the code merge was missed. The related issue and PRs are closed but the code is not there anymore.

https://github.com/jaredpalmer/formik/blob/master/src/FieldArray.tsx - Doesn't contain `removeValue` and `handleRemoveValue` methods are missing entirely.

PR #575, opened 2018-04-10
Remove duplicated interfaces

ohbarye:
Hello maintainers, thanks for the great library.

This PR removes some duplicated interfaces that I found while reading formik's code.

Those interfaces are added in https://github.com/jaredpalmer/formik/pull/141 and still remains, but they exist in `withFormik.ts` as well.

https://github.com/jaredpalmer/formik/blob/095e915ef18cdb557be09f01e5fa9e7ba2a6d19d/src/withFormik.tsx#L72-L82

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,19 +1,5 @@
 import * as React from 'react';
 
-export type CompositeComponent<P> =
-  | React.ComponentClass<P>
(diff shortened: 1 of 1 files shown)


ohbarye:
Thank you for your quick review.

PR #579, opened 2018-04-12
Add Field & FastField meta props

prichodko:
Dividing this PR https://github.com/jaredpalmer/formik/pull/476 into two parts.

PR adds meta object to Field bag to ease getting of different states. Currently meta object consists of:
* `error`
* `touched`
* `initialValue`

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -263,6 +263,7 @@ export class FastField<
     } = this.props as FieldConfig;
 
     const { formik } = this.context;
+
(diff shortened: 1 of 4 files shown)


bdox:
What's holding this up? The documentation mentions that in 1.0, a meta prop 'may' be passed to the <Field /> component... It is not.

charlax:
@jaredpalmer would you mind landing this? Any help needed here?

Issue #559, opened 2018-04-05
IE 11 Error: Object doesn't support property or method 'isNaN'


wkgalla:
## Bug, Feature, or Question?
Bug

## Current Behavior
Object doesn't support property or method 'isNaN' error in IE 11 when **input component type is number**.
And field value remains same.

## Desired Behavior
It should update value on typing new value

## Suggested Solutions
It might be due to this change https://github.com/jaredpalmer/formik/pull/437

- Formik Version: 0.11.11
- React Version: 15.6.2
- TypeScript Version: 2.6.2
- OS: windows 10
- Node Version: 8.4.0
- Package Manager and Version: npm 5.6

Ciaran0:
Having the same issue as described above on ie11

jaredpalmer:
Lets get a PR going

Ciaran0:
For anyone that needs a solution asap, to solve this I included the Number.isNan from [core-js](https://github.com/zloirock/core-js).
I like the lodash.isnan solution from @ohbarye

PR #584, opened 2018-04-16
Use lodash.isnan instead of Number.isNaN

ohbarye:
Hello, maintainers.

I'm trying to solve https://github.com/jaredpalmer/formik/issues/559 with this PR by adding `lodash.isnan` as a dependency since IE doesn't support `Number.isNaN`.

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

But I'm not sure whether this is the best approach to make polyfill for this project, so if you have a better solution in your mind I'll comply it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "hoist-non-react-statics": "^2.5.0",
     "lodash.clonedeep": "^4.5.0",
     "lodash.isequal": "4.5.0",
+    "lodash.isnan": "^3.0.2",
     "lodash.topath": "4.5.2",
     "prop-types": "^15.5.10",
     "warning": "^3.0.0"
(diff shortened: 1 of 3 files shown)


prichodko:
Libraries shouldn't include any polyfills. However, I wouldn't consider this change as adding polyfill and rather using better module for this operation. I am 👍 .

prichodko:
When you inspect the code, it's not really a polyfill, is it? It just uses a different technique to determine NaN. On the other hand, I understand the point.

vlad-zhukov:
There is no need to bring another dependency because isNaN check is a simple `value !== value` check. See #588 for an alternative to this PR.

prichodko:
TIL

ohbarye:
👍

PR #588, opened 2018-04-18
Add isNaN util function instead of Number.isNaN

### LINKED ISSUE #559: IE 11 Error: Object doesn't support property or method 'isNaN'

vlad-zhukov:
This is an alternative to #584.

Closes #584.
Fixes #559.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -6,6 +6,7 @@ import {
   isFunction,
   isPromise,
   isString,
+  isNaN,
   isEmptyChildren,
   setIn,
   setNestedObjectValues,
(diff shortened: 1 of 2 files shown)


jaredpalmer:
BOOM!

PR #590, opened 2018-04-19
Fix issue where storing Immutable values in form values causes browser hangs

bobbyrenwick:
# Bug
Having [Slate.js](https://github.com/ianstormtaylor/slate) `Value` objects in a nested `values` object leads to browser hangs.

# Current Behaviour
I have set up a `Field` for Slate.js editors that works in the same way that is demonstrated in the demo for integrating Draft.js - [Working with 3rd-party inputs #2: Draft.js](https://codesandbox.io/s/QW1rqjBLl).

The `initialValues` I send to the `Formik` component look like this:

```js
import { Value } from "slate"

const initialValues = {
    formSectionOne: {
        name: "Bob Renwick"
    },
    formSectionTwo: {
        bugReport: Value.fromJSON({}), // Actual value removed as irrelevant
    }
};
```
But the first time that Slate.js calls the `onChange` callback, and my `Field` updates the form via `setFieldValue("formSectionTwo.bugReport", change.value)` then Formik hangs at the `setIn` call because it calls `cloneDeep` on `values.formSectionTwo`, which then tries to clone the immutable object.

# Desired Behaviour
No hanging

# Solution
I replaced `cloneDeep` with a custom cloning technique to avoid deep cloning Immutable objects. This supports having Immutable object as leaves, anywhere in the values object, while retaining all the properties of the old implementation.

Formik Version: 0.11.11
React Version: 16.2.0
OS: OS X 10.11.6
Node Version: 9.11.1
Package Manager and Version: npm@5.6.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
   },
   "dependencies": {
     "hoist-non-react-statics": "^2.5.0",
-    "lodash.clonedeep": "^4.5.0",
     "lodash.topath": "4.5.2",
     "prop-types": "^15.5.10",
     "react-fast-compare": "^1.0.0",
@@ -52,7 +51,6 @@
     "@types/enzyme": "3.1.5",
     "@types/enzyme-adapter-react-16": "1.0.1",
     "@types/jest": "20.0.6",
-    "@types/lodash.clonedeep": "^4.5.3",
     "@types/lodash.topath": "4.5.3",
     "@types/prop-types": "^15.5.2",
     "@types/react": "16.0.28",
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import toPath from 'lodash.topath';
-import cloneDeep from 'lodash.clonedeep';
 
 /**
  * Deeply get a value from an object via it's path.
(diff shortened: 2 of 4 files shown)


bobbyrenwick:
Thanks for having a look at this PR so quickly. 

My reasoning behind splitting it into two functions is that the second one expects that the path is already split, so we can avoid the extra work to join the path that using a single function would necessitate, if the public API is to remain the same. I’m happy to merge it into a single function if you’re happy with the extra work?

I am happy to add extra test cases if you let me know what you’d like to test.

bobbyrenwick:
@jaredpalmer, let me know what you'd like to be done and I'll get right on it.

marek-baranowski:
Hi guys, what's the status on this? I have exactly the same issue with Slate's editor state.

marek-baranowski:
Hi @jaredpalmer, @bobbyrenwick,

What is the status on this? I have exactly the same issue with Slate's editor state causing browser to hang because of deep equality check.

Best Regards

bobbyrenwick:
@marek-baranowski I moved from formik to react-final-form and decided not to worry about implementing rich text editing for my use-case. It's something I'll probably come back to in the future but given that I've changed form libraries, I'm unlikely to be making any further changes to this PR.

A workaround that I was thinking about and never implemented was to store a serialized, string version of the form, and store that in formik, and then in the Field itself, convert that string to an immutable value to pass to Slate. I wasn't sure about the performance implications of doing that on every keystroke, but it could work.

bobbyrenwick:
@marek-baranowski, glad you got your issue sorted. It's performant enough to marshal into a string and back out again on every keystroke?

fc:
I'm seeing the same or similar problem except it is with `<FieldArray>`.  Changing a value for a `<FieldArray />` input causes about a 5 second delay.  With Chrome the performance profiler showed it spent about 99% of the time with `cloneDeep` which is called from `setIn`.  The odd part is I do not have any immutable objects like the others have described so not clear as to what is the root cause.

PR #591, opened 2018-04-21
Add flow support (in progress)

istarkov:
flow now can be easily distributed with package itself, without `flow-typed` (which has strange politics about types etc)

So here is  +-strict flow support for `Form` and `Formik`, see ./flow-test-examples/basic.js for details
It's some kind of flow test, each `$ExpectError` inside prevents from real error.

Now in progress, to going further I wanna ask one question.
To have a stricter Field type what if Field component can be passed as children render prop i.e
```javascript
<Formik>{({values, Field}) => ...</Formik>
```
In such case I could (I think ;-)) give to a Field a very good typings based on `Formik` types. So it would detect errors like `<Field name="unknownField"`, values types (as same as now `bind` works in basic.js)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,12 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[lints]
+
+[options]
+suppress_comment=\\(.\\|\n\\)*\\$ExpectError
+
+[strict]
--- a/package.json
+++ b/package.json
@@ -18,23 +18,25 @@
   "main": "dist/formik.js",
   "module": "dist/formik.es6.js",
   "typings": "dist/index.d.ts",
-  "files": [
-    "dist"
-  ],
+  "files": ["dist"],
   "scripts": {
     "test": "jest --env=jsdom && npm run size",
     "test:watch": "npm run test -- --watch",
-    "start": "cross-env NODE_ENV=development tsc-watch --project tsconfig.base.json  --onSuccess \"rollup -c\"",
+    "start":
+      "cross-env NODE_ENV=development tsc-watch --project tsconfig.base.json  --onSuccess \"rollup -c\"",
     "prebuild": "rimraf dist",
-    "build": "tsc -p tsconfig.base.json  && cross-env NODE_ENV=production rollup -c && cross-env NODE_ENV=development rollup -c && rimraf compiled",
+    "build":
+      "tsc -p tsconfig.base.json  && cross-env NODE_ENV=production rollup -c && cross-env NODE_ENV=development rollup -c && cp ./src/formik.js.flow ./dist/ && rimraf compiled",
     "prepublish": "npm run build",
-    "format": "prettier --trailing-comma es5 --single-quote --write 'src/**/*' 'test/**/*' 'README.md'",
+    "format":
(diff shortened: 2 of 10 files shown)


.flowconfig
```diff
@@ -0,0 +1,12 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[lints]
+
+[options]
+suppress_comment=\\(.\\|\n\\)*\\$ExpectError
```
TrySound: include_warnings=true

package.json
```diff
@@ -18,23 +18,25 @@
   "main": "dist/formik.js",
   "module": "dist/formik.es6.js",
   "typings": "dist/index.d.ts",
-  "files": [
-    "dist"
-  ],
+  "files": ["dist"],
```
TrySound: Do not format with prettier

package.json
```diff
@@ -18,23 +18,25 @@
...
+    "format":
+      "prettier --trailing-comma es5 --single-quote --write 'src/**/*' 'test/**/*' 'README.md'",
     "precommit": "lint-staged",
     "addc": "all-contributors add",
     "start-website": "cd website && yarn start",
     "build-website": "cd website && yarn install && yarn build",
     "gen-docs": "all-contributors generate && doctoc README.md",
-    "size": "size-limit"
+    "size": "size-limit",
+    "flowcheck": "yarn build && flow"
```
TrySound: `flow check` to not run the server. Also add `--max-warnings=0` flag to fail on any suppression.

src/formik.js.flow
```diff
@@ -0,0 +1,128 @@
...
+  validateOnBlur?: boolean,
+  /** Tell Formik if initial form values are valid or not on first render */
+  isInitialValid?: boolean,
+  /** Should Formik reset the form when new initialValues change */
+  enableReinitialize?: boolean,
+
+  onReset?: (values: Values, formikActions: FormikActions<Values>) => void,
+  onSubmit?: (values: Values, formikActions: FormikActions<Values>) => void,
+
+  render?: (props: FormikProps<Values>) => React.Node,
```
TrySound: Is this type recursive? You probably should use union: one with render and one with children like in react powerplug.

src/formik.js.flow
```diff
@@ -0,0 +1,128 @@
...
+
+export type ChildrenProps<Values> = {
+  values: Values,
+  isSubmitting: boolean,
+  dirty: boolean,
+  isValid: boolean,
+  errors: FormikErrors<Values>,
+  touched: FormikTouched<Values>,
+  ...$Exact<FormikActions<Values>>,
+  ...$Exact<FormikHandlers<Values>>,
```
TrySound: Use {||} on declaration instead please to make it more clear

jaredpalmer:
This is great. However, can this be added to flow-typed instead? I’m not that interested in maintaining these tbh

TrySound:
Supporting both type systems is a win for the project.

Good written flow definitions are usually do not introduce any effort, until you will change api. And still you may always ping us and we will take care of this.

istarkov:
Ill better shoot myself into a head than place anything on flowtyped again :-)

oonsamyi:
@istarkov Do you have actual flow definitions for Formik? I will be grateful if you send them to me.

PR #592, opened 2018-04-21
Prevent double build of esm and cjs bundles

TrySound:
In this diff I moved umd config into function and provided environment
as an argument.

Added universal external function to not bundle any dependency.

Also reused `main` and `module` fields from package.json in rollup config.


jaredpalmer:
Nice.

PR #593, opened 2018-04-22
Replace filesize plugin with size snapshot

TrySound:
Shows bundle, minified and gzipped sizes like `filesize`.

Ref #660

Also this plugin shows sizes of the bundles which imports nothing (I
call it treeshaked size) for both rollup and webpack.

So here we start tracking treeshakability.


.size-snapshot.json
```diff
@@ -0,0 +1,36 @@
+{
+  "./dist/formik.umd.production.js": {
+    "bundled": 180899,
+    "minified": 48286,
+    "gzipped": 14197
```
jaredpalmer: Do we have to commit this?

TrySound: Yes. This is the point. Every user can take a look at snapshot and get info about size and treeshakability.

TrySound:
@jaredpalmer So can we go with this?

jaredpalmer:
can this go in package.json?

TrySound:
Do you mean snapshot? I don't think it's a good idea to patch `package.json` automatically with anything except npm/yarn. This info can be too verbose in the future, so it's better to keep it in separate file.

PR #594, opened 2018-04-22
React 16.3

jaredpalmer:
This adds create-react-context to Formik and upgrades React to 16.3. It also exposes a few new helpers:

- `connect()` An HoC wrapper around `<FormikConsumer />`
- `<FormikConsumer />`
- `<FormikProvider />`


src/types.tsx
```diff
@@ -0,0 +1,252 @@
...
+ * State, handlers, and helpers made available to form component or render prop
+ * of <Formik/>.
+ */
+export type FormikProps<Values> = FormikSharedConfig &
+  FormikState<Values> &
+  FormikActions<Values> &
+  FormikHandlers &
+  FormikComputedProps<Values> & {
+    validate?: ((
+      values: Values
```
jaredpalmer: So in the past, we did not pass thru props or expose Formik's guts to general public. Formik props and context had slight differences. I'm not sure this is a good idea.

jaredpalmer: I wonder if this should just be: 

```tsx
export type FormikContext<Values> = FormikProps<Values> &
  Pick<FormikConfig<Values>, 'validate' | 'validationSchema'> & {
    registerField(name: string, resetFn: (nextValues?: any) => void): void;
    unregisterField(name: string): void;
  };
```

jaredpalmer: I guess we are using the word FormikContext to be the name of teh object returned by `create-react-context`. Technically we have never given this a name, it's just been typed as `any` for the most part. The only weird gotcha is that FastField needs to know about `<Formik>`'s props `validate` and `validationSchema` with current implementation.

TrySound:
I believe formatting `package.json` is not the best practice, so maybe it would be better to add it to `.prettierignore`?

src/connect.tsx
```diff
@@ -0,0 +1,25 @@
...
+import * as React from 'react';
+import { FormikProps } from './types';
+
+export const FormikContext = createContext<FormikProps<any>>({} as any);
+
+/**
+ * Connect any component to Formik context, and inject as a prop called `formik`;
+ * @param Comp React Component
+ */
+export function connect<Outer, Values = {}>(
```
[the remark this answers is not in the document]
jaredpalmer: Already expose withFormik. This is more low level

src/FastField.tsx
```diff
@@ -20,46 +21,34 @@ function isEqualExceptForKey(a: any, b: any, path: string) {
...
+    props.formik.registerField(props.name, this.reset);
   }
 
-  componentWillReceiveProps(
-    nextProps: Props,
-    nextContext: { formik: FormikProps<any> }
-  ) {
-    const nextFieldValue = getIn(nextContext.formik.values, nextProps.name);
-    const nextFieldError = getIn(nextContext.formik.errors, nextProps.name);
+  componentWillReceiveProps(nextProps: Props & { formik: FormikProps<any> }) {
```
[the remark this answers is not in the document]
jaredpalmer: Good call

jaredpalmer:
Can't update the tests yet because enzyme does not work yet with 16.3

See https://github.com/airbnb/enzyme/issues/1598

slorber:
great!

@jaredpalmer with FormikContext.Consumer does it mean i'll be able to inject formik context into my component? I want to build my own Inputs for ReactNative without needing to make the props wiring, as far as I understand it's what it enables

jaredpalmer:
Correct

PR #606, opened 2018-04-28
Provide more descriptive bundle extensions

TrySound:
- cjs for commonjs
- esm for es modules

Both are widely used this time and shows more clearer the purpose of one
or another bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "higher order component",
     "hoc"
   ],
-  "main": "dist/formik.js",
-  "module": "dist/formik.es6.js",
+  "main": "dist/formik.cjs.js",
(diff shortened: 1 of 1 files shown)


TrySound:
What's wrong with it?

jaredpalmer:
#592 broke the build because rollup externals were not handled correctly and were causing TS to include `tslib` in final output. Took me hours to figure this out. While working on 16.3.2 support today, i fixed the build #594 .

TrySound:
Why not include tslib in dependencies like we do with babel-runtime? A few libs could reuse it.

es6 is still the wrong name. For users it looks like untranspiled files. `esm` is more widely used these days and descriptive.

Issue #310, opened 2017-12-14
Compute withFormik's displayName from the wrapped component's displayName.


alisabzevari:
If you generate the `displayName` of the `withFormik` result from the wrapped component, this HOC can create nicer names in jest snapshots and react dev tools.

I think it can be easily done by adding the following line of code [here](https://github.com/jaredpalmer/formik/blob/1b02ae4b775864535e9f69e7a3903910e6712143/src/withFormik.tsx#L155):

``` js
C.displayName = `WithFormik(${Component.name || Component.displayName})`;
```

Would be my pleasure to create a PR for that :smiley:.

jaredpalmer:
Feel free to submit a PR

PR #609, opened 2018-05-02
fix withFormik display name

SimeonC:
Previously any debugger that uses displayName would report the component wrapped with `withFormik` as `C`. This change makes it such that debuggers will now resolve it as `withFormik(ComponentName)`.

Should fix the issue raised in #310 (~4 months is long enough to assume he's not going to raise the PR I think).


jaredpalmer:
Lgtm

PR #615, opened 2018-05-04
Warn against button with unspecified type

nathanforce:
Most browsers default `<button type="submit" />`.  This PR adds a development warning when  submitting a form via a button that does not have an explicitly specified `type` attribute.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -526,6 +526,25 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
     if (e && e.preventDefault) {
       e.preventDefault();
     }
+
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      document &&
+      document.activeElement
+    ) {
+      const type =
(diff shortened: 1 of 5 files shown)


jaredpalmer:
bitmoji

lookfirst:
@jaredpalmer If I hit 'enter' in my MUI-Next modal, I get the activeElement as being the focused field, not the button. It is unfortunate that you would accept a PR like this without any tests.

image

jmtoung:
@jaredpalmer  I also get the same error as @lookfirst when I hit 'enter'. This shouldn't have been accepted.

jaredpalmer:
Will revert tomorrow

jaredpalmer:
@lookfirst @jmtoung Can you please produce a codesandbox to demonstrate this issue? Please also upgrade to the latest version. We patched this to show the warning if and only if active element is an instance of a button.

lookfirst:
I've been running b2 which doesn't have this issue.

PR #616, opened 2018-05-07
Fix FormikErrors type

tetslee:
This supports a tree of strings, I'm not sure about i18n support.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -12,6 +12,13 @@ import {
   setNestedObjectValues,
 } from './utils';
 
+export type Extends<A, B> = A extends B ? true : false;
+export type Tree<Children, Leaf> = {
+  [K in keyof Children]?: Children[K] extends object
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Cc @weswingham is this better than what we currently have?

src/Formik.tsx
```diff
@@ -12,6 +12,13 @@ import {
   setNestedObjectValues,
 } from './utils';
 
+export type Extends<A, B> = A extends B ? true : false;
+export type Tree<Children, Leaf> = {
+  [K in keyof Children]?: Children[K] extends object
+    ? false extends Extends<Children[K], object> ? any : Tree<Children[K], Leaf>
```
tetslee: The extra check is to support when `Children` is `any`, because `any extends object ? A : B` is `A | B`.

tetslee:
Have updated PR to better support `any` type.

PR #620, opened 2018-05-08
feat: adds storybook

morgs32:
Got motivated this morning and updated an old PR that added storybook to formik [per your comment](https://github.com/jaredpalmer/formik/pull/402#issuecomment-376707401).
Should be useful. This creates a story from each example in `/examples`

Couple questions:
- Why do you have both an `examples/Basic.tsx` and a `.js` version?
- Do you need to copy the examples dir in the website? Can you use a relative path?


romainquellec:
👍

PR #622, opened 2018-05-08
feat: adds validateOnLoad

### LINKED ISSUE #288: Initial validation or initialErrors prop
What if the user has set some field values, and we do store them somewhere, and when the user returns to the app those values are invalid. Is there no way to run `validate` on initialization?

morgs32:
Resolves #288

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -31,6 +31,7 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
   static defaultProps = {
     validateOnChange: true,
     validateOnBlur: true,
+    validateOnLoad: false,
     isInitialValid: false,
     enableReinitialize: false,
   };
@@ -94,6 +95,9 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
 
   componentDidMount() {
     this.didMount = true;
+    if (this.props.validateOnLoad) {
+      this.runValidations(this.state.values);
+    }
   }
 
   componentWillUnmount() {
@@ -584,6 +588,7 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
       handleSubmit: this.handleSubmit,
       validateOnChange: this.props.validateOnChange,
       validateOnBlur: this.props.validateOnBlur,
+      validateOnLoad: this.props.validateOnLoad,
     };
   };
 
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -11,7 +11,9 @@ export interface FormikValues {
  * Should be always be and object of strings, but any is allowed to support i18n libraries.
  */
 export type FormikErrors<Values> = {
-  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : string
+  [K in keyof Values]?: Values[K] extends object
+    ? FormikErrors<Values[K]>
+    : string
 };
 
 /**
@@ -158,6 +160,8 @@ export interface FormikSharedConfig {
   validateOnChange?: boolean;
   /** Tells Formik to validate the form on each input's onBlur event */
   validateOnBlur?: boolean;
+  /** Tells Formik to validate the form in componentDidMount */
+  validateOnLoad?: boolean;
   /** Tell Formik if initial form values are valid or not on first render */
   isInitialValid?: boolean | ((props: object) => boolean | undefined);
   /** Should Formik reset the form when new initialValues change */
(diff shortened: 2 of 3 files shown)


morgs32:
Hi @jaredpalmer can we discuss before I clean up conflicts?

It appears that the react team stands behind calling setState from componentDidMount:
https://reactjs.org/docs/react-component.html#componentdidmount

And it's not a [valid eslint rule on latest react](https://github.com/yannickcr/eslint-plugin-react/issues/1754)

And as for the user experience, if the validation is synchronous, "the user won’t see the intermediate state" ([according to the first link](https://reactjs.org/docs/react-component.html#componentdidmount)).

If the validation is async, then presumably the validation logic implements some loading state UI. In which case it seems fitting that that loading UI would commence when the Form loads.

What do you think?

morgs32:
@jaredpalmer bump

jaredpalmer:
can you resolve conflicts?

morgs32:
@jaredpalmer how's this?

ianks:
While this is good, it doesn't solve the use case in #622 of havin server rendered errors passed in as props. Certain errors cannot be determined by client, so having the ability to pass in `initialErrors` is still crucial IMO

jaredpalmer:
I thought this was for initialErrors and initialTouched?

jpserrett:
The original post #288 mentioned either 'initialErrors' or a 'validateOnLoad' method. Both use cases could be argued for, and may be necessary, but this seems like a great addition that solves an issue a number of people are having. 

IMHO, I'd say that this and an 'initialErrors' prop would be two separate, and very useful, additions to the base code. In my use case, I have to make an API call to populate my initialValues, and that API call may return with my errors as well. So when the data is populated, I also need to run my custom validate method to retrieve the errors from the JSON response and apply them to the form. 

Just throwing in my $0.02.

morgs32:
@jaredpalmer bump

morgs32:
@jaredpalmer thoughts?

morgs32:
@jaredpalmer beginning to feel:
3x01_the_cabin_show_ 20

jpserrett:
Any status on this? Even if its no way, definitely not going to do it?

vedraan:
Is there some estimate on when this might get merged?

There are possible cases where some form fields are no longer valid and you should prevent saving (even though they were at the time of first submit). We need to resort to some hacks to get around this, and having this PR merged would solve all the problems.

hyperboreans:
@jaredpalmer @morgs32 we have a use case for this in production and would appreciate an update here. Ideally a merge, or at least a recap on pending issues that are blocking merge.

timini:
This looks like exactly what I'm after can we get it merged?

DVassilev:
@jaredpalmer any plans for merging this soon? Also intend to use this and would be great to fetch from npm. Thanks

cytGitHub:
哈哈哈




| |
chang
邮箱：ssh_github@163.com
|

Signature is customized by Netease Mail Master

On 11/15/2018 11:55, DVassilev wrote:

@jaredpalmer any plans for merging this soon? Also intend to use this and would be great to fetch from npm. Thanks

motiazu:
This is nice but wouldn't it be problematic without editing how `isValid` is calculated?
[This line](https://github.com/jaredpalmer/formik/pull/622) will still regard `isInitialValid` as the truth when the form state is equal to initialValues.

jaredpalmer:
In v2, validation is fired on mount by default.

motiazu:
@jaredpalmer Does this mean you will also deprecate `isInitialValid`? Otherwise, pristine forms will still ignore validation results as I stated in my previous comment in this thread.

src/Formik.tsx
```diff
@@ -31,6 +31,7 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
   static defaultProps = {
     validateOnChange: true,
     validateOnBlur: true,
+    validateOnLoad: false,
```
Andreyco: How about `validateOnMount`. It is not clear when "load" happens

jaredpalmer: Agree. Btw this is on by default in v2

morgs32: Is now a good time to use v2?

fernandopasik:
any news on merging this?

src/Formik.tsx
jaredpalmer: YES

jaredpalmer: validateOnMount makes the most sense

paillave: I am using v2.0.1-rc.12 but there is still not validateOnLoad flag. Any idea when this very important feature will be released?

Issue #558, opened 2018-04-05
Field validate and isValid


peterpetre:
## Bug, Feature, or Question?
Bug

## Current Behavior
isValid is false if errors: { email: undefined, password: undefined }

## Desired Behavior
isValid should be true if errors: { email: undefined, password: undefined }

## Suggested Solutions
other than checking only `this.state.errors && Object.keys(this.state.errors).length === 0`
should check `Object.values(this.state.errors).every(error => error === undefined)` too

## Additional Information
From docs - Field - validate : **Synchronous and if invalid, return a string containing the error message or return undefined.**

---

- Formik Version: 0.11.11
- React Version: 16.2.0
- TypeScript Version:
- CodeSandbox Link:
- OS: macOS High Sierra
- Node Version: 9.10.0
- Package Manager and Version: yarn 1.5.1

jaredpalmer:
Feel free to submit a PR.

peterpetre:
Sorry i dont have much time right now, even with the suggested solution i met more problems when using arrays or nested objects with field validate. So probably will need to deep check for undefined.

PR #624, opened 2018-05-11
Fix FastField.handleBlur leading to undefined errors and invalid forms

bobbyrenwick:
# Bug
Using `FastField` in a `Formik` form with `validateOnBlur` leads to forms being invalid when Fields are blurred without inputting anything.

# Current Behaviour
If you render the following form and then click into and out of the name field, you'll see that eventually `formikProps.errors` is logged to the console as `{ name: undefined }`. This means that when you have a form with validation you end up with the form being invalid just because someone touched a field.

```js
<Formik
  initialValues={{ name: "formik" }}
  render={(formikProps) =>  {
      console.log(formikProps.errors);
      return (
        <FastField name="name" type="text" />
      );
  }}
/>
```

# Desired Behaviour
`FastField.handleBlur` causes errors to be removed from `FormikBag.errors`, nested to any level, when `FastField.state.error === undefined`. 

# Solution
I added an `unsetIn` function that is similar to `unset` from `lodash`, but it needs to do cleanup along the way. If you have nested form values and the key you're unsetting is the final one in that object, then that field.name, error object pair also needs to be removed so that there are no empty objects in `FormikBag.errors`.

I'm not particularly happy with how this code reads, so any suggestions for changes to how it works would be most welcome!

Formik Version: 1.0.0-beta-1
React Version: 16.2.0
OS: OS X 10.11.6
Node Version: 9.11.1
Package Manager and Version: npm@5.6.0

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -6,7 +6,14 @@ import { FieldAttributes, FieldConfig, FieldProps } from './Field';
 import { validateYupSchema, yupToFormErrors } from './Formik';
 import { connect } from './connect';
 import { FormikContext } from './types';
-import { getIn, isEmptyChildren, isFunction, isPromise, setIn } from './utils';
+import {
+  getIn,
+  isEmptyChildren,
+  isFunction,
+  isPromise,
+  setIn,
+  unsetIn,
+} from './utils';
 
 export interface FastFieldState {
   value: any;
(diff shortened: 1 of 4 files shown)


ericbiewener:
Looks like this is a problem with non-FastField components as well: https://github.com/jaredpalmer/formik/issues/558

jaredpalmer:
Can we reopen this?

jaredpalmer:
Pinned for v2

jaredpalmer:
Pin

PR #628, opened 2018-05-12
chore: add Node.js 10

DanielRuf:


jaredpalmer:
Add node 9 instead?

DanielRuf:
> Add node 9 instead?

No, 8 is LTS and 10 is stable branch.

We have to update the affected dependency which uses `upath` (probably webpack according to the lock file but it is not in the package file).

DanielRuf:
Will try to resolve it.

DanielRuf:
`size-limit` is the root cause, we might just have to update the lockfile.

DanielRuf:
Success =) I knew I saw this before in other projects and it is just because of upath 1.0.4 =)

So this can be also merged now.

PR #644, opened 2018-05-18
Make errors behave as suggested in #150

### LINKED ISSUE #150: A way to keep backend errors with client validation
Here is a simple form with errors example:

```
name - no error
surname - no error
```

When I submit form, I receive backend errors: `{ surname: 'Exists' }`. So, I set field errors and now they are:

```
name - no error
surname - 'Exists'
```

Now, if I change or blur the value of the `name` field, the whole form validation will be triggered and backend error will be erased for `surname` field.

How is it possible to erase field's backend error only if the value of that field was changed?

I can think of possible solution with using form's `status` to save backend errors, checking the field's previous/new value, but it feels like a really big hack.

jf248:
Fixes #150 

I too wanted the `errors` prop to work as suggested in #150, i.e. don't lose submit errors for all fields when validate is executed from any one field's change/blur event. 

To do this I've created the, internal only, `state.validationErrrors` that is the new home of validation errors. `state.errors` now only contains errors from setErrors method.

At render `state.validationErrors` and `state.errors` are merged to create the provided `errors` prop.

Errors in `state.errors` are removed on a per field basis.


jaredpalmer:
I believe the way to do this might be the inverse of what you have. Keep api errors separate and merge them.

jf248:
Thanks, I'm going to submit another PR along the lines you suggest. It will expose new `apiErrors` and `setApiErrors` props and leave it for the user to merge `errors` and `apiErrors` if they wish.

(My thinking for this PR was trying to avoid new props.)

(Also re `setStatus` and `setErrors`: As discussed in #150 `setStatus` doesn't work well for api field-specific errors because they need to be cleared when the field's value changes. And all errors set with `setErrors` are lost on any field change/blur that calls `runValidations`.)

PR #645, opened 2018-05-19
Add apiErrors and setApiErrors

### LINKED ISSUE #150: A way to keep backend errors with client validation

jf248:
Fixes #150.

Exposes new `apiErrors` and `setApiErrors` props. `apiErrors` are automatically removed on a field-by-field basis with `handleChange` events and `setValues` / `setFieldValue` calls.

Users who want all errors in one prop could do:
```jsx
import { merge } from 'lodash';

<Formik {...props}>
  {(errors: validationErrors, apiErrors) => {
    const errors = merge(validationErrors, apiErrors);
    return (...);
  }
</Formik>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "dependencies": {
     "create-react-context": "^0.2.2",
     "hoist-non-react-statics": "^2.5.0",
+    "lodash": "^4.17.10",
     "lodash.clonedeep": "^4.5.0",
     "lodash.topath": "4.5.2",
     "prop-types": "^15.6.1",
@@ -54,6 +55,7 @@
     "@pisano/enzyme": "^3.3.0-pisano.public.1",
     "@pisano/enzyme-adapter-react-16": "^1.1.1-pisano.public.1",
     "@types/jest": "^22.2.3",
+    "@types/lodash": "^4.14.108",
     "@types/lodash.clonedeep": "^4.5.3",
     "@types/lodash.topath": "4.5.3",
     "@types/react": "16.0.28",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -45,6 +45,7 @@ const buildUmd = ({ env }) => ({
           'bool',
           'element',
         ],
+        'node_modules/lodash/lodash.js': ['omit'],
       },
     }),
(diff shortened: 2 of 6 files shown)


codepunkt:
I like the presented idea. Can't comment on the execution, but we have the same problem and would love a supported way to distinguish validation errors from api errors.

jaredpalmer:
Can you reopen?

jf248:
Sure, but can't see how to. `git push -f origin c4fb4ef:apiErrors`?

jaredpalmer:
Can’t reopen via button?

jf248:
Sorry if I'm missing something obvious. This is what I'm seeing:

screenshot from 2018-08-21 10-41-00

rostislav-simonik:
@jaredpalmer do you plan to integrate this PR, or is there alternative for that ? Thanks for info.

shamseerahammedm:
Hey any solution for this ? showing back end errors was successfull using setFieldError but as sson as some other field is changed error messages is removed

iplay88keys:
I worked around this issue by setting the "touched" state to false for all fields when the form is submitted. Then I set the error string for each field conditionally based on the presence of a formik error or the presence of an error I had set with `setFieldError` in my saga dependent on the "touched" state for that field. As such, api errors show for each field until those fields have been touched after the api errors have been initially set. 

Ex:
https://github.com/iplay88keys/my-recipe-library/blob/8c6dd42f9333e9ced5afd9d5b30f54032912f23b/ui/src/views/components/Registration.tsx#L33-L68

I would much prefer to have this in the library rather than a workaround, but this works for now. Let me know if you have additional questions.

PR #646, opened 2018-05-20
utils setIn() is cloning too much and kills PureComponent

jzimmek:
I am using formik in a rather complex and nested form application and we lately ran into some performance issues.

Some of our ```render()``` props of Field/FieldArray are containing lots of other components. Every change rerenders all of them down the tree. We tried to wrap those component inside ```render()``` with PureComponent to minimize performance impact.

But it does not help, because the current implementation of ```setIn``` is cloning object keys even if it is not necessary. This kills PureComponent.

I added a failing test case in this pull request.

This pull request is not to be merged (not yet) - but rather to discuss if this is expected behavior or a bug which should be addressed?

### DIFF
--- a/test/utils.test.tsx
+++ b/test/utils.test.tsx
@@ -197,6 +197,13 @@ describe('utils', () => {
       expect(obj).toEqual({ x: 'y' });
       expect(newObj).toEqual({ x: 'y', a: { x: { c: 'value' } } });
     });
+
+    it('does not clone unaffected keys of the object', () => {
+      const obj = { x: { name: 'n1', arr: [1] } };
+      const newObj = setIn(obj, 'x.name', 'n2');
+
+      expect(newObj.x.arr).toBe(obj.x.arr);
+    });
   });
 
   describe('isPromise', () => {


jzimmek:
Just tested to rewrite the ```setIn()``` by using the fp version of ```lodash set()``` like this:

```
export function setIn(obj: any, path: string, value: any): any {
  return _.set(path, value, obj);
}
```

Works as expected and all tests (including the one from this pull request) pass.

Unfortunatly rollup is not liking lodash out of the box and would need further setup to keep bundle size small. Never worked with rollup before, but rollup-plugin-babel + babel-plugin-lodash might work.

@jaredpalmer Is this heading into the right direction? Would love to hear your thoughts about this.

jaredpalmer:
If we can avoid another lodash drop that would be preferred

jaredpalmer:
Feel free to continue on this PR

hmaurer:
Where are we on this? I have also encountered performance issues on medium-sized forms (20-ish fields) and FastField doesn't quite behave the way I want it to.

hmaurer:
Thank you @ttrushin ! I'll try this out

jaredpalmer:
@jzimmek lets fix this.

Issue #219, opened 2017-10-25
[Proposal] — Leverage DOM API for validations


everdimension:
Hi! I like your library and I believe it's the one that chose the right direction.

There's one thing I'm missing, though. Being able to determine form validity based on native DOM API. Many people don't know how powerful it is, and others usually think that its support is not good, when in fact [it's great](http://caniuse.com/#feat=form-validation).

I understand that your library's intention is to support react native as well and therefore you're trying to be dom-agnostic. But really you're not. Hence the differences like `onChange` prop vs the `setFieldValue` prop. You cannot absolutely abstract away the differences between web and native and I think everyone agrees that it is actually okay.

Therefore I propose that you provide somewhere access to either the input dom nodes (or just the inputs' [`ValidityState`](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) object, as in `input.validity`) or maybe just to the `form` node itself.

The latter would perhaps be the least invasive solution, and with the reference to the `form` node we can actually access any input object as well.

What do you think? I'd be glad to help.

jaredpalmer:
Can you give an example of a form where you need this ?

Feel free to submit a PR /make a demo sandbox. This sounds interesting

everdimension:
Well, I don't think there can be one particular example, it's just that in general I'd say that native forms validation functionality it's both powerful and robust.

1. But alright, here's one example of a common pattern where the user is asked to "confirm" (e.g. repeat) an entered value (usually a password). We can use the `pattern` attribute and simply pass the entered value of one input as a pattern to another. A very short code snippet to give you an idea
    ```jsx
    <form>
      <input type="text" value={this.state.value} required />
      <input type="text" pattern={this.state.value} required />
    </form>
    ```
    [Here's an actual working demo](https://codepen.io/everdimension/pen/BmywVY?editors=0010).

    Obviously similar result can be achieved with the current formik api where we would just compare two values.
But there are two benefits in my example:
    1. You don't have to validate by checking individual fields, you just read the `checkValidity()` method of the form itself.
    2. We can avoid giving a `name` attribute to the second input. While I believe that in general you should always give names to your inputs, this particular type of input doesn't provide any actual information and it's not part of the data being sent, it's just a client-side validation for making sure that the user entered something correctly. If we give it a `name` attribute so that it becomes part of the `values` state, we'll to do additional work of omitting this value before sending it to the server.


[...]

PR #650, opened 2018-05-21
[WIP] Use constraint validation API

everdimension:
Hi!

I'm getting back to you regarding this issue: https://github.com/jaredpalmer/formik/issues/219. Sorry for not doing it earlier!

After some thought and analyzing how you run validations in formik, I decided that perhaps all that we need to leverage constraint validation API is to use the `input.validationMessage` property. It does exactly what formik expects:

* returns a non-empty string if the value does not pass the validation constraints
* returns an empty string otherwise

For now, basically all I did was add a `useDomValidation` prop to the `<Field />` component and conditionally call the `setFieldError` fn with the result of `input.validationMessage`.

I'd like to consult with you whether you think this is a proper direction.
There would be things to consider:

* Should be able to pass the same prop to the `<Formik />` component to apply this logic to all fields?

* Is the result of `validationMessage` what we want to put in the `errors` field? Currently this message is constructed by the browser (or whatever the user puts in the `input.setCustomValidity('error msg')`). We might want to instead allow the user to map from [`ValidityState`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-constraint-validation-api) properties to his own messages, e.g.:
`{ valueMissing: () => 'You forgot to fill in this field' }`

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -65,6 +65,8 @@ export interface FieldConfig {
    */
   validate?: ((value: any) => string | Function | Promise<void> | undefined);
 
+  useDomValidation?: boolean;
+
   /**
    * Field name
    */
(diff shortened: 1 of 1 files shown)


everdimension:
I'd love to know whether this feature is wanted and whether I should continue making progress on it. No pressure though! :)

jaredpalmer:
Hey! This bot is working as intended!

I think this is a great situation to write your own <Field> implementation using `connect()` or wrapper class around `<Field>`.

PR #651, opened 2018-05-21
Update yup import

jquense:
bring docs here inline with latest `yup` version which removed the default export in favor of named ones


jaredpalmer:
LGTM.

PR #653, opened 2018-05-22
fix undefined var in react native

hosmelq:
This fixes an error in RN where `document` is not defined.


slorber:
@jaredpalmer would you mind publishing a beta3 soon? currently beta2 handleSubmit is unusable in RN, unless using this workaround: `onPress={() => this.formikRef.submitForm()}`

slorber:
actually there is `submitForm` in the formikBag, so this is more a doc problem

Issue #643, opened 2018-05-16
Why do you collect code in a single bundle?


retyui:
**Is your feature request related to a problem? Please describe.**
I try reduce my bundels size. In my project I use one import:
```js
import { withFormik } from 'formik';
```
I would like to have the opportunity to do so:
```js
import withFormik from 'formik/withFormik';
```

**Describe the solution you'd like**
Why do you collect code in a single js bundle?
Use flatten struct!
Like as: `lodash-es`, `date-fns`, `react-router`, `ramda`, `recompose`

retyui:
@jaredpalmer  do you mean `cp package.json dist/package.json && cd dist && npm run build`

jaredpalmer:
Correct

PR #660, opened 2018-05-29
Added ESM build

retyui:
https://github.com/jaredpalmer/formik/issues/643

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 dist
+esm
 compiled
 *.log
 coverage
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   "module": "dist/formik.es6.js",
   "typings": "dist/index.d.ts",
   "files": [
-    "dist"
+    "dist",
+    "esm"
   ],
   "peerDependencies": {
     "react": ">=15"
@@ -29,7 +30,11 @@
     "test:watch": "npm run test -- --watch",
     "start": "cp ./index.js ./dist/index.js && cross-env NODE_ENV=development rollup -w -c | tsc -w -p tsconfig.base.json",
     "prebuild": "rimraf dist",
-    "build": "cross-env NODE_ENV=production tsc -p tsconfig.base.json  && rollup -c && rimraf compiled && cp ./index.js ./dist/index.js",
+    "build": "npm-run-all --parallel build:*",
+    "build:esm": "cross-env NODE_ENV=production tsc -p tsconfig.esm.json",
+    "prebuild:esm": "rimraf esm",
+    "build:umd": "cross-env NODE_ENV=production tsc -p tsconfig.base.json  && rollup -c && cp ./index.js ./dist/index.js",
+    "postbuild:umd": "rimraf compiled",
     "prepublish": "npm run build",
     "format": "prettier --trailing-comma es5 --single-quote --write 'src/**/*' 'test/**/*' 'README.md'",
     "precommit": "lint-staged",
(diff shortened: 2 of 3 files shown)


retyui:
@TrySound 
I attach a link to the issues https://github.com/jaredpalmer/formik/issues/643

------------------------------------------------------------------------------------------
**Is your feature request related to a problem? Please describe.**
I try reduce my bundels size. In my project I use one import:
```js
import { withFormik } from 'formik';
```
I would like to have the opportunity to do so:
```js
import withFormik from 'formik/esm/withFormik';
```

**Describe the solution you'd like**
Why do you collect code in a single js bundle?
Use flatten struct!
Like as: `lodash-es`, `date-fns`, `react-router`, `ramda`, `recompose`

retyui:
@jaredpalmer help me!

TrySound:
This is actually a bad solution which creates the problem of large import statements sections. It's better to provide better treeshakability. By the way recompose is gonna remove path imports.

retyui:
Example:
https://unpkg.com/react-redux@5.0.7/
This module is very user friendly.
if I do not need a bundle, I will use the import from the `dist` folder
if I need the cmj version I'm doing the import from the `lib` folder
if I need the version I can use the import from the `es` or `src` folder

Now for some reason you decided to represent the entire bundle.

TrySound:
The problem is that you should choose the way to import. But you don't need to choose. Just use single entry point and do not care about package structure.

But the way that package is not friendly. The import is not universal you should choose also cjs or esm. If you choose esm then you cannot run this code in node without transpiling node_modules and if you choose cjs your bundle will be bigger.

Use single entry point and allow bundler make all these decisions for you. If you care about bundle size then help to improve treeshakability. I created a tool for tracking (https://github.com/TrySound/rollup-plugin-size-snapshot).

retyui:
@TrySound 
> This is actually a bad solution which creates the problem of large import statements sections

In our case, there is no such problem!

> If you care about bundle size then help to improve treeshakability. 

I use webpack 4 and tree shaking doesn't work there.
 
Who does not care about the size of the original files, let them use your single entry point.

I have a rigid frame for file sizes. And I already chose the import method. Only formik does not support this method. Why do you oppose?

TrySound:
@retyui Treeshaking works if library code and configuration doesn't suck. The problem of this library is top level function calls and transpiled classes without pure annotations (is the a config option or TS sucks here?). Prefering babel over typescript could solve the problem.

retyui:
I created a test repository, tell me please who sucks the configuration or module?
https://github.com/retyui/test

TrySound:
I mean config in formik build process, it's not user land problem. The point is that untranspiled code is treeshakable enough, some cases can be handled with pure annotations. Transpiled code has problems which can be solved with the right configuration.

retyui:
@TrySound 
If not our problem - Then I suggest you now in the lodash-es repository
and create an issue, about not need to provide untranspiled code 
and that the functionality of this library would be in one file

TrySound:
lodash suck because of default export which has a bunch of effects. Simple reexports will works perfectly.
https://unpkg.com/lodash-es@4.17.10/lodash.default.js

retyui:
@TrySound It will take a lot of time? Can I help?

jaredpalmer:
I would prefer to stay with TypeScript. We are going to be updating stuff for TS 2.9 as soon as it drops

TrySound:
@jaredpalmer Well is there a way to disable transpiling class properties in typescript?

TrySound:
Yeah, I use it too.

TrySound:
There's some problem with ts generated code elimination. Trying to figure out what he problem.

klis87:
For a proper webpack tree shaking, pls also add `"sideEffects": false` to `package.json` - I assume there is no sideeffect in this lib so reexports are safe not to include them when they are not imported

TrySound:
Waiting this for merge so we could prove treeshakability.
https://github.com/jaredpalmer/formik/pull/593

PR #674, opened 2018-06-04
Fix #673. Fix fastfield using cDU

### LINKED ISSUE #673: FastField "field.value" doesn't get typed value
**Describe the bug**

The FastField component seems to not be passing the the keyboard typed value into the `field.value` injected prop of a custom input component, since this prop is passed down to the inner input of the field, the typed text dosen't get rendered (this happens too if we don't use a custom input component).

**To Reproduce**

Checkout the following sandbox and start typing in the input.

[Edit Formik and Material-ui](https://codesandbox.io/s/1478jo353), 

**Expected behavior**

The input should display all typed characters.

**Suggested solution(s)**

I tracked back some earlier versions and it looks like the issue was introduced at version 1.0.0-alpha.r16.3.1

**Additional context**

I'm still in the process on adopting the library, i'm trying to integrate FastField with Material-UI's TextField, I was able to integrate the normal Field component with Material-UI without any issue.

---

- Formik Version: v1.0.0-beta.2
- React Version: 16.3.2
- TypeScript Version: N/A
- CodeSandbox Link: [here](https://codesandbox.io/embed/1478jo353)
- OS: Windows 10
- Node Version: N/A
- Package Manager and Version: N/A

jaredpalmer:
I (think) this fixes `<FastField>` on 16.4. I wasn't sure if we should do this in `componentDidUpdate` or if we should do this by storing old props on state and use `getDerivedStateFromProps`

https://reactjs.org/blog/2018/05/23/react-v-16-4.html#2-compare-incoming-props-to-previous-props-when-computing-controlled-values

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -26,8 +26,6 @@ class FastFieldInner<Props = {}, Values = {}> extends React.Component<
   FieldAttributes<Props> & { formik: FormikContext<Values> },
   FastFieldState
 > {
-  reset: (nextValues?: any) => void;
-
   constructor(
     props: FieldAttributes<Props> & { formik: FormikContext<Values> }
   ) {
@@ -36,15 +34,6 @@ class FastFieldInner<Props = {}, Values = {}> extends React.Component<
       value: getIn(props.formik.values, props.name),
       error: getIn(props.formik.errors, props.name),
     };
-
-    this.reset = (nextValues?: any) => {
-      this.setState({
-        value: getIn(nextValues, props.name),
-        error: getIn(props.formik.errors, props.name),
-      });
-    };
-
-    props.formik.registerField(props.name, this.reset);
 
     const { render, children, component } = props;
 
@@ -80,10 +69,6 @@ class FastFieldInner<Props = {}, Values = {}> extends React.Component<
     if (!isEqual(nextFieldError, prevFieldError)) {
       this.setState({ error: nextFieldError });
     }
-  }
-
-  componentWillUnmount() {
-    this.props.formik.unregisterField(this.props.name);
   }
 
   handleChange = (e: React.ChangeEvent<any>) => {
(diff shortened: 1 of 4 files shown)


prichodko:
Random thought:

Does it mean that if parent for some reason re-renders with different `value`, your `value` (in a state) would be overwritten? Can it even happen in inputs? 🤔

prichodko:
Why do we use this then - https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L422? If there is a check inside the `FastField` component, which should reset it when parent updates state with new values.

jaredpalmer:
I think you are right. I don't think we need the resetFn anymore.

jaredpalmer:
I wonder if we should move fastfield to just use sCU instead of trying to sync up local state.

prichodko:
What if regular `Field` implemented sCU?

prichodko:
If it would make sense, I can create a PR - do we have some project to test it out?

jaredpalmer:
i can test it on a big app. I think FastField should implement sCU the more i think about it.

prichodko:
Do you think regular `<Field />` with SCU could be as performant as `<FastField />` ?

jaredpalmer:
After this release I think I am going to move FastField to use sCU.

vlad-zhukov:
There is also no need to use react-lifecycles-compat for this component!

AaronW9090:
>Do you think regular <Field /> with SCU could be as performant as <FastField /> ?

This is true except that the issue isn't just Fields that are being updated, it's everything that Formik wraps. This could be an arbitrarily deep component tree that is re-rendering for no reason. It's especially a problem when using the withFormik HOC because it's less clear what is being controlled by Formik.

No reason not to implement sCU into Field anyway but it doesn't completely solve the performance issue.

PR #679, opened 2018-06-06
Bump to TS 2.9.1

jaredpalmer:
Not 100% sure, but looks we got bit by the breaking TS change related to string / symbols key values?

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "tsc-watch": "^1.0.21",
     "tslint": "5.5.0",
     "tslint-react": "3.2.0",
-    "typescript": "^2.9.1",
+    "typescript": "2.9.1",
     "yup": "0.21.3"
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "tsc-watch": "^1.0.21",
     "tslint": "5.5.0",
     "tslint-react": "3.2.0",
-    "typescript": "2.8.3",
+    "typescript": "^2.9.1",
     "yup": "0.21.3"
   },
   "lint-staged": {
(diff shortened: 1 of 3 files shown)


jaredpalmer:
@weswigham how do we fix this?

https://travis-ci.org/jaredpalmer/formik/jobs/388978234#L475

weswigham:
Either use `Extract<keyof Values, string>` to limit to only the string values, or accept `string | number | symbol` (or shorter and backcompat for before `keyof` handled symbols or numbers: `keyof any`) on the other side.

package.json
```diff
@@ -83,7 +83,7 @@
     "tsc-watch": "^1.0.21",
     "tslint": "5.5.0",
     "tslint-react": "3.2.0",
-    "typescript": "2.8.3",
+    "typescript": "^2.9.1",
```
lookfirst: Why move to `^`?

jaredpalmer: Good catch

src/types.tsx
```diff
@@ -78,21 +78,10 @@ export interface FormikActions<Values> {
   /** Manually set values object  */
   setValues(values: Values): void;
   /** Set value of form field directly */
-  setFieldValue(
-    field: keyof Values,
```
jaredpalmer: @weswigham I don't even think we retained type safety here because of the overload. Also AFAIK, can't really use `Extract` either because `setFieldValue` accepts dot path string like `social.facebook`.

weswigham: Given there's a `setFieldValue(field: string, value: any, shouldValidate?: boolean): void;` overload below it, I just assumed this overload existed to give completions for the known top-level fields.

jaredpalmer: Oh you are totally right. I forgot that's what happened. 

So, just to be clear, `Extract<Values, string>` is not backwards compat?  hrmmm.

weswigham: You can say `keyof Values & string`

Issue #659, opened 2018-05-26
FastField does not register user input in React 16.4


nzacca:
**Describe the bug**
`FastField` does not register the changed value when entering text into an input.

**To Reproduce**
https://codesandbox.io/s/o52powyy69

**Expected behavior**
As a user, when I enter text into the textbox, the text should appear in the box.

**Suggested solution(s)**

**Additional context**
This appears to be related to the change in `getDerivedStateFromProps` where now it is triggered on every render. https://reactjs.org/blog/2018/05/23/react-v-16-4.html#bugfix-for-getderivedstatefromprops

- Formik Version: 1.0.0-beta.2
- React Version: 16.4.0
- TypeScript Version: 2.8.3
- CodeSandbox Link: https://codesandbox.io/s/o52powyy69
- OS: N/A
- Node Version: N/A
- Package Manager and Version: N/A

eugene1g:
Confirming the same behavior sine updating for 16.4. 
This means that fields build on `FastField` become read-only with the current version of React.

/cc @jaredpalmer as this is quite a serious thing :)

nzacca:
Just wanted to add to @eugene1g comment above - the typed value is being written to the field but it is being overwritten/cleared after `getDerivedStateFromProps` executes.

jaredpalmer:
Can someone submit a PR?

eugene1g:
*crickets*

I think the behavior of `FastField` will be highly complex because it hits against two of the trickiest parts of API: choosing `props` vs `state`, and syncing uncontrolled components to the master state. 

The naive PR would be to pick the local state value, but that would introduce a new subtle bug that if the value is updated elsewhere in Formik, the input text shows in FastField will be out of date. A less naive PR would capture the original passed value, and do a 3-way comparison between "original / state / latest prop", but then we're getting into conflict resolution. 

(related discussions: https://github.com/facebook/react/issues/12898#issuecomment-392117539 and https://github.com/facebook/react/issues/12898#issuecomment-392035163 )


As FastField is such a core component, and any changes to it could be volatile, I see this as an architecture-level decision/PR and I definitely don't understand Formik goals/roadmap well enough to make a useful suggestion.

prichodko:
[...]

PR #684, opened 2018-06-07
Fix #483. Run Field validate() prior to submit

### LINKED ISSUE #483: Field#validate not running on form submit

jaredpalmer:
This is half of the PR to fix #483. Still need to reimplement the pre-submit behavior if `validationSchema` is present.

This code is gross. Comments welcome.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -19,6 +19,7 @@ import {
   isString,
   setIn,
   setNestedObjectValues,
+  getIn,
 } from './utils';
 
 export class Formik<ExtraProps = {}, Values = object> extends React.Component<
@@ -40,7 +41,12 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
   hbCache: {
     [key: string]: (e: any) => void;
   } = {};
-  fields: { [field: string]: (nextValues?: any) => void };
+  fields: {
+    [field: string]: {
+      reset?: ((nextValues?: any) => void);
+      validate?: ((value: any) => string | Promise<void> | undefined);
+    };
+  };
 
   constructor(props: FormikConfig<Values> & ExtraProps) {
     super(props);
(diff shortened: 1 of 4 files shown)


src/Formik.tsx
```diff
@@ -311,14 +323,36 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
...
         );
-        return;
+        Promise.all([
+          maybePromisedErrors.then(x => x, e => e),
+          ...fieldKeysWithValidation.map(
+            f =>
+              Promise.resolve<string | undefined | PromiseLike<any>>(
+                this.fields[f].validate!(getIn(this.state.values, f))
+              ).then(x => x, e => e) // always catch
+          ),
```
jaredpalmer: Help! Someone make this better!

src/types.tsx
```diff
@@ -226,7 +226,15 @@ export type FormikProps<Values> = FormikSharedConfig &
   FormikActions<Values> &
   FormikHandlers &
   FormikComputedProps<Values> & {
-    registerField(name: string, resetFn: ((nextValues?: any) => void)): void;
+    registerField(
+      name: string,
+      fns: {
+        reset?: ((nextValues?: any) => void);
```
prichodko: Reset is not used anymore.

prichodko:
Will try to help later today. We will see if this (https://github.com/jaredpalmer/formik/issues/659#issuecomment-394201172) will work somehow better.

jaredpalmer:
@prichodko any progress on that codesandbox? 

Also going to make you a contrib right now.

jaredpalmer:
So I think this is breaking change because Field now requires register fns to test? idk.

jaredpalmer:
@slightlytyler I had to make submitForm async to even test this one

jaredpalmer:
I am going to cut an experimental release with this in it, so we can run on our apps.

Andreyco:
@jaredpalmer do you think this might cause issues with `this.fields` going out of sync when `Field` is unmounted? I guess `unregisterField` does not cause parent's update.

gsaandy:
What about sending the mounted `fields` with submit form? it can be used for saving only those field values

Issue #389, opened 2018-01-30
Track warnings like errors


jonathanj:
Currently a field is either invalid or valid. In most form-heavy applications there is a grey area where a value might be outside of some "known good" parameter but not necessarily invalid, resulting in a warning that doesn't affect form-validity but is still composed of validation functions. Extending this thinking, it's possible that fields might have a number of meta states (with errors and warnings being just two) that can be built out of validators. One such (real-world) example is a "visibility" state: Is some condition met (or not met)? Then this field (or group of fields) should have their visibility affected.

I filed a [similar issue](https://github.com/erikras/redux-form/issues/2558) for redux-form (which already has warnings) a while ago but have since shied away from redux-form.

I think this could be implemented in a future proof way (while retaining backwards compatibility) by introducing a new setter function and a new render prop (an object mapping validation type keys `error`, `warning`, etc. to objects of validation results).

jaredpalmer:
Can you give me an example of a warning? I have yet to truly see one in the wild.

prichodko:
I think two base states are really `touched` and `error`. Everything else (e.g. visibility based on some condition) could be achieved with `status`.

jonathanj:
@jaredpalmer In our form-heavy application we have cases where it's important we capture as much valid data upfront as possible but we're hesitant to outright declare some user input as invalid because the rules would be arbitrary ones made up by ourselves ("How long have you lived here?") and could prevent valid (but perhaps extreme) input from being submitted, a warning deals with "fat finger" cases like these quite well in my experience.

@prichodko Status sounds like it can only have one value which seems somewhat limiting? Is my assumption incorrect?

cyoung-mspark:
I have an app form with a select list and one of the form options requires some additional notifications to the user. We were using redux-form's warning state to highlight a reminder to the user that selecting the option requires that they understand their choice. We could have handled it with an event listener but it was a nice way of putting that message in the form input feedback area.

donysukardi:
Here's an example of a warning from our application 
![image](https://user-images.githubusercontent.com/410792/40840842-eb485ac4-65da-11e8-95f4-12ba01e4ce70.png)

bytasv:
@jaredpalmer any updates on this? I see PR is up and author has added tests. Just started using Formik and bumped into issue of missing warnings as they are needed for my usecase :)

volodymyr-kushnir:
I need something like this as well.
[...]

PR #685, opened 2018-06-08
Add support for form warnings

donysukardi:
Related Issue: https://github.com/jaredpalmer/formik/issues/389

This is pretty much copy and paste of error handling that's currently in Formik, except that warnings should not be blocking.


jaredpalmer:
Please add tests.

donysukardi:
I've added tests

minipai:
👍

src/withFormik.tsx
```diff
@@ -65,6 +65,12 @@ export interface WithFormikConfig<
    * throws an error object where that object keys map to corresponding value.
    */
   validate?: (values: Values, props: Props) => void | object | Promise<any>;
+
+  /**
+   * Warning function. Must return an error object or promise that
+   * throws an warning object where that object keys map to corresponding value.
+   */
+  warn?: (values: Values, props: Props) => void | object | Promise<any>;
```
pmeller: IMHO it should be also possible to compute warnings using Yup Schema by passing `warningsSchema` parameter to keep the API consistent.

Issue #561, opened 2018-04-05
Pass props through FieldArray


kserjey:
## Feature

I have a view component for rendering an array of fields.

```js
function createFormikArray({ title, initialValue }, renderItem) {
  return function FormikArray({ name, form, ...arrayHepers }) {
    return (
      <React.Fragment>
        {/* view that uses title, initialValue and renderItem props */}
      </React.Fragment>
  }
}
```

It would be very useful if I could do smth like that:

```js
<FieldArray
  name='users'
  component={FormikArray} 
  title='My Field' 
  initialValue={{ firstname: '', lastname: '' }}
  {/* and so on */}
/>
```

kserjey:
Hey, anybody there?

jaredpalmer:
You can use a closure with `render` (just like React Router)

jaredpalmer:
Passing through props gets messy. The only reason Formik does this with Field is beacause Field is the only component which _can_ take care of rendering for you. Every other component is just logic. Hope that gives some context.

charlax:
The nice thing about using `component` vs. `render` is that `form` is passed as well... If you use `render` then you have to pass it manually (logical).

Why is it messy? `react-router`, `material-ui` and others do this. Just asking out of curiosity. I know it adds a bit more magic to what props are passed, but then nested `render` get pretty quickly painful to read too...

plxel:
Also Field component passes props to underlying field Component
```
<Field
    component="textarea"
    name={`somename.${index}`}
    className="style"
    placeholder="placeholder"
/>
```
So I thought this will be ok for using the same logic for FieldArray, but it is not... it is really confusing

cdpautsch:
Any update on this? This is a feature that would be useful. We are passing a custom component to `FieldArray`, and passing props via `FieldArray` to our custom component allows us to adjust what renders. I see there's an outstanding PR for it, but it hasn't been updated since March.

PR #693, opened 2018-06-13
Pass props down in FieldArray

joshsalverda:
Right now custom props aren't being passed down to the component when using:

`<FieldArray component={CustomComponent} customProp={beep}/>`

It works as expected already with Field, so extending the same functionality to FieldArray.

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -260,9 +260,10 @@ class FieldArrayInner<Values = {}> extends React.Component<
         validationSchema: _validationSchema,
         ...restOfFormik
       },
+      ...rest
     } = this.props;
 
-    const props = { ...arrayHelpers, form: restOfFormik, name };
+    const props = { ...arrayHelpers, form: restOfFormik, name, ...rest };
 
     return component
       ? React.createElement(component as any, props)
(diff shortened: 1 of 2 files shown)


joshsalverda:
After looking through the tests and re-reading the documentation around the component prop, it seems like passing extra props down through Field may have been an unintended side effect of some other change... Nothing specifically calls out that functionality. If that's the case let me know, I can always switch to using render instead (which I'm doing now in the meantime).

kserjey:
I'm also interested in it, but in #561 it was declined...
I think that maybe there is a better way to solve it. Just pass all custom props to `props`, like it is done in `withFormik`.

joshsalverda:
Thanks for the history @kserjey 

I can see where Jared is coming from with his point about the rendering... I'm fine with using a render prop anyways in my use case, I don't need anything fancy. Funny that Jared mentioned React Router, since I don't like that they don't pass props down either 😆

jaredpalmer:
Can you reopen?

joshsalverda:
Hi @jaredpalmer, any reason to reopen this? I had actually forgotten about it 😅Unless you've changed your mind about supporting this feature I would say it's fine to keep closed.

Also Formik is awesome, thanks for building and actively supporting it.

jwallet:
I'm wondering why this is not the default behavior when using component. I'm expecting that the parent will pass down its props to its component child, like Field and FastField do (maybe unintended) but I still prefer to use `component` prop instead of `render`. Will like to see it happens

PR #708, opened 2018-06-26
Add initialErrors to Formik and mapPropsToErrors to withFormik

yanchith:
UPDATED

This PR adds:

- `initialErrors` to `<Formik />`
- `mapPropsToErrors` to `withFormik`

It also extends tests around `initialValues` and `enableReinitialize` to handle `initialErrors`.

Interactions with other features:

- `isInitialValid` could be inferred from `initialErrors` if not provided explicitly, but the default `false` probably makes more sense with or without `initialErrors` so I left it as is.
- `touched` is affected by `initialErrors`, since the docs usually recommend this pattern for rendering error messages: `{errors.fieldName && touched.fieldName && <FieldError error={errors.fieldName}>}` and the validation message would not be displayed otherwise.

One other major change is the removal of `componentDidUpdate`. It is replaced by `getDerivedStateFromProps`, because `componentDidUpdate` should `setState` only if the props *changed*. If `initialErrors` doesn't change, the errors are still present and should not be ignored. Old tests for `componentDidUpdate` were removed and replaced with tests for `getDerivedStateFromProps`. 

The logic for resolving `values` and `errors` from `props.initial{Values,Errors}` and `state.{values,errors}` is also changed from `resetForm` to object merge, where local state always overwrites props provided values. This allow us to display the most recent local validation error if there is one, or the original error, if there are no local validation errors produced.

Tasks:

- [x] Add docs
- [x] Add example

Partially Fixes: https://github.com/jaredpalmer/formik/issues/288

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -70,7 +70,8 @@ in `initialValues`. If you are using [`validationSchema`] (which you should be),
 keys and shape will match your schema exactly. Internally, Formik transforms raw
 [Yup validation errors](https://github.com/jquense/yup#validationerrorerrors-string--arraystring-value-any-path-string)
 on your behalf. If you are using [`validate`], then that function will determine
-the `errors` objects shape.
+the `errors` objects shape. If you are using [`initialErrors`] and no further
+validation was performed yet, it will match the ones provided.
 
 #### `handleBlur: (e: any) => void`
 
@@ -134,7 +135,7 @@ Trigger a form submission.
 
 #### `submitCount: number`
 
-Number of times user tried to submit the form. Increases when [`handleSubmit`](#handlesubmit-values-values-formikbag-formikbag--void) is called, resets after calling  
+Number of times user tried to submit the form. Increases when [`handleSubmit`](#handlesubmit-values-values-formikbag-formikbag--void) is called, resets after calling
 [`handleReset`](#handlereset---void). `submitCount` is readonly computed property and should not be mutated directly.
 
 #### `setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void`
@@ -267,7 +268,7 @@ don’t use both in the same `<Formik>`.
 ### `enableReinitialize?: boolean`
 
 Default is `false`. Control whether Formik should reset the form if
-[`initialValues`] changes (using deep equality).
+[`initialValues`] or [`initialErrors`] changes (using deep equality).
 
 ### `isInitialValid?: boolean`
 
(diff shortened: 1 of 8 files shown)


yanchith:
Hey, I noticed, it's not working correctly in all cases yet. I'll fix it, add more tests afterwards and then let you know.

yanchith:
Fixed, AFAIK. There might be some gotchas around `enableReinitialize` as it now doesn't `resetForm` anymore, but that behavior could be restored easily enough.

jaredpalmer:
This will take some time to review.

jaredpalmer:
Also for next time, always good to push a commit with failing tests so you know that they actually run and can fail in the CI.

yanchith:
Sorry for the longish PR and tests that passed. I'll do the failing commit next time. The PR was initially very small and simple, but the behavior was not intuitive, which I realized only after adding the example.

The one question I would like your input on is what is the ideal behaviour of `initialValues` and `initialErrors` with and without `enableReinitialize` during component updates.

The original was:
- with `enableReinitialize` it would `resetForm` to new values, forgetting user state
- without it the `initialValues` were ignored completely on updates

The problem is that's usually not what the user wants with `initialErrors` 

My idea is to alter the behavior a bit:
- with `enableReinitialize` it would stay as is (`resetForm`)
- without it the current values (and errors) would be computed via a merge with the ones in local state having precedence, eg `const newStateValues = { ...initialValues, ...state.values }`

This is currently not implemented, but I'll add it based on your feedback.

yanchith:
Hey @jaredpalmer, congrats on the 1.0 🍾 ❤️

To give the update on this, I've noticed in the Release Notes, that you want to think about `initialErrors`, `initialTouched`, etc... Turns out it was okay-ish to do without first-class support for these in our app (we added a wrapper or two), and I completely understand if you want to keep the library from accumulating too many features. 

If you do decide you want those features in, I am still more than happy to complete this (see above comment for possible approaches), just don't want to do it if it's only going to get dropped.

danlupascu:
Hello. Can you please review this PR and merge it if it's ok? I think this is the only feature that is missing right now. I think there are a lot of use cases when the form is submitted, an action is triggered, a request to the server is made in a middleware and in case of a failed request the errors should be added  to the form but there is no nice way to do that .. I know that we can pass the `setErrors` function in the action payload but that's more of a hack than a real solution.

jaredpalmer:
Agree. This will probably be the last bit of v1. Will not get to it until after React Conf though. Sometime next week

danlupascu:
Thank you very much!!! Can't wait to see this PR merged, I'm using this approach in two projects that I'm working on now and it will be very helpful.

eballeste:
@jaredpalmer this before hooks? waaaa

jaredpalmer:
Yeah maybe not. Hooks first

danlupascu:
Hi there. Any updates here?

yanchith:
I think it's time to close this due to bit rot. I'll gladly resubmit this against a newer version if there is interest from the maintainers and community, just ping me.

PR #710, opened 2018-06-26
RN doc: use submitForm instead of handleSubmit

slorber:
it looks to me by convention that "handleX" are suited for dom usage. even if a recent fix allows `handleSubmit` to work on RN on the next beta (not yet published) I think the `handleSubmit` method has no additional value for RN projects, so calling `submitForm` looks more appropriate to me

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,7 +580,7 @@ const MyReactNativeForm = props => (
       onChangeText={text => props.setFieldValue('email', text)}
       value={props.values.email}
     />
-    <Button onPress={props.handleSubmit} title="Submit" />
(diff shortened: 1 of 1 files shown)


sshmyg:
@slorber Don't you think changes in API names will lead to big problems among formik current users?

slorber:
@sshmyg I don't change the api name, I'm only changing the doc, and it only affect the RN users. Note that the current doc RN examples do not work against latest release, that's why I'm correcting the doc in the first place: to provide examples that actually work fine. I don't think this will lead to big problems, it will actually help people trying to use Formik right now on a RN project, as now the doc will provide working examples

Issue #510, opened 2018-03-13
How to correctly organize array of objects form?


kolyasya:
## Bug, Feature, or Question?
Question

At the moment there is only one example of FieldArray which handles array of strings.
https://github.com/jaredpalmer/formik#fieldarray

Is it possible to implement an array of objects? In case each of `friends` has `firstName`, `lastName` and `age` for example.

What is a proper way to do it?

latviancoder:
Here is what I used in my app:

```
<Form>
    <FieldArray
        name="friends"
        render={arrayHelpers => (
            <div>
                {values.friends.map((friend, index) => (
                    <div key={index}>
                        <Field name={`friends.${index}.name`}/>
                        <Field name={`friends.${index}.age`}/>
                        <button
                            type="button"
                            onClick={() => arrayHelpers.remove(index)}
                        >
                            -
                        </button>
                    </div>
                ))}
                <button
                    type="button"
                    onClick={() => arrayHelpers.push({ name: '', age: '' })}
                >
                    +
                </button>
            </div>
        )}
    />
</Form>
```

And here is an example validationSchema:

```
const friendValidationSchema = object().shape({
    name: string().required('name required'),
    age: number().required('age required').typeError('age must be a number')
});

const validationSchema = object().shape({
    friends: array().of(friendValidationSchema)
});
```

kolyasya:
@latviancoder thank you very much, it works!

mrchief:
@latviancoder Are you able to set `initialValues` using your approach? I'm currently using redux-form which doesn't seem to have support for initialValues + array of objects, so if Formik support it, I might as well move to Formik.

latviancoder:
Yes you can, take a look at documentation :)

mrchief:
The [documentation](https://github.com/jaredpalmer/formik#fieldarray) only has simple arrays but anyhow, I managed it with redux-forms. The trick was to not use `FieldArray` :smile:. Here's a [CodeSandbox](https://codesandbox.io/s/1vqqn68zx4), if anyone comes looking for a similar solution.

KeweiCodes:
@latviancoder  how is it that `friends.${index}.age` should work as the input name when you pushed { name: '', age: '' } ? 

If this is how things are supposed to work, then this is clearly not in in docs.

KeweiCodes:
@jaredpalmer help?

jaredpalmer:
[...]

PR #714, opened 2018-06-27
Update README.md Regarding Array of Objects

KeweiCodes:
Refer to: https://github.com/jaredpalmer/formik/issues/510#issuecomment-400526175

Also fixed minor no-key bug in the doc.


KeweiCodes:
Credits to @latviancoder for the code snippet.

PR #716, opened 2018-06-28
IsValidating

jaredpalmer:
Adds much overdue isValidating to Formik state.

We should consider moving to enums instead of boolean flags and then derive boolean flags for isSubmitting and isValidating.

We should also deprecate submitCount with a get() and replace it with submitAttemptCount or something.


jaredpalmer:
This has gone stale.  Closing this and going to reattempt with strings instead of booleans too.

PR #724, opened 2018-07-02
Fix: collecting field-level validations results before submit

almsh:
Looks like `setIn` doesn't mutating it's first argument, but returns new object instead.


jaredpalmer:
Will cut a release

jaredpalmer:
Fixed in 1.0.0-beta.4

Issue #608, opened 2018-05-02
Feature & Question: Hook for manual transformations


ghost:
## Feature & Question?
Feature/Question: an way/hook where to do yup casting /manually transformations.
lets say input text to uppercase.

## Current Behavior
No option or,  onChange={(name, value) => setFieldValue(name,  value.toUpperCase()  <-----here the transformation  )} but that feels not right putting business logic there.

## Desired Behavior
to be discussed, any hook between setting values and triggering validate in withformik function.

## Suggested Solutions
na

## Additional Information
see yup transform https://github.com/jquense/yup#mixedtransformcurrentvalue-any-originalvalue-any--any-schema example
```
var schema = yup.string().transform(function(currentValue, originalvalue){
  return this.isType(value) && value !== null
    ? value.toUpperCase()
    : value
});

schema.cast('jimmy') //=> 'JIMMY'

```
---

- Formik Version: v1.0.0-alpha.6
- React Version: na
- TypeScript Version: na
- CodeSandbox Link: na
- OS: na
- Node Version: na
- Package Manager and Version: na

sshmyg:
Does anyone from contributors team see this issue?

sshmyg:
@jaredpalmer Hi, can you inform, does this issue will be resolved in v1?

jaredpalmer:
I think there will be transform escape hatches in v1 for transform and maybe parse. However, I am not certain that _more_ Yup integration is coming/needed. The defaults seem to cover most use cases. Advanced yup usage like cast and transform can be utilized in a custom `validate` prop and using the exported `transformYupToFormErrors` function. This last bit needs more documentation. In fact, we've discussed internally about deprecating the `validationSchema` function from `formik` core and creating a seperate `formik-schema` package with this feature + necessary helper functions. Having to manage 2 ways to do validation makes the codebase much harder to maintain

sshmyg:
Yup schemas useful because it has some default vaildation features like `require`, `min`, `max`, and much more. You can add some other quick validation with `test` if you need it.
And if you want rid of yup, so formik users should do a lot of manual work to validate simple form.

sshmyg:
@jaredpalmer Reopen, please.

tedeh:
[...]

Issue #601, opened 2018-04-26
Trim not working


sketchbuch:
As a follow on to https://github.com/jaredpalmer/formik/issues/473

Well here is an example:

https://codesandbox.io/s/2jo7vox1n0

I have strict set to false. The validation for firstname detects whitespace and show a message but I don't want that, I want it to just trim the string and not show any message.

Jaikant:
It seems a few folks are facing this issue, including me.

vleandersson:
The reason for this is that Formik will not use the transformed values from the validation, but rather the original ones. So, using any transform methods from Yup with Formik will not have any implications on the final submit body. You will need to manually handle this. One example of such an implementation can be found on https://github.com/jaredpalmer/formik/issues/473 

```
<Field
  name="name"
  render={({ field, form: { isSubmitting, handleBlur } }) => (
    <div className="field-section">
      <Label htmlFor="name">Name</Label>
      <input
        {...field}
        onBlur={(e) => {
          const val = (e.target.value || '').replace(/\s+/gi, ' ');
          setFieldValue(field.name, val.trim());
          handleBlur(e);
        }}
        id="name"
        disabled={isSubmitting}
        type="text"
        placeholder="Your name"
      />
      <ErrorMessage name="name" component={ValidationError} />
    </div>
  )}
/>
```

johnrom:
Also, check out [format and parse](#2255).

vleandersson:
Amazing update @johnrom 

Until that is in, I've created a small set of helpers to solve this issue outside of Yup and Formik: https://www.npmjs.com/package/formik-yup

aj07mm:
Here: https://github.com/jaredpalmer/formik/issues/473#issuecomment-593587854

wayneparkes:
@vleandersson, your answer got me closer to a solution but I found that the field validator was validating against the original input value - not the trimmed value.

What I discovered, is that you need to invoke `handleBlur(event)` before carrying out any modifications to the input value, otherwise Formik will disregard any value set via `setFieldValue` and use the original event value if `handleBlur(event)` is called after.

Example:
```
<input
  {...field}
  onBlur={(event) => {
    handleBlur(event);
    const trimmedValue = (event.target.value || '').replace(/\s+/g, '');
    setFieldValue(field.name, trimmedValue);
  }}
  id="name"
  disabled={isSubmitting}
  type="text"
  placeholder="Your name"
/>
```

Issue #473, opened 2018-03-01
How does yup's trim work?


sketchbuch:
I tried adding trim() to my formik validation schema:

    Yup.object().shape({
        label: Yup.string().trim().required('some error msg'),
    });

But the field value is not being trimmed.

prichodko:
Should work like this - https://github.com/jquense/yup#stringtrimmessage-string-schema, do you get different result? If so, create a reproducible example on https://codesandbox.io/.

jquense:
`trim` is also a transfrom so if you are validating in 'strict' mode it won't run as with any other transform

sketchbuch:
Well here is an example:

https://codesandbox.io/s/2jo7vox1n0

I have strict set to false. The validation for firstname detects whitespace and show a message but I don't want that, I want it to just trim the string and not show any message.

rinoma:
Same problem here, trim doesn't seem to be working

jquense:
`strict()` doesn't take an argument (maybe it should) `strict(false)` doesn't turn off strict mode it enables it.

if I remove strict() it works fine: https://codesandbox.io/s/j35mz5xmz3

sketchbuch:
@jquense In your eample trim isn't working... the is always a space at the end of my firstname and also the space is visible in the alerts

jquense:
it seems like it...you'd have to ask @jaredpalmer tho, i'm just the yup guy :P

sketchbuch:
@jquense 

Why does strict(false) turn ON strict? That sounds backwards to me - I gues strict(true) turns off strict mode?

ArnieGA:
I've just stumbled upon the very same problem while attempting to TitleCase a Yup.string() inside the validationSchema.  The handleSubmit function doesn't return the results from transform.  For the time being I'll have to solve this by creating a new object inside the handleSubmit code block which will hold the curated values that come from the values argument.  It would be nice to update Formik to provide for this.

rpedroni:
Hey guys, any update on this? After digging deep into @jquense 's yup, apparently I was looking at the wrong place :P
It would be really useful if Formik could submit the transformed values.
For now, I have to externally/manually transform values into their correct api format, which is kind of prone to errors.

miphe:
@jaredpalmer Looks like Formik isn't passing transformed values (from Yup) to submit handler. This feels like an important functionality - any thoughts on it?

mahatosourav91:
[...]

PR #728, opened 2018-07-03
#608 fix yup transform before form submit

sshmyg:
Fixes for https://github.com/jaredpalmer/formik/issues/608

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -122,10 +122,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
    * Run validation against a Yup schema and optionally run a function if successful
    */
   runValidationSchema = (values: FormikValues, onSuccess?: Function) => {
-    const { validationSchema } = this.props;
-    const schema = isFunction(validationSchema)
-      ? validationSchema()
-      : validationSchema;
+    const schema = this.getValidationSchema();
     validateYupSchema(values, schema).then(
       () => {
         this.setState({ errors: {} });
@@ -332,7 +329,21 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
   };
 
   executeSubmit = () => {
-    this.props.onSubmit(this.state.values, this.getFormikActions());
+    const schema = this.getValidationSchema();
+    const { values } = this.state;
+    const actualizedValues = schema ? schema.cast(values) : values;
+
+    this.props.onSubmit(actualizedValues, this.getFormikActions());
+  };
+
+  getValidationSchema = () => {
+    const { validationSchema } = this.props;
+
+    if (!validationSchema) {
+      return;
+    }
+
+    return isFunction(validationSchema) ? validationSchema() : validationSchema;
   };
 
   handleBlur = (eventOrString: any): void | ((e: any) => void) => {
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import isEqual from 'react-fast-compare';
 import warning from 'warning';
+import deepmerge from 'deepmerge';
 import { FormikProvider } from './connect';
 import {
   FormikActions,
@@ -20,6 +21,7 @@ import {
   setIn,
   setNestedObjectValues,
   getActiveElement,
+  getIn,
 } from './utils';
 
 export class Formik<ExtraProps = {}, Values = object> extends React.Component<
@@ -34,13 +36,18 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
   };
 
   initialValues: Values;
-
+  didMount: boolean;
   hcCache: {
     [key: string]: (e: string | React.ChangeEvent<any>) => void;
   } = {};
   hbCache: {
     [key: string]: (e: any) => void;
   } = {};
+  fields: {
+    [field: string]: {
+      validate?: ((value: any) => string | Promise<void> | undefined);
+    };
+  };
 
   constructor(props: FormikConfig<Values> & ExtraProps) {
     super(props);
@@ -49,10 +56,12 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
       errors: {},
       touched: {},
       isSubmitting: false,
+      isValidating: false,
       submitCount: 0,
     };
+    this.didMount = false;
+    this.fields = {};
     this.initialValues = props.initialValues || ({} as any);
-
     warning(
       !(props.component && props.render),
       'You should not use <Formik component> and <Formik render> in the same <Formik> component; <Formik render> will be ignored'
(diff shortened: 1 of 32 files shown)


src/Formik.tsx
```diff
@@ -332,7 +329,17 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
...
+    const { values } = this.state;
+    const actualizedValues = schema ? schema.cast(values) : values;
+
+    this.props.onSubmit(actualizedValues, this.getFormikActions());
+  };
+
+  getValidationSchema = () => {
+    const { validationSchema } = this.props;
+
+    return isFunction(validationSchema) ? validationSchema() : validationSchema;
```
jaredpalmer: IIRC Pass props and values to validation schema function?

src/Formik.tsx
```diff
@@ -332,7 +329,17 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
   };
 
   executeSubmit = () => {
-    this.props.onSubmit(this.state.values, this.getFormikActions());
+    const schema = this.getValidationSchema();
+    const { values } = this.state;
+    const actualizedValues = schema ? schema.cast(values) : values;
```
jaredpalmer: This is breaking change

jaredpalmer: Can you describe when this is used and add tests for when cast() is called?

src/Formik.tsx
sshmyg: https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L147
Probably it's a bug in current release. Because props as arguments for validationSchema adds only here
https://github.com/jaredpalmer/formik/blob/master/src/withFormik.tsx#L127

src/Formik.tsx
sshmyg: Why it's breaking change?
If you use validation schema so you've expected that transformation will work on submit, but if you don't so, nothing change.
And even so, v1 version, I think admit this.

I have a bunch of inputs described with yup schemas, and a lot of them contains `transform` and `test` yup method. For example, user input amount in dollars, but server expect to get this field in cents. Simple one line of code, and you get it for all schemas you need.

As for me, I've migrate to `formik` from `react-formal` and as a result I have a lot of yup schemas, but some of them doesn't work as expected, because of `test` and `transform functions` and some others bugs related to yup usage within formik.

sshmyg: @jaredpalmer Have you ever tried to test real yup schema and not fake object?

sshmyg: `const ValidationForm = FormFactory({ validationSchema: { validate } });`
This is a pice of code from here  https://github.com/jaredpalmer/formik/blob/master/test/withFormik.test.tsx#L179
And all others `validationSchema` has the same form, just `validate` custom method and that's all.

But when I try to add new test like this
```
const validationSchema = yup.object().shape({
        opensource: yup.string()
            .transform(function(value: string) {
                return value
                    ? value.toUpperCase()
                    : value;
            })
    });

const FormWithInput = () => (
      <Formik
        onSubmit={onSubmit}
        initialValues={{ opensource: 'yay' }}
        validationSchema={validationSchema}
      >
```

I get a warning, about uncaught Promise exception.
I've reviewed all tests, and didn't find ` yup.object()` as validationSchema, just fake `{ validation }`.

sshmyg:
@jaredpalmer Check latest commit, please. I've added test case for yup transform method.

flybayer:
I'm really wanting this one fixed!  I'm always needing to trim whitespace from string inputs etc. I try to use yup().string().trim() because it seems like it should just work with Formik, but it doesn't.

See also #473 and #601

sshmyg:
@jaredpalmer Any thoughts?

sketchbuch:
Why do all the issues about broken transforms just get closed???

jaredpalmer:
This was a mistake. This bot went overboard. I’m working with GitHub support to reopen.

jaredpalmer:
Please reopen

alfonmga:
Any ETA when this will get merged/implemented?

flybayer:
To appease the stale bot: I think it's safe to say there's still a lot of folks wanting this feature!

sshmyg:
@jaredpalmer ?

Dman757:
Any chance of this making it in any time soon? @jaredpalmer @sshmyg V2?

cocacrave:
ETA please

sshmyg:
@Dman757 I don't know about yup support in v2 version. I am not a main contributor of formik, just want to fix this annoying bug.

miodrage:
Any news?

jaredpalmer:
I don't think cast before submit makes sense and it will awkward (if not impossible or a terrible idea) to use with typescript as the signature of values will change magically out of the blue.

jquense:
@jaredpalmer as a data point we use react-formal with typescript and return the cast value from onSubmit, typing it isn't really any harder than the existing case as far as i can tell. That said we type the value based on the schema e.g. `TValue = ReturnType<(typeof schema)['cast']>`. In practice yup casts don't change the type really, they insist that the type is correct or apply type specific transforms, i find it helpful for reducing the "turn empty strings from inputs into something more meaningful" code in form libraries.

Just my 2 cents tho, it's also really not hard to cast oneself in the submit handler if folks want that behavior. May also be nice to just return the cast value along with the form value since the work is done regardless and saves ppl from doing it twice.

ayepRahman:
Any updates on this =) ?

josejulio:
Is there anything we can do here to help getting this reviewed and merged? Thanks!

vleandersson:
Since this has been a bit stale for a while, I've created a small set of helpers to help out with this issue. https://www.npmjs.com/package/formik-yup

Looking forward to the day we can do this internally since Yup and Formik is quite amazing. Till then

loanburger:
As mentioned by others above, I am also tryign to trim string before the form submits.

using this code:
```
export const Contact = yup.object<IContact>().shape({
  contactName: yup
    .string()
    .transform(
      (value: yup.StringSchema, originalValue: yup.StringSchema): yup.StringSchema => {
        const schema = yup.string();
        return schema.isType(value) && value !== null ? value.trim() : value;
      },
    )
    // .trim('The contact name cannot include leading and trailing spaces')
    // .strict(true)
    .min(1, 'The contact name needs to be at least 1 char')
    .max(512, 'The contact name cannot exceed 512 char')
    .required('The contact Name is required'),
});
```
The string is trimmed in the function, but the Formik Form state is not updated so when the form is submitted i still includes leading and trailing spaces. My hope is that there is a way to trigger the Form state change to update it.

miodrage:
@loanburger Here is a wrapper.

```
import React from 'react';
import { Formik, FormikConfig } from 'formik';
import _ from 'lodash';

const trimObjValues = (obj: any) =>
  _.mapValues(obj, value => (typeof value === 'string' ? value.trim() : value));

interface TrimmedFormikProps extends FormikConfig<any> {}

const TrimmedFormik = (props: TrimmedFormikProps) => (
  <Formik
    {...props}
    onSubmit={async (val, formikBag) =>
      props.onSubmit(
        trimObjValues(await props.validationSchema.validate(val)),
        formikBag,
      )
    }
  />
);

export default TrimmedFormik;

```

Temirtator:
This is really issue, please solve this problem

christo8989:
@jaredpalmer 

"As you can see above, validation is left up to you. Feel free to write your own validators or use a 3rd party library. At The Palmer Group, we use Yup for object schema validation. It has an API that's pretty similar to Joi and React PropTypes but is small enough for the browser and fast enough for runtime usage. Because we ❤️ Yup sooo much, Formik has a special config option / prop for Yup object schemas called validationSchema which will automatically transform Yup's validation errors into a pretty object whose keys match values and touched. This symmetry makes it easy to manage business logic around error messages."

I'm not sure if you ever plan on supporting Yup transforms. At the minimum, will you add a disclaimer to the website so developers know transforms are not supported?

src/Formik.tsx
dalawwa: Thank you for the PR !
This is highly desired, although in our usecase, we need to be able to validate multiple fields depending on each other as described [there](https://github.com/jquense/yup/issues/706)

It looks like to achieve this, we would need to pass the `options` argument to `cast`, unless you found a different approach, which I would be happy to use.

Cheers

PR #730, opened 2018-07-03
Replace 'any' on <Form /> with proper types

tgriesser:
I excluded `onSubmit` since using that as a prop would defeat the purpose of the helper.


jaredpalmer:
oh shoot, when was `Exclude` introduced in TypeScript???

jaredpalmer:
Sorry about that, I got trigger happy on mobile.

PR #733, opened 2018-07-03
Preserve base types on <Field />

tgriesser:
in Field.d.ts:

Before:
```ts
export declare const Field: React.ComponentClass<any> & {
    WrappedComponent: React.ComponentClass<any>;
};
```

After:
```ts
export declare const Field: React.ComponentClass<FieldAttributes<{}>> & {
    WrappedComponent: React.ComponentClass<(React.InputHTMLAttributes<HTMLInputElement> & FieldConfig & {
        formik: FormikContext<any>;
    }) | (React.SelectHTMLAttributes<HTMLSelectElement> & FieldConfig & {
        formik: FormikContext<any>;
    }) | (React.TextareaHTMLAttributes<HTMLTextAreaElement> & FieldConfig & {
        formik: FormikContext<any>;
    })>;
};
```

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -226,4 +226,4 @@ class FieldInner<Props = {}, Values = {}> extends React.Component<
   }
 }
 
-export const Field = connect<FieldAttributes<any>, any>(FieldInner);
+export const Field = connect<FieldAttributes<{}>, any>(FieldInner);


jaredpalmer:
I think this will mess up React Native

tgriesser:
Is `<Field />` usable in React Native? The implementation looks pretty tied to the DOM.

jaredpalmer:
Yeah you are totally right. How does this impact custom field components?

tgriesser:
Yeah, this doesn't actually work well without being able to provide the additional props as a generic that should exist on the custom field component.

```jsx
<Field<Custom.Props> 
   component={Custom} 
   customProp={true} 
/>
```

Will take a look at an alternate approach here.

jaredpalmer:
In TS 2.9 we can utilize JSX generics. The problem is that in order to be backwards compat it would have to be opt-in (default to `any`) or else we would break every single usage of Field. I actually believe this is worth it as it is better than just plain old `any`.

jaredpalmer:
Can you reopen?

tgriesser:
You can't use generics with the current implementation though, `const Field = connect(...)`

jaredpalmer:
Oh shoot you are correct. What if we change the implementation of connect to use a class instead of SFC?

tgriesser:
Yep, that's why I had closed it :) 

Was just going to just take a shot doing with a class or a `function Field<P = any>(...)`. Same thing for `<FastField />`

tgriesser:
Closing in favor of #735

PR #737, opened 2018-07-05
fix: Set touched state on change event

### LINKED ISSUE #599: Odd behavior of touched
If I set up a form like this 

```js
class FormTest extends React.Component {

    state = {
        foo: {
            num: -1
        }
    };

    render()
    {
        return (
            <div>
                <Formik
                    initialValues={this.state.foo}

                        console.log("Submitted: ", JSON.stringify(values));
                        actions.setSubmitting(false);
                    }}

                    render={({errors, touched, isSubmitting, isValid}) => (
                        <Form>
                            <Field type="text" name="num" validate={v => isNaN(+v) ? "Invalid Number" : null}/>
                            {errors.num && touched.num && <div>{errors.num}</div>}
                            <button type="submit" disabled={ isSubmitting }>
                                Submit
                            </button>
                        </Form>
                    )}
                /></div>
        )
    }
}
```
## Current Behavior

I can observe the very odd behavior related to the "touched" status of the num field.

I load the component, change the input to "-1a". No errors is displayed, because the field is not "touched" yet. If I blur once, it counts as touched and shows the expected behavior: Toggling from "Invalid Number" according to the validity of the input.

Even more odd: If I submit the form once first, the field is flagged as touched instantly, and the first edit changes the error display

## Desired Behavior

Changing a field value is clearly "touching" the field, especially if just focusing and bluring it is "touching". While I understand the current behavior it seems neither expected nor really desirable.

## Suggested Solutions

Flag changed fields as "touched".

## Additional Information

Full Example: https://github.com/fforw/curly-octo-funicular
---

- Formik Version: 0.11.11
- React Version: 16.3.2
- TypeScript Version: ---
- CodeSandbox Link: ---
- OS: Linux 
- Node Version: v8.11.1
- Package Manager and Version: yarn 1.6.0

eckdanny-osi:
Closes #599

sets `touched#<field>` to `true` in `handleChange`.

*before*: `touched` only true after losing focus

formik-before

*after*: `touched` true when value is changed

formik-after

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -233,6 +233,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
         this.setState(prevState => ({
           ...prevState,
           values: setIn(prevState.values, field!, val),
+          touched: setIn(prevState.touched, field!, true),
         }));
 
         if (this.props.validateOnChange) {
--- a/test/Formik.test.tsx
+++ b/test/Formik.test.tsx
@@ -181,6 +181,23 @@ describe('<Formik>', () => {
           });
         expect(validate).not.toHaveBeenCalled();
       });
+
+      it('sets touched state', () => {
+        const tree = shallow(BasicForm);
+
+        // Simulate a change event in the inner Form component's input
+        tree
+          .find(Form)
+          .dive()
+          .find('input')
+          .simulate('change', {
+            persist: noop,
+            target: {
+              id: 'name',
+            },
+          });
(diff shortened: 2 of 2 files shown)


Andreyco:
This is intentional behaviou I believe. There are good reasons for this. Let @jaredpalmer confirm

jaredpalmer:
@Andreyco is correct. touched fires onBlur. Current behavior is intentional. You can build your own version of Field or your own component to get the behavior you desire with setFieldValue and setFieldTouched.

PR #743, opened 2018-07-06
pass form to field validation

evollu:
**Use case**
I need to do cross-field validation, like field1 should be less than field2.
I tried to do it with form level `validation` but it overrides field level validation results. On the other hand, I can't do this with Field's validation because it can't access other values.

**Proposal**
pass formikBag to `Field`'s validation method

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -159,7 +159,7 @@ class FieldInner<Props = {}, Values = {}> extends React.Component<
     const { setFieldError } = this.props.formik;
     const { name, validate } = this.props;
     // Call validate fn
-    const maybePromise = (validate as any)(value);
+    const maybePromise = (validate as any)(value, this.props.formik);
     // Check if validate it returns a Promise
     if (isPromise(maybePromise)) {
       (maybePromise as Promise<any>).then(


Andreyco:
Have this usecase as well and I used nasty hack to achieve this (pass values from different source). 
We should def consider this, maybe make API similar to Form's validate funcion

evollu:
@Andreyco agreed. It will be nice if Field's validation can allow returning an object. I can make another PR if @jaredpalmer agrees to move forward with this idea

Andreyco:
I don't think we need to return anything else but string. You want to validate one field and one field only. Input params are what we should focus on.

evollu:
@Andreyco then would current framework allow me to show error message on both fields when user only touches one filed?
E.g. there is a start date field and an end date field. 
when user change start date to be later than end date, I what to show error message on both fields.

Andreyco:
No, as it is now, it will not - only “focused” field is validated. But I think Formik should validate all mounted fields instead. Each field should self validate.

evollu:
@Andreyco triggering validation on all fields all the time seems unnecessary because this is a fairly rare use case. how about a function in FormikBag to trigger an `all field validation`?

Andreyco:
I am not sure at all 😬 Before field level validations were introduced, all fields were validated on every change (still are), have naver seen any slow downs or any other problems. But yes, Formik should do the least work required... lets wait for @jaredpalmer's opinion

Andreyco:
Looking at the codebase and I am agains passing anything but values to `validate` handler (people tend to abuse things).
Question is, what is "better" API.
```js
// Now
validate?: ((value: any) => string | Promise<void> | undefined);

// option A
validate?: ((value: any, values: object) => string | Promise<void> | undefined);

// option B
validate?: ((values: object) => string | Promise<void> | undefined);
```

evollu:
options B will be a breaking change. also, validate function will need to know field's name to access its own value, which is annoying for simple use case
But if only values get passed into the function, I will need to have a way to trigger validation on specific other fields (I have fields which does server-side validation, can't trigger it every time user types in another field)

Andreyco:
> options B will be a breaking change. also, validate function will need to know field's name to access its own value, which is annoying for simple use cases

You're right

> I think I can call validateField to achieve this. it is in FormikBag already

Yep, I added `validateField` in last PR, Jared though it would be useful and it is indeed. Maybe passing actions is not that bad after all. Would it be useful and pass it to form level `validate` function as well?

jaredpalmer:
Still thinking about this. I think `value, values` is strange, but probably what people want?

evollu:
@jaredpalmer why is this PR closed?

jaredpalmer:
I think formikBag makes the most sense. Let's update PR and tests so we can merge

evollu:
@jaredpalmer test updated

Issue #738, opened 2018-07-05
Fields not *always* marked as touched


Andreyco:
## Current Behavior
Only fields whose values are provided as part of `initialValues` are marked as touched when form is submitted.

## Steps to Reproduce
Set `initialValues={{}}` and submit form. Dump `state.touched` and see it empty.

## Expected behavior
All existing field should be marked as touched, even if they are not specified on `initialValues`

## Suggested solution(s)
Thanks to current API (`registerField` and `unregisterField`) we have map of mounted fields. Work with this data, instead of what is provided in `state.values`

## Additional context

**CodeSandbox Link**:
https://codesandbox.io/s/vqjmnkvjn7
---

- Formik Version: all

jaredpalmer:
I think this makes sense. Go for it.

Andreyco:
On it...

latviancoder:
This could be incredibly helpful for the project we are currently working on. Right now we have to manually create "empty objects" and it's a very tedious task, because there are many large and highly dynamic forms.

houkanshan:
Will this also fix the "uncontrolled to controlled error"? In your [CodeSandbox demo](https://codesandbox.io/s/vqjmnkvjn7), if I type some thing in the input, this familiar error will appear:

![image](https://user-images.githubusercontent.com/657962/43257717-1b45da8c-9103-11e8-8736-aa4d1cf91e45.png)

latviancoder:
@houkanshan thank you for this comment, helped me a lot!

weyert:
Hmm, interesting @houkanshan wondering where do you get `touched` and `error` from in your code snippet?

houkanshan:
@weyert updated my code. it's
```js
      const error = getIn(form.errors, name);
      const touch = getIn(form.touched, name);
```

weyert:
Of course, had to be late yesterday :D

latviancoder:
Is there any progress on this?

Davidmycodeguy:
How is the progress on this?

releaf:
I was able to wok around this by adding an onKeyUp event to the input I want to be validated as having been touched like this:
` onKeyUp={() => setFieldTouched(`${name}`, true)}`
https://codesandbox.io/s/5y2vlw594l

saschb2b:
I also noticed this problem when you want to track an input where the onBlur should not be the only result of focused. Like a checkbox.
I had to use the onFocus handler with a timeout to prevent immediate error popping up.
```
[...]

PR #745, opened 2018-07-08
Touch fields

Andreyco:


Andreyco:
closes #738

src/Formik.tsx
```diff
@@ -393,9 +394,12 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
   submitForm = () => {
     // Recursively set all values to `true`.
     this.setState(prevState => ({
-      touched: setNestedObjectValues<FormikTouched<Values>>(
-        prevState.values,
-        true
+      touched: deepmerge(
+        setNestedObjectValues<FormikTouched<Values>>(prevState.values, true),
```
Andreyco: Keeping it backward compatible

src/Formik.tsx
```diff
@@ -393,9 +394,12 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
...
+        Object.keys(this.fields).reduce(
+          (prev, curr) => ({ ...prev, [curr]: true }),
```
jaredpalmer: this.fields could contain deep paths like `social.facebook`, does this address this?

jaredpalmer: more explicitly, it could contain something like: 

```
{
  name: ...,
  'social.facebook': ...
}
```

test/withFormik.test.tsx
```diff
@@ -310,15 +310,30 @@ describe('withFormik()', () => {
...
         tree
           .find(Form)
           .find('form')
           .simulate('submit', {
             preventDefault: noop,
           });
+
         expect(tree.find(Form).props().touched).toEqual({
           name: true,
+          email: true,
```
jaredpalmer: Cool.

examples/FieldLevelValidation.js
```diff
@@ -7,7 +7,7 @@ const FieldLevelValidation = () => (
   <div>
     <h1>Pick a username</h1>
     <Formik
-      initialValues={{ username: '', email: '' }}
+      initialValues={{}}
```
jaredpalmer: You still should do this unless we give Field's value a default of `''`.

jaredpalmer: The thing is, if you use the props-only API, `initialValues` prop really isn't optional.

src/Formik.tsx
Andreyco: nope, haven't considered this use case. Will test that, no prob.

examples/FieldLevelValidation.js
Andreyco: Agree.

jaredpalmer:
retest this please

Issue #726, opened 2018-07-02
Switch Formik<Config, Values> to Formik<Values, Config> for better TS JSX Generic


jaredpalmer:
**Is your feature request related to a problem? Please describe.**
Using new JSX generics in TypeSCript is not as smooth as it could be.

**Describe the solution you'd like**
Switch `Formik<ExtraProps = {}, Values = object>` to `Formik<Values = object, ExtraProps ={}>` 

**Additional context**

PR #747, opened 2018-07-08
Begin to fix types, related to #733 and #735

tgriesser:
Ok so this is a new pass at #733 and #735, originally meant to get better types on `<Field />` but it grew a little in scope from there.

I found myself starting to get much more ambitious with cleaning up types in general but decided to stop and talk through some of these ideas/changes first and see if this is a direction I should continue.

Some goals:

**Use `any` and `... as ___` as little as possible.**

This can be useful occasionally, but IMO it's used a little much here, mostly to work around things that could otherwise be better expressed with the type system.

The fact that you can currently write a `<Field />` component with `render`, `children` and `component` and have it pass the type checker feels like a bug. The lack of strong typing here also leads to the need to manually cast in a bunch of places which can lead to subtle bugs. 

I split the constructor check which is similar in the various components to a util function (since that'll still be useful for JS consumers), but didn't begin to tackle proper typings to correct this in TS. 

Would you agree that proper type definitions for the various components would be useful here?

**Utilize TS 2.9 and generics as well as we can internally to track types.**

The generics are pretty neat in the fact that they are passed properly, TS inference does a good job of, for example, tracking what additional props should be defined on `<Field />` based on the `component={CustomComponent}` as long as the generic is passed through to the definition:

```ts
  /**
   * Field component to render. Can either be a string like 'select' or a component.
   */
  component?:
    | string
    | React.ComponentType<ExtraProps & FieldProps<any>>
    | React.ComponentType<void>;
```

**Unify type naming**

A pattern I've begun to use in my own React components and started to implement here is to define a namespace with the same name as the component. I've found it simpler to organize/reference via `ClassName.Props`, without needing to import a bunch of different typings when referencing them cross module. Don't feel strongly about it either way, just a suggestion.

**`<Values, ExtraProps>` vs `<ExtraProps, Values>`**

Didn't do anything with this, but kept thinking that this would probably be a lot nicer to work on if #726 were to be rolled into it. Especially since `ExtraProps` usually doesn't need to be defined explicitly by the end-user, as mentioned above, even with a default `{}` it appears to properly infer the props as defined by a custom component prop. Thoughts on whether I should roll that into some of the work here?

I'm going to add a few comments on the PR with other thoughts/notes as I was working through it. So this is still WIP (though the update type definitions seems to work pretty well as-is). I'm mostly interested to hear any feedback on these ideas / whether moving in the direction of getting a bit more strict/correct with the types internally is something welcomed/worthwhile to work on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "jest": "^23.1.0",
     "jest-cli": "^23.1.0",
     "lint-staged": "4.0.2",
-    "prettier": "1.11.1",
+    "prettier": "^1.13.7",
     "raw-loader": "^0.5.1",
     "react": "^16.3.2",
     "react-dom": "^16.3.2",
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -36,7 +36,7 @@ class FastFieldInner<ExtraProps, Values> extends React.Component<
       value: getIn(props.formik.values, props.name),
       error: getIn(props.formik.errors, props.name),
     };
-    warnRenderProps('FastField', props);
+    warnRenderProps('FastField', props, 'component');
 
     // Register the FastField with the parent Formik. Parent will cycle through
     // registered FastField's validate fns right prior to submit
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -1,11 +1,10 @@
 import * as React from 'react';
 import isEqual from 'react-fast-compare';
-import warning from 'warning';
-import { FieldAttributes, FieldConfig, FieldProps } from './Field';
+import { Field, FieldConfig, FieldProps, FieldInner } from './Field';
 import { validateYupSchema, yupToFormErrors } from './Formik';
-import { connect } from './connect';
+import { FormikConsumer } from './connect';
 import { FormikContext } from './types';
-import { getIn, isEmptyChildren, isFunction, isPromise, setIn } from './utils';
+import { getIn, isFunction, isPromise, setIn, warnRenderProps } from './utils';
 
 export interface FastFieldState {
   value: any;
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -71,7 +71,7 @@
     "jest": "^23.1.0",
     "jest-cli": "^23.1.0",
     "lint-staged": "4.0.2",
-    "prettier": "1.11.1",
+    "prettier": "^1.13.7",
```
tgriesser: Had to bump this to get it working with some of the more recent TS syntax.

src/FastField.tsx
```diff
@@ -307,4 +294,27 @@ class FastFieldInner<Props = {}, Values = {}> extends React.Component<
...
+  FastField.State
+> {
+  static WrappedComponent = FastFieldInner;
+
+  render() {
+    return (
+      <FormikConsumer<Values>>
+        {formik => (
+          <FastFieldInner<ExtraProps, Values>
+            {...this.props as FastField.Props<ExtraProps>}
```
tgriesser: I was having trouble determining the exact need for this casting, the trace here was pretty tough to follow. My hunch was that it had to do with the lack of properly discriminating between the `GenericFieldHTMLAttributes` union combined with the  `render` / `children` / `component`. I believe it goes away if we do a better job discriminating the types earlier. Similar story in the other components.

src/Field.tsx
```diff
@@ -41,13 +40,13 @@ export interface FieldProps<V = any> {
...
 
-export interface FieldConfig {
+export interface FieldConfig<ExtraProps> {
   /**
    * Field component to render. Can either be a string like 'select' or a component.
    */
   component?:
     | string
-    | React.ComponentType<FieldProps<any>>
+    | React.ComponentType<ExtraProps & FieldProps<any>>
```
tgriesser: This generic is what allows us have the additional props available on the `<Field />` inferred from the component props. Same idea should be applicable to the `<Formik />` component.

src/connect.tsx
```diff
@@ -1,22 +1,51 @@
...
-  Consumer: FormikConsumer,
-} = createContext<FormikContext<any>>({} as any);
+export const { Provider, Consumer } = createContext<FormikContext<any> | void>(
+  undefined
+);
+
+export function FormikProvider<Values>(
+  props: ProviderProps<FormikContext<Values>>
+) {
+  return <Provider value={props.value} children={props.children} />;
```
tgriesser: This isn't `{...props}` due to a subtle difference in the typings with `create-react-context`. [PR open](https://github.com/jamiebuilds/create-react-context/pull/21/files) to fix.

src/connect.tsx
```diff
@@ -1,22 +1,51 @@
...
+}
+
+export function FormikConsumer<Values>(
+  props: ConsumerProps<FormikContext<Values>>
+) {
+  return (
+    <Consumer
+      children={formik => {
+        if (typeof formik === 'undefined') {
+          throw new Error(
```
tgriesser: I've found this useful in my own use of the new `context` API, to either throw or warn and return `null` when a context dependent field is used outside of the provider. Thoughts?

src/types.tsx
```diff
@@ -217,7 +206,7 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
  * of <Formik/>.
  */
 export type FormikProps<Values> = FormikSharedConfig &
-  FormikState<Values> &
+  Formik.State<Values> &
   FormikActions<Values> &
```
tgriesser: We could probably move several of these types under a `Formik` namespace for consistency.

src/utils.ts
```diff
@@ -130,3 +131,20 @@ export function getActiveElement(doc?: Document): Element | null {
     return doc.body;
   }
 }
+
+export function warnRenderProps(c: string, props: any, ignored: string) {
```
tgriesser: This is actually inconsistent between `<Formik />` and `<Field />, <FastField />, etc.` - in `Formik` a `component` prop takes precedence and in `Field` a `render` prop takes precedence. Could we pick one (even if types are changed to eliminate this as a problem in TS).

tgriesser:
I also started doing a little digging on issues and found #561 - where it's mentioned that `<Field />` is the only component that should pass through props. 

With this in mind, would it also make sense to remove the generic `ExtraProps` entirely from 
`<Formik />`?

jaredpalmer:
@tgriesser I think so. IIRC, it's used in withFormik?

tgriesser:
So looking at `withFormik` it currently uses a `Props` generic, and this makes sense to have there since it needs them for `mapPropsToValues`. It currently passes these props along to the `<Formik />` component via `{...this.props as any}` 
https://github.com/jaredpalmer/formik/blob/547bbfcbe65558abeced4f40222890eae093119e/src/withFormik.tsx#L147-L151

but it doesn't actually need to since `<Formik />` doesn't pass any of the props along to the child, it only passes what's returned from `this.getFormikBag()`.

https://github.com/jaredpalmer/formik/blob/547bbfcbe65558abeced4f40222890eae093119e/src/Formik.tsx#L575-L577

@jaredpalmer So yeah - if this sounds reasonable I can work on getting rid of it. Do you have any reservations on any of the other things I mentioned here?

Issue #751, opened 2018-07-10
Dependent Dropdowns and Dynamic Fields


parlays:
Hello!  I have read through many of the issues and online examples but I have not found an example of using dynamic fields.

I would like to add or remove form fields based on the value of one of the dropdowns.  If the dropdown has "option A" selected, I would like to add two more dropdowns and populate them via an AJAX call.  If "option B" is selected, I would like to add an input box and remove the dropdowns displayed for "option A".

I tried using a custom react component but it is not clear how to add or remove form elements based on the value of another form element.  Thanks!

SimpleCookie:
I am looking into the exact same thing right now. Still haven't found any answer.
I can't and shouldn't do setField inside render, otherwise that would be an option for me. But it's a major anti-pattern.

SimpleCookie:
Okay @parlays I have solved the issue. It's not a pretty solution but it's a workaround.
I have a custom component which obviously calls `setFieldValue`, but it also calls a `onValueChangeCallback` to my container/page, which in turn updates the "options" prop to my form based on the option chosen.

To update the dynamic field's value, I am also calling `setFieldValue` on the `onChange` for my static field.

I'm sorry, I know my explanation is poor and hard to follow. My code and form are quite complex so it's difficult to give you a simple explanation.

But basically, what you want to do is to handle it through a callback I guess.

pavanmehta91:
@SimpleCookie Can you give an example to show this?

Chris-Boe:
@jaredpalmer Do you have any recommendations for dynamic and dependent fields?

ajitfawade:
> @pavanmehta91 I'm sorry I don't have the possibility of creating an example for you right now.
> But .. Say you have 2 dropdowns, parent A and dependant child B.
> 
> When you change A, you want to update the options of B accordingly.
> What you need to do is, you need to first use `setfieldValue` to set value of A, and then you need to update the options of B, I did this using a callback. And then you also need to use `setfieldValue` to set the initial value of B.

Could you please give an example for the above case?

dzg:
Any word on this? What's the most elegant way to solve?

goldenshun:
I came across this looking for a way to set default fields in Formik's `values` when a conditional set of form fields is rendered. This is a simplified version of what my code looks like:

```
const ConditionalFields = (props) => {
  const { handleChange, setFieldValue, values } = props;

  useEffect(() => {
    // If my values are undefined, set them in Formik
[...]

PR #763, opened 2018-07-13
add stateReducer prop to intercept state changes

cdock1029:


src/Formik.tsx
```diff
@@ -118,28 +122,53 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
...
+  ) => {
+    if (!this.didMount) {
+      return undefined;
+    }
+
+    return this.setState(
+      state => {
+        let newStateToSet =
+          typeof stateToSet === 'function' ? stateToSet(state) : stateToSet;
+        newStateToSet = this.props.stateReducer!(state, newStateToSet);
```
ematipico: The exclamation mark: is it a TypeScript thing?

cdock1029: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

Since it's an optional property, it will show an error that it might be undefined, but it won't be because of `defaultProps`

cdock1029:
### Usage:

Probaby can do a lot with this, but specifically.. easily create dynamic fields:

```jsx
<Formik
  initialValues={{
    specialDate: format(this.state.specialDate, 'YYYY-MM-dd'),
    message: '',
  }}
  validationSchema={schema}
  stateReducer={state => {
    const { values } = state

    // check if specialDate is 2 or more days in future...
    const { specialDate } = value        
    if (isAfter(specialDate, addDays(new Date(), 1))) {
      values.dynamicInput = values.dynamicInput || ''
      return state
    }
    if (values.hasOwnProperty('dynamicInput')) {
      delete values.dynamicInput
      return state
    }
    return null
  }}
  {...rest}
/>
```

https://github.com/jaredpalmer/formik/issues/751

dupski:
+1 for this. Currently I cant see an easy way to do the class Category -> Subcategory drop-down pattern in formik. I need to be able to nullify the SubCategory when the Category changes.

Also I am using the `validate` prop on formik to intercept field value changes [yuk] :P

ematipico:
In order to be a state reducer, it should always return the initial state right? Like in redux, where this feature took inspiration. Also, it should always return a new object? Correct me if I am wrong.

dupski:
A reducer should either return the same state if there is no change, or a new state object if there is a change. Looking at the example above it returns null in some cases which is probably not right.

The main thing I am after is the ability to observe formik state outside of the render method.

I may be going about it the wrong way - unfortunately I couldn't find an example for what I'm trying to achieve. I'll try to mock up some code at the weekend and see if I can do it in render :)

cdock1029:
This was just an initial proof of concept to do what others are describing here... Observe state and change it given some condition. As Jared said in another thread to is probably more of a state middleware vs an actual reducer.

Returning null from getDerivedStateFromProps just tells react to not change state. Means if user doesn't provide a middleware then we aren't modifying internal state. 

GDSFP was just the simplest way to hook into it. A true reducer would have to be implemented like in downshift. The internal set state isn't needed here but would be if you went down the true reducer path.

Doing dynamic state changes at render isn't really possible because we can't call setState there... Or am I incorrect?

PR #780, opened 2018-07-20
Support field level validation via yup schema

jquense:
This s quick and dirty spike on supporting per field validation by default for a form level yup schema.

I may need to tweak the yup api a bit more but this, i think demonstrates what i have in mind

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,28 +1,28 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 144255,
-    "minified": 39643,
-    "gzipped": 12000
+    "bundled": 144429,
+    "minified": 39700,
+    "gzipped": 12009
   },
   "./dist/formik.umd.development.js": {
-    "bundled": 174435,
-    "minified": 50097,
-    "gzipped": 15119
+    "bundled": 174609,
+    "minified": 50154,
+    "gzipped": 15123
   },
   "./dist/formik.cjs.production.js": {
-    "bundled": 44141,
-    "minified": 22101,
-    "gzipped": 5521
+    "bundled": 44399,
+    "minified": 22158,
+    "gzipped": 5525
   },
   "./dist/formik.cjs.development.js": {
-    "bundled": 45029,
-    "minified": 23001,
-    "gzipped": 5862
+    "bundled": 45287,
+    "minified": 23058,
+    "gzipped": 5863
   },
   "dist/formik.esm.js": {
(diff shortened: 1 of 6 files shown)


jaredpalmer:
I really like this.

morgs32:
@jquense just curious about the motivation? Performance right?

Problem perhaps is `yupToFormErrors` recreates the errors object so if you had an error in another field it would be erased I think.

jquense:
The motivation is to be able use yup with validation patterns I'm used to. Validating every field on any change is an odd approach for most of my form use-cases. And yes we'll have to merge the error object in this case. tho isn't that already happening for the existing individual field validations?

carlosagsmendes:
This looks great. Any idea about when it will be available? Thanks in advance

jaredpalmer:
@eonwhite do you see anything wrong with this?

morgs32:
@jaredpalmer @jquense 
Looks like `yupToFormError` recreates `errors` on scratch. That will mean running validation on **one** field will blow away errors on all other fields. Am I right? Perhaps a test to prove it doesn't?

https://github.com/jaredpalmer/formik/blob/57c4427ae52ba51ce359c21367b6678ab2861c1e/src/Formik.tsx#L655)

morgs32:
And if you can field level errors, then would this also resolve #150 ?

bradens:
This looks awesome, any chance of getting it merged soon?

jquense:
sorry folks haven't had enough time to try and finish this up, hopefully soon!

chris-fsl:
Any way this one might get updated soon??

himselfego:
@jquense is there any chance of getting this feature?

jquense:
So looking back on this, under the current code i'm sure how necessary this is anymore. The current behavior is generally fine if somewhat inefficient. I'd say that doing this is no-regrets anyway except that syncing the error state may have a few edge cases where validating one field would clear the error for another field, e.g. a validation built on top of a `when()`.

perhaps it's worth making this configurable or maybe there is something fancy i could do in yup to resolve backwards when there are interdependent fields. thoughts @jaredpalmer ?

jaredpalmer:
I guess the fanciest way would be to have Yup/formik degrade to form-level if it encountered a .when(), ref, or test? I have no idea how that would work in Yup, but that would make this non-breaking.

jaredpalmer:
@jquense see #1578 for the v2 version of this

jaredpalmer:
See #1578 for current progress. Closing.

Issue #727, opened 2018-07-02
1.0.0-beta.4: form `isValid` is not correctly set


codepunkt:
## Current Behavior
`isValid` is `false` after validation even though no errors exists on a single field form

## Steps to Reproduce
https://codesandbox.io/s/q336r5pnkw
Type in a username. The validation that is performed on the field level on change is a combination of sync `yup` and async username validation (mocking an api call checking for the existance of a username). The `isValid` prop of the form does not update according to the validation results.

## Expected behavior
When neither a sync yup validation error nor an async validation error occurs, `isValid` is true.

## Suggested solution(s)
No clue. Am i doing the validation right?

**CodeSandbox Link**:
https://codesandbox.io/s/q336r5pnkw

---

- Formik Version: 1.0.0-beta.4
- React Version: 16.3.2
- Browser and Version: Google Chrome Version 67.0.3396.99
- OS: Windows 10

tgriesser:
The issue here is that the [`setFieldError`](https://github.com/jaredpalmer/formik/blob/7fc86da2114933b59f0176ab42a9aa07fb6a5647/src/Formik.tsx#L489) uses the  [`setIn`](https://github.com/jaredpalmer/formik/blob/master/src/utils.ts#L25-L48) helper, which does not remove undefined values, so the errors object ends up as `{errors: { username: undefined }}`. The [`isValid`](https://github.com/jaredpalmer/formik/blob/7fc86da2114933b59f0176ab42a9aa07fb6a5647/src/Formik.tsx#L552-L553) currently just checks for `Object.keys(this.state.errors).length === 0` which is incorrect.

@jaredpalmer thoughts on the best place to fix this, in the `setIn` helper, in the `setFieldError` implementation or in the `isValid` check?

Signature for `setFieldError` should also probably be changed from 
```
setFieldError = (field: string, message: string)
``` 
to 
```
setFieldError = (field: string, message: string | undefined)
```

jaredpalmer:
I see. Yeah this is a bug (and a bad one). Would the PR to change setIn implementation to just lodash’s fp’s _set help?

codepunkt:
`set` from `lodash` doesn't remove undefined values afaik and would also result in `{ errors: { username: undefined }}`.

tgriesser:
Also building with any of the fp helpers, even cherry picking via `lodash/fp/set` significantly increases bundle size.

jaredpalmer:
[...]

PR #781, opened 2018-07-21
fix: remove key from object on setIn with undefined value (#727)

codepunkt:


src/Formik.tsx
```diff
@@ -493,7 +493,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
     );
   };
 
-  setFieldError = (field: string, message: string) => {
+  setFieldError = (field: string, message: string | undefined) => {
```
jaredpalmer: Do we need to update setFieldValue and setFieldTouched too?

codepunkt: I'm not sure - i updated the `setFieldValue` typings because of a comment of @tgriesser in #727.
Let me have a look!

codepunkt: Seems we only need to update `setFieldError`.

codepunkt:
I don't know of any gotchas. I have to admit that i'm not a heavy user of `formik` yet, so i probably can't tell. This is just a thing that kept me from digging deeper and investing more time into the library, so i tried to help.

At least all tests are run and we've got no regressions on that end. That's gotta be something! 😆

Issue #775, opened 2018-07-18
Error in FormikErrors type after upgrading to v1.0.1


ErlendS:
## Current Behavior
My TypeScript project does not compile after upgrading to v1.0.1 because the FormikErrors type has changed.

![screenshot 2018-07-18 11 46 57](https://user-images.githubusercontent.com/13867964/42874044-6ba90e56-8a80-11e8-9f0e-914fc5dd800b.png)

![screenshot 2018-07-18 11 54 11](https://user-images.githubusercontent.com/13867964/42874425-6a652e7a-8a81-11e8-9daa-2f3096d800fe.png)

## Steps to Reproduce

Upgraded from 0.11.11 to 1.0.1

## Expected behavior

I expected the FormikErrors type to have values of string, but it has `{}` instead.

Snippet from Formik source:
```
export declare type FormikErrors<Values> = {
  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : {}
};
```
Notice that it seems to recursively add FormikErrors until it sets any field to `{}`

## Suggested solution(s)

```
export declare type FormikErrors<Values> = {
  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : string
};
```
After going through the release notes and the readme of 1.0.1, and it does not seem to me like Formik has changed its error structure since 0.11.11. I also console.logged the errors object to confirm that it's an object of strings.

## Additional context

**CodeSandbox Link**:

---

- Formik Version: 1.0.1
- React Version: 16.4.1
- TypeScript Version: 2.9.1
- Browser and Version: N/A
- OS: OS X 10.11.6
- Node Version: 8.11.1
- Package Manager and Version: Yarn 1.7.0

jaredpalmer:
Formik errors can be nested object of strings. We have been using existing 1.0.x with 2.9. Can create a sandbox to reproduce this type error?

hubgit:
The [documentation](https://github.com/jaredpalmer/formik/blob/master/README.md#errors--field-string-string-) says that errors are strings, so the types should at least _allow_ strings.

I found [an issue asking for React nodes to be allowed as error messages](https://github.com/jaredpalmer/formik/issues/266), so perhaps the errors should be allowed to be `any`?

thomastvedt:
[...]

PR #798, opened 2018-08-01
Fix: FormikErrors type in (#775)

rei-m:
- fix FormIkErrors types
- see #775


src/types.tsx
```diff
@@ -11,7 +11,9 @@ export interface FormikValues {
  * Should be always be and object of strings, but any is allowed to support i18n libraries.
```
hubgit: This documentation line has problems, but it suggests that the final error type should be `any`.

rei-m: I see. so I'll change `string` to `string | any` 👌

hubgit: Why not just `any`? 😃

rei-m: just any OK👌I referred to previous commit. but it was old 😅

rei-m: I fixed errors type and merged commit message.

PR #807, opened 2018-08-04
Optimize rendering performance, prevent repetition validate

jinker:
When changing value of a field, avoid rendering other fields and trigger other fields' validate function.

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -6,7 +6,7 @@ import {
   GenericFieldHTMLAttributes,
   FormikContext,
 } from './types';
-import { getIn, isEmptyChildren, isFunction, isPromise } from './utils';
+import { getIn, isEmptyChildren, isFunction } from './utils';
 
 /**
  * Note: These typings could be more restrictive, but then it would limit the
@@ -78,6 +78,8 @@ export interface FieldConfig {
 
   /** Inner ref */
   innerRef?: (instance: any) => void;
+
+  enableDefaultUpdateFunc?: boolean;
 }
 
 export type FieldAttributes<T> = GenericFieldHTMLAttributes & FieldConfig & T;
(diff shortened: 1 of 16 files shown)


src/Field.tsx
```diff
@@ -78,6 +81,17 @@ export interface FieldConfig {
 
   /** Inner ref */
   innerRef?: (instance: any) => void;
+
+  shouldFieldUpdateOrEnableDefault?:
```
jaredpalmer: how about just `shouldUpdate`?

src/Field.tsx
```diff
@@ -78,6 +81,17 @@ export interface FieldConfig {
...
+    | boolean
+    | ((
+        props: FieldAttributes<any> & { formik: FormikContext<any> }
+      ) => boolean);
+
+  /** Should field need validate again */
+  shouldValidate?: (
+    meta: { values: any; value: any; fieldReason: string }
```
jaredpalmer: what is `fieldReason`?

src/types.tsx
```diff
@@ -203,9 +203,17 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
...
+    values: Values,
+    meta: {
+      fieldReason?: string;
+    }
   ) => void | object | Promise<FormikErrors<Values>>);
 
+  /**
+   * Whether validate on other changing at the initialization
+   */
+  validateFieldOnOtherChangeAtInit?: boolean;
```
jaredpalmer: What does this do?

jaredpalmer:
I think we should move these optimizations to `<FastField>`

src/Field.tsx
jinker: `fieldReason` is the reason trigger validate, or the changing field

src/Field.tsx
jinker: en, `shouldFieldUpdateOrEnableDefault` may be boolean type, and indicates whether built-in logic is enabled.

src/types.tsx
jinker: Sorry, a deprecated code, deleted.

PR #816, opened 2018-08-07
implemented allowInvalidSubmission prop

julian-r:
Proposal

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -33,6 +33,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
     validateOnBlur: true,
     isInitialValid: false,
     enableReinitialize: false,
+    allowInvalidSubmission: false,
   };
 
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Not sure why this got closed. Hrmmm. This not got wayyy too aggressive.

jaredpalmer:
I wonder if we ought to make this function like `allowSubmit(state: FormikState): boolean` and pass that form state.

miturostislav:
I'm dealing with a situation when I have to handle the failed submit. It may be worth  adding a new prop like `onFailedSubmit(errors: FormikErrors<any>, formikActions: FormikActions<Values>): void` and in case we want to submit the form anyway, we can do it in this function.

matths:
#466 and #816 both got closed and never merged, although there have been PR with actual solutions to this kind of request. Why that? Is there a solution to this kind of problem at the moment?

PR #827, opened 2018-08-14
Set filed to null when field type is number and value is empty

andrepcg:
I have a number input and when I remove the last digit the field takes an empty string value (`""`). This is breaking some form submissions on the server side (graphql) as the input is defined as `types.Float`

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -327,7 +327,7 @@ export class Formik<ExtraProps = {}, Values = object> extends React.Component<
           });
         }
         val = /number|range/.test(type)
-          ? ((parsed = parseFloat(value)), isNaN(parsed) ? '' : parsed)
+          ? ((parsed = parseFloat(value)), isNaN(parsed) ? null : parsed)
           : /checkbox/.test(type) ? checked : value;
       }
 


jaredpalmer:
I need to run some tests on this. This maybe a hidden null bomb.

jaredpalmer:
@eonwhite should we merge this? Will this break our currency inputs?

eonwhite:
Hmm the main problem with `null` is that the Typescript types for `input`'s `value` prop prohibit null.

But more generally, the right way to handle a number input is to make it a string and then convert it into a number at payload-processing time. You can have values "3." vs "3.0" vs "3.00" which all represent different string states, but all ultimately resolve to the same number. So the type of the Formik value should be a string, and the handleSubmit should do the work of converting it into a number or null to satisfy GraphQL.

eonwhite:
Or a better alternative actually @andrepcg if you want inputs that are GraphQL-friendly, don't use `handleChange` -- make a `NumberInput` component that tracks its value as an internal state string and then calls Formik's `setFieldValue` with either null or a number.

andrepcg:
I was just trying to make use of the input `type=number` that already does some processing for you (at least restricts your input to numbers only)

eonwhite:
Yeah, but even though input `type=number` guarantees a valid number, the same number can still take different forms while you're typing.

So its value prop needs to be a string at least some of the time, otherwise you can't distinguish `3.` from `3` or `3.0`, and the user will get weird feedback like typing `3.` and having an `0` automatically appended or seeing the `.` deleted.

So my recommendation if you want a higher-level abstraction that guarantees that the result is number or null, just wrap in a little reusable `NumberInput` component.

PR #830, opened 2018-08-16
Fix error props when the Value does not extend object

bkonkle:
This resolves errors like this for me:

```
Type '{} | undefined' is not assignable to type 'string | undefined'.
  Type '{}' is not assignable to type 'string'.
TextField.tsx(40, 3): The expected type comes from property 'error' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<TextField> & Readonly<{ children?: ReactNode; }> & Readonly<Partial<TextFieldProps>>'
```

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -11,7 +11,7 @@ export interface FormikValues {
  * Should be always be and object of strings, but any is allowed to support i18n libraries.
  */
 export type FormikErrors<Values> = {
-  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : {}
+  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : Values[K]
 };
 
 /**
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -11,7 +11,7 @@ export interface FormikValues {
  * Should be always be and object of strings, but any is allowed to support i18n libraries.
  */
 export type FormikErrors<Values> = {
-  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : Values[K]
+  [K in keyof Values]?: Values[K] extends object ? FormikErrors<Values[K]> : string
 };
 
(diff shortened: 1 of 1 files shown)


jaredpalmer:
I think this makes sense. Honestly, choosing most sensible fallback types here is above my TS knowledge.

cc @weswigham does this look okay?

bkonkle:
I'm second guessing myself now - if `Values[K]` is something other than `string`, like `number`, my problem will come back. I'm revising it - I think it should default to `string`.

bkonkle:
I'm _using_ it like this, BTW:

```
        <TextField
          label="Email"
          name="username"
          value={values.username}
          error={touched.username ? errors.username : undefined}
          handleChange={handleChange}
          block
        />
```

bkonkle:
For anyone else running into the same issue, here's how I'm working around in the meantime:

```
import {InjectedFormikProps} from 'formik'

type FormikErrors<V> = {
  [K in keyof V]?: V[K] extends object ? FormikErrors<V[K]> : string
}

type WithFormikProps<P, V> = InjectedFormikProps<P, V> & {errors: FormikErrors<Values>}
```

bkonkle:
Awww, my workaround causes problems when I go to wrap my HOC with `withFormik`, though:

```
Types of property 'errors' are incompatible.
                      Type 'FormikErrors<Values>' is not assignable to type 'FormikErrors<Values> & FormikErrors<Values>'.
```

I'll have to wait for the merge or build my own package locally. :-)

EQuimper:
Thank you @bkonkle get same error and didn't find solution until this :)

weswigham:
> cc @weswigham does this look okay?

It's simply a matter of what type is most sensible when it's not a nested object. I don't know enough about formik to know what the runtime behavior of usages of this type are.

jaredpalmer:
@weswigham you rock. 🙏 thanks!!

PR #837, opened 2018-08-21
WIP: Controlled Formik

jaredpalmer:


jaredpalmer:
@jquense can I get you advice on this? I am having some difficulty hooking up the changed field value to the controlled `onChange` prop. Would it be easier to use `uncontrolled`

jquense:
I think `uncontrollable` is usually the correct way to do this sort of thing, there are a lot more edge cases around controlling values than first appear. At the very least you really want to do this as a wrapping component, managing the state and props inline is messy, much simpler to have the base component just assume its always controlled and handle the uncontrolled behavior above it. 

In practice as well i've found it's usually better to split values up into different callbacks, so one for errors, the form value, touched, etc. It ends up being more flexible and mirrors the input onchange/value pattern a bit better. That can get a bit messy if you need to waterfall changes (e.g. trigger a value change then a `touched` change as a result of the new values. Often tho you can usually just avoid that, and fire all the relevant callbacks upfront.

Happy to offer any particular thoughts if you have them. I've used `uncontrollable` for react-bootstrap, react-formal, react-widgets, all with a lot of success and very few downsides.

jaredpalmer:
Word. I am going to attempt a prototype and may DM you with some questions.

PR #842, opened 2018-08-22
fix array helpers push side effect issue

### LINKED ISSUE #841: `arrayHelpers.push` pushes reference which causes unwanted side effects
## Current Behavior

When using `arrayHelpers.push` with an object it keeps reference of that object until it's used
formik push reference

## Steps to Reproduce

Use `arrayHelpers.push` with an object multiple times, update any input
## Expected behavior

keep all inputs seperate
## Suggested solution(s)

Just use clone depp when pushing
## Additional context

**CodeSandbox Link**:
https://codesandbox.io/s/jzx4nypz5
---

- Formik Version: 1.0.3
- React Version: 16.4.2
- TypeScript Version:
- Browser and Version:
- OS: 
- Node Version:
- Package Manager and Version:

islam3zzat:
Fixes side effect that happen when using object in `arrayProps.push`
Use `cloneDeep` when pushing new value in FieldArray 

| Before | After |
|:-:|:-:|
| formik push reference |  fix 841 |

fixes #841

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import cloneDeep from 'lodash.clonedeep';
 import { connect } from './connect';
 import {
   FormikContext,
@@ -125,7 +126,7 @@ class FieldArrayInner<Values = {}> extends React.Component<
 
   push = (value: any) =>
     this.updateArrayField(
-      (array: any[]) => [...(array || []), value],
+      (array: any[]) => [...(array || []), cloneDeep(value)],
       false,
       false
     );
--- a/test/FieldArray.test.tsx
+++ b/test/FieldArray.test.tsx
@@ -95,6 +95,37 @@ describe('<FieldArray />', () => {
       const expected = ['jared', 'andrea', 'brent', 'jared'];
       expect(formikBag.values.friends).toEqual(expected);
     });
+    it('should push clone not actual referance', () => {
+      let personTemplate = { firstName: '', lastName: '' };
+      let formikBag: any;
+      let arrayHelpers: any;
+      ReactDOM.render(
+        <TestForm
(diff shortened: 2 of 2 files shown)


klis87:
@jaredpalmer Are you sure it was the right decision to use `cloneDeep` here? In my opinion app developers (they know when) should be responsible for creating shallow/deep copy when required (like generally in React/Redux world) Defensive copying decreases performance and could cause some issues, for instance an Immutable.js/Mobx.another object could be passed to `push`, are we share lodash will correctly support deep copies for any object possible?

islam3zzat:
@klis87 Even with `immutable-js` the issue still there, if it the Map is not recreated every time.
https://codesandbox.io/s/jzx4nypz5

klis87:
@islam3zzat Thats weird, but personally I dont use immutable js so I dont know why this happens. But see this example, if you put new object inside push method, then everything works fine. https://codesandbox.io/s/p52qp1z1wm

klis87:
Anyway, in my opinion problem must be somehow deeper, probably Formik mutates objects somewhere, if it didnt, we wouldnt have issues passing the same object instance to `push`. I feel that deepcopy is workaround of mutability problem.

klis87:
I dont have time to investigate this, but I suppose that `setIn` must have some mutation going on in some array scenarios https://github.com/jaredpalmer/formik/blob/master/src/utils.ts#L25 Maybe it would be just better to use lodash FP set method? We use Lodash clone deep in current `setIn` method anyway, and it could potentially decrease application bundles (there is quite a chance another vendor lib/app itself could use it in other places)

klis87:
@islam3zzat I guess that Lodash FP `set` woudnt have this issue?

islam3zzat:
Yes, it won't help here

Issue #813, opened 2018-08-07
Use validateField on async validation will leak uncaught error.


houkanshan:
## Current Behavior

Use `validateField` on async validation will leak uncaught error.

## Steps to Reproduce

1. Open https://codesandbox.io/s/zxqyrqpxjx
2. Click the button. 

Then we got this:

![image](https://user-images.githubusercontent.com/657962/43759372-9a67a3c0-9a51-11e8-95ad-3d8acebaee8b.png)

## Expected behavior

The rejected promise in the `<Field/>`'s `validate` should be handled, without showing `Uncaught (in promise) "ERROR_MESSAGE"` in console or in error overlay.

## Suggested solution(s)

I think it's this line forgot that to handle the promise rejection?

https://github.com/jaredpalmer/formik/blob/f2d0d3bbe4e9801c556ba8f5a59eb46576cedb96/src/Formik.tsx#L166

## Additional context

A workaround is to just `return 'error message'` in the promise instead of throw it.

**CodeSandbox Link**: https://codesandbox.io/s/zxqyrqpxjx

---

- Formik Version: 1.0.2
- React Version: 16.4.2
- TypeScript Version: -
- Browser and Version: Chrome Version 68.0.3440.84
- OS: macOS Sierra Version 10.12.6
- Node Version: v6.12.2
- Package Manager and Version: npm@3.10.10

jaredpalmer:
Can you submit a PR?

houkanshan:
Will have a try, maybe next week.

PR #855, opened 2018-08-28
Fix uncaught error when use validateField on async validation

houkanshan:
Fix https://github.com/jaredpalmer/formik/issues/813


test/Field.test.tsx
```diff
@@ -131,6 +131,42 @@ describe('A <Field />', () => {
...
+      const FormFields = () => <Field name="name" validate={validate} />;
+
+      const tree = mount(<TestForm component={FormFields} />);
+      tree
+        .find(FormFields)
+        .props()
+        .validateField('name');
+      await Promise.resolve()
+        .then()
+        .then(); // XXX: Waiting for validateField's promise, should be fixed :(
```
houkanshan: Can't figure out a proper way to wait for the state's change, any idea?

PR #886, opened 2018-09-08
 Correct Validation.md with [Yup] part

steadylearner:
1. I rewrote line **81**, **143** and **199** with **[ import Yup from "yup"; ]** to **[ import * as Yup from "yup"; ]** inside github fork page.
2. I also suggest you modify some [**[Codesandbox examples]**](https://github.com/jaredpalmer/formik#example).
(I write this here for I can't modify code sandbox examples.)
  
   1) **Yup Lastest Version is 0.26.3** and every example with it is not compatible with new Yup API
   2) It also has to be from **[ import Yup from "yup"; ]** to **[ import * as Yup from "yup"; ]**

### DIFF
--- a/docs/guides/validation.md
+++ b/docs/guides/validation.md
@@ -78,7 +78,7 @@ npm install yup --save
 ```jsx
 import React from 'react';
 import { Formik, Form, Field } from 'formik';
-import Yup from 'yup';
+import * as Yup from "yup";
 
 const SignupSchema = Yup.object().shape({
   firstName: Yup.string()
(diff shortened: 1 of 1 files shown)


jaredpalmer:
I just updated the sandboxes.

steadylearner:
Thanks!

steadylearner:
I just found that two Codesandbox examples are not updated yet.
I hope that you correct them also and this message find you. 

1. [Draft js example](https://codesandbox.io/s/QW1rqjBLl)
2. [Component Lifecycle Example](https://codesandbox.io/s/pgD4DLypy)

PR #893, opened 2018-09-11
Using handleBlur with field name types were wrong

henrikra:
`handleBlur(e: any): void;` was capturing the types even though `handleBlur<T>(fieldOrEvent: T): T extends string ? ((e: any) => void) : void;` already takes care of the case if the parameter is something else than a string :)

That is why there is no need for both and tsc is happy


jaredpalmer:
Can you fix types?

henrikra:
I can but please check the RN crash again because that wasnt fixed 😄

jaredpalmer:
?

PR #898, opened 2018-09-12
Bump deps, fixes @types/react 16.4

jaredpalmer:
Bump to 

- React 16.4.5 
- TS 3.1.0
- TSLint / TSLint React

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,10 @@
     "warning": "^3.0.0"
   },
   "optionalDependencies": {},
+  "resolutions": {
+    "@types/react": "16.4.14",
+    "@types/react-dom": "16.0.7"
+  },
   "devDependencies": {
     "@pisano/enzyme": "^3.3.0-pisano.public.1",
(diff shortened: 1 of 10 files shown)


jaredpalmer:
Is yarn down? wtf.

jaredpalmer:
@weswingham I know it’s not your space but do you know why Formik’s render prop is now an implicitly any when it was inferred before?

weswigham:
The new react .d.ts uses a new entry point in the jsx namespace to allow for inference from `defaultProps` (using conditional types), and something about how that type + formik's type are written prevents inference from succeeding when defaultProps are present. TS 3.1 has a change that _might_ help (it helps a simplified example), but it also might not be enough on it's own (since the types are quite complicated).

weswigham:
Fyi, this should be fixed in TS 3.1 which aughta release very soon.

jaredpalmer:
Cool. I’ll wait.

jaredpalmer:
@wesigham it seems that with the new guard / function change, i can’t figure out how type optional render props (ie children?, component?, render?). Any suggestions?

PR #904, opened 2018-09-13
Added isSubmitting boolean to onSubmit handler

wilcoschoneveld:
In the following case:
```
        <Formik>
            onSubmit={values => {
                doSubmit(values).then(onClose);
            }}
            render={({ handleSubmit, ...rest }) => (
                    ...
                </form>
            )}
        />
```

Repeatedly pressing enter in the form would cause multiple calls to handleSubmit. (html5 spec: https://stackoverflow.com/a/3403470/1798137). You could prevent this with the following code:

```
<form onSubmit={!isSubmitting ? handleSubmit : undefined}>
   ...
</form>
```

However, in the case of `onSubmit=undefined` the browser does a default GET request on the page which causes the application to refresh. You need a `event.preventDefault` call but Formik's `handleSubmit` already takes care of this. Therefore I propose to send the `isSubmitting` state to the `onSubmit` handler so the user can prevent double submits.

Alternatives would be to prevent double submits in `handleSubmit` by checking for `isSubmitting` state, however maybe the user intentionally wants to re-submit if the case of a server hang?

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -437,7 +437,11 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
   };
 
   executeSubmit = () => {
-    this.props.onSubmit(this.state.values, this.getFormikActions());
+    this.props.onSubmit(
+      this.state.values,
+      this.getFormikActions(),
+      this.state.isSubmitting
+    );
   };
 
   handleBlur = (eventOrString: any): void | ((e: any) => void) => {
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -183,7 +183,11 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
   /**
    * Submission handler
    */
-  onSubmit: (values: Values, formikActions: FormikActions<Values>) => void;
+  onSubmit: (
+    values: Values,
+    formikActions: FormikActions<Values>,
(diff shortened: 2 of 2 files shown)


jaredpalmer:
Why not just disabled the submit button with `disabled={isSubmitting}`

wilcoschoneveld:
If there is a button with `type="submit"` then yes, that would work. But if there is no such button then you can submit indefinitely.

Demo: https://codesandbox.io/s/8krxk719wl
Steps: type in the input field and press enter multiple times

background:
https://www.w3.org/TR/html51/sec-forms.html#implicit-submission
https://www.tjvantoll.com/2013/01/01/enter-should-submit-forms-stop-messing-with-that/

It's a very specific use-case. I have mixed feelings about this, what do you think??

wilcoschoneveld:
For my project I solved the problem by setting the button `type` to `submit`. Still I think there might be a very rare use-case where this would still be a problem, but for now it's not necessary. Thanks for the feedback!

sijad:
I have a use case in react native, I'm submitting the form outside of render form by calling `this.formikRef.submitForm();` but there's no easy way to prevent `submitForm` there, because I can't access to `isSubmitting` via formik reference (`this.formikRef`), the only way I can think of is to add a local state.

Issue #529, opened 2018-03-21
setFieldValue should return a Promise or receive a callback


today-:
## Current Behavior
setFieldValue() is async because it uses setState(), but it doesn't receive a callback or return anything.
That's why it's hard to get modified values right after that, for example if I want to run submitForm().

## Desired Behavior
setFieldValue() will return a promise, so I could await it and get changed values.

## Suggested Solutions
return a Promise from setFieldValue() OR add additional callback to function declaration

- Formik Version: 0.11.11
- React Version: 16.2.0
- CodeSandbox Link: https://codesandbox.io/s/x9445yl5lp

prichodko:
Can you elaborate more on your use case? Submitting form after every single change seems really strange to me.

latviancoder:
Temporary hacky workaround I am using - put submitForm into a `setTimeout`.

today-:
Yes, setTimeout made this job, and even Promise.resolve works, but both looks like kludges

zakangelle:
A common use case for this is when you have a form with a set of filters, and you want your new filter values to be applied each time a filter is changed. With a callback option, this would look something like this:

```
<Select
  value={values.city}
  onChange={value => setFieldValue('city', value, submitForm)}
>
  {options.map(option =>
    <Option key={option.id} value={option.value}>{option.name}</Option>
  )}
</Select>
```

As @today- said, since `setFieldValue` uses `setState` (which is async), having some kind of callback would be nice for this type of use case.

gabriel-cliqz:
I also need the callback function.
In some cases is very handy

timc13:
It would be nice if FastField & Field exposed this callback as a prop as well.

FadiAboMsalam:
any update to this as i also need callback or promise or at least a way of doing this sync

timc13:
This is pretty essential.  There are many cases after you update a value on a field, you want to have access to the value immediately, or perhaps the `errors` or `touched` state.

mihkelmark:
I also found myself in dire need of this. 

I need to calculate a form's `total` value while others (`vat` and `sum`) are being typed: `calculatedTotal = sum * (vat / 100)`

Currently these two console logs log the same thing:

```
      console.log('values before', values)
      setFieldValue('total', calculatedTotal)
      console.log('values after', values)
```

elmeerr:
[...]

PR #909, opened 2018-09-14
feat(setFieldValue): support for callback

Woile:
relates #529 

Hey @jaredpalmer this is my first attempt at #529, I want to know if I'm in the right direction. I'm not proficient in typescript btw. 
I'm thinking about adding the same bahavior for the methods inside `getFormikActions`, mostly the `set*` ones:

```
setError
setErrors
setFieldError
setFieldTouched
setFieldValue
setStatus
setSubmitting
setTouched
setValues
setFormikState
```

Please review, and tell me what do you think
Thanks!

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -362,7 +362,10 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
     value: any,
     shouldValidate: boolean = true
   ) => {
-    if (this.didMount) {
+    return new Promise(resolve => {
+      if (!this.didMount) {
+        resolve();
+      }
       // Set form field by name
       this.setState(
         prevState => ({
@@ -371,11 +374,13 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
         }),
         () => {
           if (this.props.validateOnChange && shouldValidate) {
-            this.runValidations(this.state.values);
+            this.runValidations(this.state.values).then(() => resolve());
+          } else {
+            resolve();
           }
         }
       );
-    }
+    });
   };
 
   handleSubmit = (e: React.FormEvent<HTMLFormElement> | undefined) => {
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ legacy.d.ts
 .idea
 *.orig
 .DS_Store
+.vscode/
 
 node_modules
 
--- a/test/Formik.test.tsx
+++ b/test/Formik.test.tsx
@@ -569,6 +569,48 @@ describe('<Formik>', () => {
       ).toEqual('ian');
     });
 
+    it('setFieldValue promise is called succesfully with no validation', async () => {
+      const tree = shallow(BasicForm);
+      await tree
+        .find(Form)
+        .props()
+        .setFieldValue('name', 'ian', false);
+
+      expect(
+        tree
+          .update()
+          .find(Form)
+          .props().values.name
+      ).toEqual('ian');
+    });
+
+    it('setFieldValue promise is called succesfully with validation after validation', async () => {
+      const validate = jest.fn().mockReturnValue({});
+
+      const tree = shallow(
+        <Formik
+          initialValues={{ name: 'jared' }}
+          onSubmit={noop}
+          component={Form}
+          validate={validate}
+        />
+      );
+
+      await tree
+        .find(Form)
+        .props()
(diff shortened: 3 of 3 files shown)


jaredpalmer:
I was thinking about this the other day! Should we do a callback or promise?

Woile:
I went with callback because it looked more straightforward to me, but we can go for promises. 
The "benefit" I see with using promises is one less signature in the function, and that is more intuitive, at least for me, the first thing I did, when I had the need for this, was using `.then(...)`.

But sincerely, I wouldn't know if particularly for this case, there's a difference.

Would you like to see a Promise based implementation?

tpict:
Personally, I'd like a Promise that rejects if `this.runValidations` returns errors

Woile:
Okay I'll make an implementation with promises

prichodko:
Hey @Woile, have you started working on it? If not, I can tackle this as a part of `2.0`.

Woile:
I'm planning on updating next week, I've been super busy.

Is it ok?

Woile:
@prichodko @jaredpalmer I've made `setFieldValue` use promises. Before modifying the other functions, could you give some feedback? do you like the approach?

Also not sure what to do in case `!this.mount` (check the comment)

Regards!

src/Formik.tsx
```diff
@@ -366,20 +366,33 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
...
+        resolve();
+      }
       // Set form field by name
       this.setState(
         prevState => ({
           ...prevState,
           values: setIn(prevState.values, field, value),
         }),
         () => {
+          const values = this.state.values;
```
prichodko:
```suggestion
          const { values } = this.state;
```

src/Formik.tsx
```diff
@@ -366,20 +366,33 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
...
           if (this.props.validateOnChange && shouldValidate) {
-            this.runValidations(this.state.values);
+            this.runValidations(values).then(combinedErrors => {
+              const isValid = Object.keys(combinedErrors).length === 0;
+              if (!isValid) {
+                reject(combinedErrors);
```
prichodko: Honestly, I am not really sure if those methods should return any value at all. Most of the use cases just want to wait until it's set to call different handler. For example:

```js
onClick={async value => {
  await setFieldValue(name, value)
  submitForm() // or some different handler, which then reads necessary values from the state
}
```

what would be the use case for:

```js
onClick={async value => {
  try {
    const value = await setFieldValue(name, value) // what now?
  } catch (errors) {
     console.log(errors) // all errors, why?
  }
}
```

instead of:
```js
onClick={async value => {
    console.log(formikBag)
    await setFieldValue(name, value)
    console.log(formikBag) // updated
}
```

Maybe I am missing something, will wait for other thoughts.

Woile:
@prichodko I've updated to no promise rejection. The code is simpler now, and the rejected test was not needed anymore.

Should I start adding the promise to the other methods?
```
setError
setErrors
setFieldError
setFieldTouched

setStatus
setSubmitting
setTouched
setValues
setFormikState
```

prichodko:
Thanks for working on it! However let's wait for @jaredpalmer's thoughts as well.

src/Formik.tsx
tpict: I'll have a play around tomorrow morning and see if I can come up with an example. I think I have a use case in mind but I might be overthinking things

tpict: @Woile Yeah I was overthinking it. I was worried about a stale `errors` object after the Promise resolves but that's not the case in practice. I agree with @prichodko :)

Maxwell2022:
@prichodko can we re-open this PR? I guess everyone enjoyed the xmas break 😁

Woile:
If I get the ok I don't have any issue in finishing it

jaredpalmer:
The proble

jaredpalmer:
The problem is that we can't support this with hooks in v2 right now, because there is no equivalent of the second arg to setState yet. I want to merge this or something like it, but don't want to make this decision until there is clarity on what will or wont be possible

Woile:
Hey folks, quick update on this, I'm not gonna continue because I'm not up to date with hooks yet, and I've not played with formik v2. Once formik v2 is out, I'll update my app, and then I'll probably pick this up again, in case no one else has.

PR #915, opened 2018-09-16
Rerender <FastField /> when  isSubmitting has changed

### LINKED ISSUE #906: Rerender <FastField /> when `isSubmitting` has changed
Currently `<FastField />` isn't being rerendered when `isSubmitting` changes. Is there a good reason to not rerender it?

`isSubmitting` is a form wide attribute that doesn't change too often. From my experience typing or tabbing through the form is what causes bad perfrormance if all the fields are being rerendered.

I can create a PR If you would consider adding this.

stalmok:
Resolves #906

### DIFF
--- a/.storybook/example.stories.js
+++ b/.storybook/example.stories.js
@@ -7,6 +7,7 @@ import Arrays from '../examples/Arrays';
 import Basic from '../examples/Basic.js';
 import CustomInputs from '../examples/CustomInputs';
 import ErrorMessage from '../examples/ErrorMessage';
+import FastField from '../examples/FastField';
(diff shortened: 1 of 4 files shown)


jaredpalmer:
Please resolve conflicts

jaredpalmer:
Actually may revert this. I’m not sure this is needed.

stalmok:
My scenario is that the fields stay disabled if there are validation errors when user is trying to submit the form. Correct me if I'm wrong, Formik sets `isSubmitting=true` on submit and sets it back to `isSubmitting=false` if there are any validation errors.

How could I work around that while still using `<FastField />`? The performance with the `<Field />` is not tolerable.

PR #925, opened 2018-09-20
Alter errors and touched object on index change

dlehmhus:
Alter errors and touched object after index changing operations on the values array. (E.g. move, unshift, swap) 
Otherwise, after modifying the array, errors get assigned to the wrong input.


jaredpalmer:
I see. Can you make the tests pass?

dlehmhus:
@jaredpalmer works as expected now!

PR #939, opened 2018-09-26
Fix calls to validation with outdated values

### LINKED ISSUE #938: Validation called with wrong values on immediate consecutive updates
## Current Behavior

When `handleChange` is called twice without Formik having a chance to rerender/update state in between, the last write to values wins and previous writes are lost for the validation function. The `values` state is still correct, but `validate` will get called with outdated / non-existent values.

Assume the initial value is `{ currencyCode: '', amount: '' }`.

Then any user-land component code is effectively calling

```
handleChange({ target: { name: "currencyCode", value: "EUR" } })
handleChange({ target: { name: "amount", value: "3" } })
```

without Formik having a chance to update in-between.

This leads to the validation running for

```
{ currencyCode: "EUR", amount: "" }
{ currencyCode: "", amount: "3" }
```

instead of

```
{ currencyCode: "EUR", amount: "" }
{ currencyCode: "EUR", amount: "3" }
```

or just

```
{ currencyCode: "EUR", amount: "" }
{ currencyCode: "EUR", amount: "3" }
```

as the first update is lost.

## Steps to Reproduce

~~See steps noted down on https://codesandbox.io/s/40nj93okl9~~

~~The reproduction got a bit longer as I also wanted to illustrate the use-case for it.~~~

*I created a simplified version now to make the problem clearer: https://codesandbox.io/s/pk7mk0nny0*

## Expected behavior
Calling `handleChange` multiple times should not lead to validation with wrong values. The validation should be called with the right values instead.

## Suggested solution(s)
Wait for `setState` to finish before running the validation (as validation uses `this.state`).

I opened a PR fixing this: https://github.com/jaredpalmer/formik/pull/939

~~**CodeSandbox Link**: https://codesandbox.io/s/40nj93okl9~~~ *longer version*
**CodeSandbox Link**: https://codesandbox.io/s/pk7mk0nny0 *simplified version*

---

- Formik Version: 1.3.0
- React Version: 16.5.2
- TypeScript Version:
- Browser and Version:
- OS: 
- Node Version:
- Package Manager and Version:

dferber90:
Closes #938


src/Formik.tsx
```diff
@@ -324,14 +324,17 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
 
       if (field) {
         // Set form fields by name
-        this.setState(prevState => ({
-          ...prevState,
-          values: setIn(prevState.values, field!, val),
-        }));
-
-        if (this.props.validateOnChange) {
-          this.runValidations(setIn(this.state.values, field, val));
```
dferber90: As the above code sets the values, this code used to run with outdated values in certain cases. We may only run the validation once the state update has been commited, so we move the validation into the callback.

src/Formik.tsx
```diff
@@ -448,13 +451,16 @@ export class Formik<Values = {}, ExtraProps = {}> extends React.Component<
...
-        this.runValidations(this.state.values);
-      }
+      this.setState(
+        prevState => ({
+          touched: setIn(prevState.touched, field, true),
+        }),
+        () => {
+          if (this.props.validateOnBlur) {
+            this.runValidations(this.state.values);
+          }
```
dferber90: It does not matter here, but I moved it for good measure. I can also undo this part.

jaredpalmer: I would move this back and save a render.

dferber90: I force-pushed now to remove it.

Issue #942, opened 2018-09-27
Field ownProps are not passed into render function


zidik:
According to documentation:
Field's render props are an object containing:
- field: An object containing onChange, onBlur, name, and value of the field
- form: The Formik bag.
- **Any other props passed to field**

## Current Behavior

**myCustomProp** is not passed to the render function 

```
<Field
     name="lastName"
     myCustomProp="YES!"
     render={({ field, form, ...ownProps }) =>
          "This is not passed in: " + ownProps.myCustomProp
     }
/>
```

BUT it is, when using custom component:

```
<Field name="firstName" component={CustomInputComponent} />

const CustomInputComponent = ({ field, form, ...ownProps }) =>
  "Works here: " + ownProps.myCustomProp;
```

## Steps to Reproduce

- run code in CodeSandbox
- Following result appears:

> Works here: YES!
> but not here: undefined

## Expected behavior
- Following result appears:

> Works here: YES!
> but not here: YES!

## Additional context

**CodeSandbox Link**:
[![Edit llp0zjqrp9](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/llp0zjqrp9)

---

- Formik Version: 1.3.0
- React Version: 16.3.2
- TypeScript Version: N/A
- Browser and Version: Chrome 69.0.3497.100 (64bit)
- OS: Win 10
- Node Version: 8.12.0
- Package Manager and Version: Yarn 1.10.1

jaredpalmer:
Duplicate pf #957

prajavk:
Sorry for commenting on closed ticket. 
@jaredpalmer I m facing same issue with Field render. Can someone give working example related to fix?
FYI @christianchown
<Field
     name="lastName"
     myCustomProp="YES!"
     render={({ field, form, ...ownProps }) =>
          "This is not passed in: " + ownProps.myCustomProp
     }
/>

christianchown:
Hi @prajavk, don't think this is going to change. Your best (only?) hope is not to use `render` (or child-as-function), but use `component`, which receives `myCustomProp`:
```
function MyComponent({myCustomProp}) {
  return <>{/* myCustomProp === "YES!" */}</>;
}

<Field name="lastName" myCustomProp="YES!" component={MyComponent} />
```

PR #957, opened 2018-10-04
passed additional props to the `render` prop function

adsingh:
fixes bug #942

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -39,6 +39,7 @@ export interface FieldProps<V = any> {
     name: string;
   };
   form: FormikProps<V>; // if ppl want to restrict this for a given form, let them.
+  [key: string]: any; // Additional props
 }
 
 export interface FieldConfig {
@@ -163,7 +164,7 @@ class FieldInner<Values = {}, Props = {}> extends React.Component<
     const bag = { field, form: restOfFormik };
 
     if (render) {
-      return (render as any)(bag);
+      return (render as any)({ ...bag, ...props });
     }
 
     if (isFunction(children)) {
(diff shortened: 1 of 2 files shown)


adsingh:
If you think that the added test case creates duplication, I can merge it with the test above it.

jaredpalmer:
Hey! Thanks for this. I used to think this was the move for a while. However, this pass thru in `render` is redundant. You can pass  props directly to the components that need them inside of `render` creating a closure.

christianchown:
Hi @jaredpalmer, I'm not sure why this is closed off? Saying _"you can pass props directly to the components that need them inside of `render`"_ doesn't hold when using `Field` compositionally. Use case:

I have `class DecoratedField extends React.Component<{decoratedRenderProp}>`

DecoratedField is composed with a `<Field render={props.decoratedRenderProp} plusSomeOtherProps={whatever} />`

I would like to use it like
```
<CustomForm>
  <DecoratedField decoratedRenderProp=(({field, form, plusSomeOtherProps}) => ....} />
</CustomForm>
```
`plusSomeOtherProps` is only supplied by `DecoratedField` so is not available to me when composing a `CustomForm`, so I can't pass it directly in `render`. I am having to entertain some ridiculous structures to try to overcome this.

Or is there some other way of achieving this that I am not seeing?

christianchown:
@kelkes I didn't get any reply, and this issue is marked as closed and https://github.com/jaredpalmer/formik/issues/942 is marked as "wontfix". Not sure it's going to get seen

However, I believe this should get fixed (and there's an inconsistency in that extra props are passed to components, but not to render functions or child-as-functions). Shall we open a new issue/PR?

kelkes:
sure. lets do that!

christianchown:
As @adsingh seems to have dropped off the mat, I'll fork his branch to add the required change from: https://github.com/jaredpalmer/formik/pull/957#issuecomment-430537213

PR #964, opened 2018-10-05
Handle reset by default in Form

nbouscal:
This will make `<button type="reset">` inside the form trigger the form's reset handler. This seems like expected behavior for the `Form` component, because it's almost the exact same functionality as the `submit` case.

I'm not sure which documentation changes you'll want to make if you accept this. I've made the one that seems most clearcut.

### DIFF
--- a/docs/api/form.md
+++ b/docs/api/form.md
@@ -4,12 +4,12 @@ title: <Form />
 custom_edit_url: https://github.com/jaredpalmer/formik/edit/master/docs/api/form.md
 ---
 
-Form is a small wrapper around an HTML `<form>` element that automatically hooks into Formik's `handleSubmit`. All other props are passed directly through to the DOM node.
+Form is a small wrapper around an HTML `<form>` element that automatically hooks into Formik's `handleSubmit` and `handleReset`. All other props are passed directly through to the DOM node.
 
 ```jsx
 // so...
 <Form />
 
 // is identical to this...
-<form onSubmit={formikProps.handleSubmit} {...props} />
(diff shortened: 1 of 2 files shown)


docs/api/form.md
```diff
@@ -4,12 +4,12 @@ title: <Form />
 custom_edit_url: https://github.com/jaredpalmer/formik/edit/master/docs/api/form.md
 ---
 
-Form is a small wrapper around an HTML `<form>` element that automatically hooks into Formik's `handleSubmit`. All other props are passed directly through to the DOM node.
+Form is a small wrapper around an HTML `<form>` element that automatically hooks into Formik's `handleReset` and `handleSubmit`. All other props are passed directly through to the DOM node.
```
jaredpalmer: I would reverse the order here. handleSubmit should come first and reset second?

nbouscal: Yep that sounds good, I just went with alphabetical but agree that submit is more important/common and should go first.

jaredpalmer: Question for you. How does this work with buttons without `type`? Just confirming that submit still takes precedence, yes?

nbouscal: It should be the same default behavior, which iirc is browser-dependent. I haven’t tested it super thoroughly, but this should only affect buttons that explicitly have type reset. (No browsers default to reset, only submit or button.)

jaredpalmer:
@eonwhite Are we good with this?

jaredpalmer:
Will cut a release.

jaredpalmer:
Tomorrow.

Issue #946, opened 2018-10-01
Add Rich Harris' Treeshake rollup plugin to build


jaredpalmer:
## Current Behavior

Formik doesn't ensure that it can be treeshaken.
## Desired Behavior

Formik can always be treeshaken. 
## Suggested Solution

Add the rollup plugin and fail the build if formik can't treeshake.

## Who does this impact? Who is this for?

Everyone

TrySound:
Rich's tool is just rollup under the hood which is not able to treeshake transpiled classes at all.

PR #973, opened 2018-10-08
add tree-shaking to exports (#946)

Saifadin:
Adding tree-shaking to Formik according to (#946 ). This will allow better bundling for users out there, whenever they only import parts oft he application.

@jaredpalmer input is welcome

### DIFF
--- a/src/ErrorMessage.tsx
+++ b/src/ErrorMessage.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { FormikContext } from './types';
 import { getIn, isFunction } from './utils';
-import { connect } from './connect';
+import connect from './connect';
 
 export interface ErrorMessageProps {
   name: string;
@@ -47,7 +47,7 @@ class ErrorMessageImpl extends React.Component<
   }
 }
 
-export const ErrorMessage = connect<
+export default connect<
   ErrorMessageProps,
   ErrorMessageProps & { formik: FormikContext<any> }
 >(ErrorMessageImpl);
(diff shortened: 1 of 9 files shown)


TrySound:
@Saifadin Your changes doesn't make the code treeshakable. Default exports does not change anything.

Saifadin:
Okay good to know, what exactly does then? Any pointers would be helpful ^^

TrySound:
https://github.com/jaredpalmer/formik/pull/974/files#diff-29fd455fe4a1c8d19c5e5b1420ce14faR26

TrySound:
It's merged. Try to rebase to see results.

TrySound:
@jaredpalmer Why you don't use default exports?

jaredpalmer:
@TrySound Akin FB Haste, we believe that everything should have a unique name in a codebase. Default exports allow people to name things differently across files. So we almost exclusive use named exports a they are more explicit and have better autocomplete in TS.

TrySound:
@jaredpalmer Agree. I'm not TS user however still think default exports are quite useless and introduce even more problems like interop between cjs and esm.

https://github.com/oliviertassinari/react-event-listener/pull/79

jaredpalmer:
aside from being a huge breaking change, stylistically this can't be merged. Are there other ways to ensure tree-shaking? Or other bundle optimizations? Should we allow preshaking by altering the builds?

Saifadin:
I will give it another try today.

TrySound:
https://github.com/jaredpalmer/formik/pull/976

jaredpalmer:
Can i close?

Saifadin:
yes

PR #974, opened 2018-10-08
Add size snapshot

TrySound:
It will proove any actions from users who want to make the library
treeshakable.

Ref https://github.com/jaredpalmer/formik/pull/973 https://github.com/jaredpalmer/formik/issues/946


jaredpalmer:
Awesome.

PR #976, opened 2018-10-09
Annotate pure calls to enable treeshaking

TrySound:
In this diff I tweaked esm output with PURE annotations like this

```
var _a;
var FormikProvider = (_a =
/*#__PURE__*/
createContext({}), _a.Provider),
    FormikConsumer = _a.Consumer;
function connect(Comp) {
  var C = function (props) {
    return createElement(FormikConsumer, null, function (formik) {
      return createElement(Comp, __assign({}, props, {
        formik: formik
      }));
    });
  };

  C.WrappedComponent = Comp;
  return hoistNonReactStatics(C, Comp);
}

```

They will allow webpack and uglify to eliminate more unused code. Take a
look at size snapshot.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -15,21 +15,21 @@
     "gzipped": 5533
   },
   "dist/index.js": {
-    "bundled": 41589,
+    "bundled": 38223,
     "minified": 21696,
     "gzipped": 5577
   },
   "dist/formik.esm.js": {
-    "bundled": 40381,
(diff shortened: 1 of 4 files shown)


TrySound:
Babel will be helpful for one more optimization. So it's worth to have it in build step.

jaredpalmer:
SWEEET

jaredpalmer:
@TrySound Babel will be great. Thank you for setting up.

PR #983, opened 2018-10-10
setIn: keep the same object if nothing is changed

Anber:
As it was mentioned in https://github.com/jaredpalmer/formik/pull/646, `setIn` should return the same object if the new value is equal to current.


jaredpalmer:
@slightlytyler Does this look okay?

PR #993, opened 2018-10-12
URL to Docs from Help page fixed

Saifadin:
Issue was that the current help-page leads to `https://jaredpalmer.com/formik/docs/en/overview.html` But the correct url is `https://jaredpalmer.com/formik/docs/overview.html`.

The language part is currently non-existent for the docs, therefore a language based url to the docs, does not make sense.


Saifadin:
@jaredpalmer if you want a small refactor to separate the docs by language, so adding an `en` doc, I can do that in a separate PR if needed.

PR #994, opened 2018-10-12
Don't try to alter not existing errors or touches

dlehmhus:
Only alter the errors or touched array if an error or touched array exists. Otherwise, a new error or touched array would get created containing the value null.


dlehmhus:
@jaredpalmer, I think this is still a relevant problem!

PR #995, opened 2018-10-12
Return result of field level validation

dlehmhus:
Return error object when running field level validation.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -153,9 +153,9 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
   /**
    * Run field level validation
    */
-  validateField = (field: string) => {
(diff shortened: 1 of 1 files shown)


prichodko:
Please explain reasoning behind this PR and why reading it from Formik state is not sufficient.

dlehmhus:
@prichodko you are right, you could also check Formik error state. But since validateField doesn't return anything when validation is complete you wouldn't know if Formik state is already containing the field errors you where checking for with validateField or not. In this case, the only solution I can think of (and don't really like) would be to check periodically Formiks isValidating state. 
Since validateForm returns a Promise, I think validateField should do the same. When you only want to show an error message under a field after calling validateField, there is probably no need for that. But if you want to modify some state after calling validateField, a Promise would come in handy!

PR #1001, opened 2018-10-14
Add warnings support

prichodko:
Documentation is missing, but I think it would be better to tackle it closer to v2 release.


ematipico:
Quick question: in which part of formik workflow we can fill the warning messages? At the moment we can fill errors through the `validate` or `validationSchema`. Should have something similar as well?

jaredpalmer:
@ematipico is onto something. TBH without some equivalient of validationSchema or validate for warnings this PR could be perhaps just replaced by adding a `setFieldStatus(name: string, value: any)` and documenting warnings as a use case.

prichodko:
@ematipico great point, thanks.

@jaredpalmer yeah that makes sense, I will narrow it down for now. Although it feels a bit cumbersome when we consider setting warnings on `onSubmit` or removing warnings on `onBlur`.

prichodko:
Updated. 

I have to do this check  - https://github.com/jaredpalmer/formik/pull/1001/files#diff-18d1f112a814aca08509a74bded25fbfR497, because default value for status is `undefined`, maybe it would make sense initialize as an object?

PR #1005, opened 2018-10-15
v2

jaredpalmer:


prichodko:
I haven't really dug into tests recently, but I remember that some tests are using `enzyme` and some tests are using `ReactDOM` - we should definitely unify that. 

Could you elaborate why do you want to get rid of `enzyme`?

jaredpalmer:
@prichodko we can do all of our tests with just ReactDOM.

Saifadin:
@jaredpalmer what are your exact plans for this? Maybe a to-do list for single steps would be beneficial for contributions.

guilhermedecampo:
I agree with @Saifadin. Would be awesome to know the next steps indeed!

jaredpalmer:
See #1063

Issue #718, opened 2018-06-28
withFormik: Provide a namespace where to inject the props


ThiefMaster:
Currently all props are injected on the top level. It would be nice if there was an option to specify a namespace, just like [`propNamespace` in redux-form](https://redux-form.com/7.4.2/docs/api/reduxform.md/#-code-propnamespace-string-code-optional-).

PR #1025, opened 2018-10-19
Adds a namespace option on withFormik, to consolidate props

tgriesser:
Pretty self-explanatory, instead of spreading the formikProps into the component, this gives the option of providing them as a single prop, either `formik` or another if a string is passed for the option. 

Addresses #718

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,35 +1,35 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 175856,
-    "minified": 47562,
-    "gzipped": 13810
+    "bundled": 176571,
+    "minified": 47789,
+    "gzipped": 13890
   },
   "./dist/formik.cjs.production.js": {
-    "bundled": 40587,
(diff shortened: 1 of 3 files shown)


jaredpalmer:
Having a single prop that alters the shape of the render is a code smell. This can also be implmnted in user land— You can write your own hoc by using `<FormikConsumer>` that has this behavior if you want to.

tgriesser:
> Having a single prop that alters the shape of the render is a code smell.

Totally agree for normal components, like `<Formik />`, but in this case I wouldn't consider it a prop altering the shape of the render, but a config option altering the shape of the component, since it's an HOC.

The issue is that when you're composing components w/ recompose you generally don't want to pass a ton of props unnecessarily, I would consider that a code smell as well. I'd actually be fine with an option to disable passing *any* formik props and then using a custom `FormikConsumer` as you suggested to get what is needed further down in the context, but as it stands I don't see how what you suggested helps.

PR #1052, opened 2018-10-26
Pass form to field validation

joefiorini:
Continues #743, which was out of date from master. @jaredpalmer Any chance of getting this merged? We really like the field-level validation functions formik provides, but need this to do some of our validations. Thanks!


idoo:
@jaredpalmer any chance to get it merged?

Issue #850, opened 2018-08-27
Additions needed to make FormikErrors and FormikTouched typings account for field arrays


bbenezech:
Hi Jared!

https://github.com/jaredpalmer/formik/blob/master/src/types.tsx#L13-L15
should read something like

```typescript
export type FormikErrors<Values> = {
  [K in keyof Values]?: Values[K] extends any[]
    ? FormikErrors<Values[K][number]>[] // [number] is the special sauce to get the type of array's element. More here https://github.com/Microsoft/TypeScript/pull/21316
    : Values[K] extends object ? FormikErrors<Values[K]> : string
};
```

Same for FormikTouched I guess.

Cheers and millions of thanks for your hard work!

josephluck:
Agreed on this, running in to an issue in our project atm with a field of type `string[]` which should have an error property of `string` rather than `FormikErrors<Values[K]>` - this looks like a good fix!

jaredpalmer:
Feel free to submit a PR

frantic1048:
To be more strict, may we limit the definition of  "Array" to a normal typed Array(`T[]`), not Tuple(`[T1, T2, T3, ...]`) ?

And I think we should only take care of `string` key of an `object`.

```ts
type FormikErrors<Values> =
  Values extends Array<infer T> ? Array<FormikErrors<T>>
  : Values extends object ? { [K in Extract<keyof Values, string>]?: FormikErrors<Values[K]> }
  : string
```

JSoet:
Reopen - I just ran into this too... 

But I'm not sure that the solution mentioned above is correct, if you have an object in your values like
```
type tempType {
  arrayVal: number[]
}
```

then `errors.arrayVal` will be of type `number[]`? Which doesn't seem correct to me?

For me with a schema like the following:
```
validationSchema: Yup.object().shape({
    arrayVal: Yup.array()
      .required("required")
      .of(Yup.number())
      .min(1, "Minimum value: 1"),
  }),
```

Then it gives me a single string in `errors.arrayVal`. In which case the type should be 
```
export declare type FormikErrors<Values> = {
    [K in keyof Values]?: Values[K]  extends any[] ? string[] : Values[K] extends object ? FormikErrors<Values[K]> : string;
};
```

But I'm not sure if there's a way to create a different type of schema where you can get a different result from yup depending on how you create your validation schema?

@bbenezech @josephluck @frantic1048 ?

JSoet:
[...]

PR #1053, opened 2018-10-28
Support arrays in FormikErrors and FormikTouched typescript typings

christopherthielen:
This allows typescript typing of `FormikErrors` to support array fields.
Taken from https://github.com/jaredpalmer/formik/issues/850#issue-354255981

Closes https://github.com/jaredpalmer/formik/issues/850

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -11,18 +11,18 @@ export interface FormikValues {
  * Should be always be and object of strings, but any is allowed to support i18n libraries.
  */
 export type FormikErrors<Values> = {
-  [K in keyof Values]?: Values[K] extends object
-    ? FormikErrors<Values[K]>
-    : string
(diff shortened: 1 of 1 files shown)


christopherthielen:
Not stale

JSoet:
Also just ran into this ... but I think it should be a type like this:
```
export declare type FormikErrors<Values> = {
    [K in keyof Values]?: Values[K]  extends any[] ? string[] : Values[K] extends object ? FormikErrors<Values[K]> : string;
};
```
Explained myself in #850, not sure maybe there's other ways to get different results back depending on your yup schema?

JSoet:
@christopherthielen See my comment above, I think it's correct.... or have you tried with the type that you currently have in this PR and found that it works for you?

Issue #691, opened 2018-06-13
Make initialValues optional, touch Field on submit


charlax:
**Is your feature request related to a problem? Please describe.**

For extra context, see #445

I'm trying to port some of my simple forms from `redux-form` to `formik`. One first issue I had was that `initialValues` is required.

I'm using some HOC that creates full-fledged forms from a component describe only the form fields (e.g. for a user profile: first name, last name, password, etc.). I find that it is duplicate work to define all fields in the form component, and then to redefine all those keys with an empty value (most often an empty string `''`) - especially since `redux-form` gets away without it.

Something that looks like this:

```javascript
// UserProfilePage.js

import UserProfileForm from './UserProfileForm'

// ...

export default createPage({component: UserProfileForm, url: '/users/self', ... })
```

```javascript
// UserProfileForm.js

export default const UserProfileFrom = () => (
  <div>
    <Field name="firstname" label="First name" /> 
     ...
  </div>
);

```

Formik would have me do this instead:

```javascript
// UserProfilePage.js

// ...

export default createPage({
  component: UserProfileForm,
  url: '/users/self',
  initialValues: { firstname: '', lastname: '' },
  // ...
});
```

... which is pretty redundant.

**Describe the solution you'd like**

I think `initialValues` and `mapPropsToValues` could be made optional, defaulting all values to an empty string `''`.

On top of that, fields defined using `Field` could be registered so that when the form is submitted, they are marked as touched. Currently, they are not, which means that any error that is for a field that wasn't explicitly defined in `initialValues` or `mapPropsToValues` will not be shown - you can see an example here: https://codesandbox.io/s/1o7lkpmk8j

**Describe alternatives you've considered**

For the second issue (auto-touching `Field`), an alternative could be that the developer sets all the keys returned in the errors as touched, but that seems very hacky.

Let me know what you think!

jaredpalmer:
This will now be possible thanks to #684 . We'll now be able to construct initialValues (if not specified) by looking at what fields have mounted and assuming they are just strings.

Andreyco:
@charlax hey, this is now possible with beta 5! We should probably mark `initialValues` optional for Typescript users (you still have to pass at least empty object) but I am afraid people start to complain why fields are not marked as touched (when not using `Field` component)

peteruithoven:
[...]

PR #1055, opened 2018-10-29
Update initialValues to be optional

wachunga:
Been using Formik for a while now and it would definitely be convenient (less 😭) to have `initialValues` be optional, defaulting to an empty string as per #691.

This is my first time contributing to Formik so I could use some guidance. The logical place to do this seems to be `registerField`, but this approach still leads to the React warning about an uncontrolled input becoming controlled because in the initial `Field.render`, the value is undefined.

Suggestions? The only thing I can think of is calling `registerField` earlier in the lifecycle, eg `getDerivedStateFromProps` instead of `componentDidMount`.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -25,6 +25,8 @@ import {
   getIn,
 } from './utils';
 
+const DEFAULT_INITIAL_VALUE = '';
+
 export class Formik<Values = object, ExtraProps = {}> extends React.Component<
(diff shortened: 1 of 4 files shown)


wachunga:
I just noticed that field registration used to happen earlier prior to 1.3.1. 🤔 

@jaredpalmer or @Andreyco: did you have an approach in mind for optional initialValues that I'm overlooking?

Issue #1046, opened 2018-10-25
Hooks Rewrite


jaredpalmer:
```jsx
import React, {
  useContext,
  useState,
  createContext,
  createElement,
} from 'react';

const FormikContext = createContext(null);

export function Formik({
  initialValues,
  validate,
  onSubmit,
  validateOnBlur = true,
  validateOnChange = true,
  ...props
}) {
  const [values, updateValues] = useState(initialValues);
  const [errors, updateErrors] = useState({});
  const [touched, updateTouched] = useState({});
  const [submitAttemptCount, updateSubmitAttemptCount] = useState(0);
  const [isSubmitting, updateIsSubmitting] = useState(false);
  const [isValidating, updateIsValidating] = useState(false);

  function validateForm(vals = values) {
    updateIsValidating(true);
    return Promise.resolve(validate ? validate(vals) : {})
      .then(x => x, e => e)
      .then(e => {
        // return or take a callback?
        updateErrors(e);
        updateIsValidating(false);
      });
  }

  function getFieldProps(name, type) {
    return {
      value:
        type === 'radio' || type === 'checkbox'
          ? undefined // React uses checked={} for these inputs
          : values[name],
      onChange(e) {
        e.persist();
        updateValues(
          prevValues => ({ ...prevValues, [name]: e.target.value }),
          () => {
            if (validateOnChange && validate) {
              validateForm();
            }
          }
        );
      },
      onBlur() {
        updateTouched(
          prevTouched => ({ ...prevTouched, [name]: true }),
          () => {
            if (validateOnChange && validate) {
              validateForm();
            }
          }
        );
      },
    };
  }

  async function submitForm() {
    updateTouched(setNestedObjectValues(values, true));
    updateIsSubmitting(true);
    updateSubmitAttemptCount(prev => prev++);
    try {
      await validateForm();
      const errors = await onSubmit(values);
      if (errors) {
        updateErrors(errors);
      }
      updateIsSubmitting(false);
    } catch (errors) {
      updateErrors(errors);
      updateIsSubmitting(false);
    }
  }

  function handleSubmit(e) {
    e.preventDefault();
    submitForm();
  }

  const ctx = {
    values,
    updateValues,
    errors,
    updateErrors,
    touched,
    updateTouched,
    submitAttemptCount,
    updateSubmitAttemptCount,
    isSubmitting,
    updateIsSubmitting,
    isValidating,
    validateOnChange,
    validateOnBlur,
    getFieldProps,
[...]

Issue #1026, opened 2018-10-20
Using styled components with formik is incredibly slow


mjangir:
I'm just playing with formik library and found performance issue with custom component in Field tag when the custom component uses styled components heavily.

When I keep pressing the key in an input box which is wrapped in multiple hierarchy of styled components, it causes huge lagging but instead of that, if I use simple divs, it is comparatively faster.

joycollector:
@mjangir We are having the same issues. Using debounce to handle this.

yashaRO:
I may not have as many styles as you do, but I do have all custom components. I saw this in another thread and it turned out to be true
https://github.com/jaredpalmer/formik/issues/671#issuecomment-424792264

johnmcdowall:
I'm seeing the same issue with one input in a form which is a styled component - pressing and holding any key causes a large and noticeable lag

johnmcdowall:
Incidentally, I've tried using `<FastField />` and it hasn't made any difference.

dontmakemethink:
I have a form with 10 fields and I'm encountering the same issue even with `<FastField />`.

flynnham:
@joycollector, where are you placing the `debounce`?

joshperrin:
I am having this issue as well. I tried <FastField /> with no success either.

I was thinking I may have to use vanilla CSS with Formik to fix this bug.

devjones:
I've had similar major performance issues with relatively short forms: a custom drop down and 4-5 custom input fields.  I'm using tailwind.js (similar to styled components) and Yup for validation.

dgopsq:
Same issue here, I can confirm that production is not affected. It lags only in dev mode.

jaredpalmer:
That’s interesting. Can you / someone reproduce this behavior in a sandbox?

dgopsq:
*I haven't check*, but trying around a bit I think that production is affected too, but way to less than dev.
BTW I solved my problem setting the form value from `onBlur` instead of `onChange`.

idanhen:
Same issue here, production is effected - but way less

grachet:
same

seeden:
same issue.

jaredpalmer:
This should be fixed with 1.4.0. https://github.com/jaredpalmer/formik/releases/tag/v1.4.0

devjones:
Thanks!

johnmcdowall:
[...]

Issue #671, opened 2018-06-01
Slow on large form.


joonhocho:
I have a large form with 20~30 fields.
It seems formik becomes slow when there are large number of fields.
It seems to validate every field on every keystroke.
Is there a good way to mitigate this issue?

jaredpalmer:
Use `<FastField />`

joonhocho:
@jaredpalmer  Can I use FastField in place for all Field without any changes? Any documentations on this? Thank you!

jaredpalmer:
`<FastField>` is best used on pure fields that don't impact other fields either by changing another field's value or error message. State is kept locally in FastField and then synced up to the top level `<Formik/>` component once onBlur is called.

joonhocho:
@jaredpalmer Thank you I will try and report back if there are any issues.

hmaurer:
One issue I have found with `<FastField>` is that since the state is synced on blur, if the form is submitted by pressing `Enter` in a field, the latest value of that field won't be synced before submit. That's particularly annoying, and I haven't yet found a work-around that doesn't involve patching FastField :(

derekcannon:
Could you wrap `handleChange` in a function that triggers `onBlur` if the `Enter` key is pressed?

jaredpalmer:
You can use shouldComponentUpdate to wrap Field to optimize perf

Virgomax:
I have the same issue than @hmaurer with `<FastField>`. Please, is there going to be a fix? or a nice workaround?

mdmad7:
using create react app and semantic ui i built ~40 fields. input is very slow whilst running the app in dev mode. but when i run yarn build and run the app with serve -s build, there is no delay in input fields. can i expect the same results in whilst running on a production server?

jaredpalmer:
Use FastField. It works now.

rohanBagchi:
Try this maybe:
https://github.com/jaredpalmer/formik/issues/1026#issuecomment-474691215

anand-neema:
I have 1206 fields on the page and i am using FastField as well and it's veryyyyy slow any suggestion to improve this?

Basically i am using table and all 200 rows are inline editable with 6 columns. That takes too long to compare in shouldUpdate

wellyal:
It may be a noob question. Even though how can I use FastField If I use a third part input component like material-ui's TextField

johnrom:
[...]

PR #1063, opened 2018-10-30
[v2] Hooks

jaredpalmer:
## Improvements

- [x] Create `<Field as>` prop (#1406)
- [x] Add `meta` prop to `<Field children>` (#1406)
- [x] Break out `useFormik` from `<Formik>`
- [x] `onReset` and `handleReset` tests
- [x] async onSubmit (no more `setSubmitting(false)` needed)
- [x] Make validation cancellable (#1026, #671)
- [x] `useField`'s second element is field metadata (#343)
- [x] Add `initialErrors`, `initialTouched`, `initialStatus` (#1410)
- [x] Fix `isValid` (#1410)
- [ ] `<Field >`'s uses `handleBlur('path')` and `handleChange('path')` override 
(#1113)
- [x] Upgrade to TypeScript 3.5 and React 16.9 (#1559)

## Breaking Changes

- Breaking TypeScript change: `FormikActions` have been renamed to `FormikHelpers`
- Formik no longer uses `console.error` in its warnings, but rather `console.warn`

## Deprecations

These are just warnings for now.

- [x] `<Field render>` (#1406)
- [x] `<Field component>` (#1406)
- [x] `isInitialValid` and fix `isValid` (#1410)

## Documentation

- [ ] `useField`
- [ ] "Creating your own `<FieldGroup>`"
- [ ] Checkboxes and Radio inputs
- [ ] `<FormikConsumer>`
- [ ] `useFormikContext`
- [ ] `useFormik`
- [ ] Rewrite examples with hooks
- [ ] Rewrite tutorial with hooks
- [ ] Migration Guide
- [ ] Blog post

Feel free to start making PR's to this branch. This is going to take a minute to figure out as there are a LOT of new paradigms. 

#1046 #1005


weyert:
Why the rename of `FormikContext` to `FormikCtx`?

tgriesser:
I don't have the time to make this into a formal PR right now but maybe it'll provide a little inspiration. It's a proof-of-concept of Formik working with hooks. Because of the imperative API exposed by formik, the key I found is to mostly use refs internally rather than state, otherwise the values will get stale, e.g. if you set a field and call submitForm in the same event handler. It's just a first pass, so I'm sure there's plenty of room for improvement, but it seems to work alright:

https://gist.github.com/tgriesser/fdd08da075fec80328792319606c76e5

jaredpalmer:
@tgriesser hrm. that's definitely a footgun with calling submit form in the same handler. i wonder if there is a way to avoid that aside from ref + force update. refs won't show up in dev tools for example.

weyert:
How can I extend it with warnings?

jaredpalmer:
If we like the reducer approach then you'll be able to do whatever you want if we decide to allow a `stateReducer` prop. However, that debate will come later.

jaredpalmer:
Was on vacation last week. I see a clear path forward to releasing an alpha, but it will have breaking changes unless there is a hooks equivalent to the second argument callback of setState in useReducer (https://github.com/facebook/react/issues/14221).

There has been no further comments on my issue. However, in writing it all up I discovered perhaps a better solution is to remove the automagical validation behavior when calling `setXXXX`. Instead, these just do what they say and then it is up to you to then call `validate()` in your custom handlers. The other option is to remove the ability override the top-level `validateOnChange` and `validateOnBlur`

weyert:
Hey @jaredpalmer. I was wondering what the status of the hooks version is. Are you happy with the current state of it? Can we start using it for non-production use?

jaredpalmer:
It's pretty close. FieldArray probably doesn't work yet. But Field and Formik and Form will work.

jaredpalmer:
The types are messed up but will be able to fix that shortly. Also the 3rd argument to setFieldValue and setFieldTouched no longer works.

jaredpalmer:
I am seriously considering removing field-level validation. Aside from the fact that we don't use it, it requires child-parent registration which is quite awkward and also does not work with the prop-only API. Also it is the cause of the performance bottleneck with styled-components and various other issues. Need to think it through over the next few days.

jaredpalmer:
For those interested in hacking with stuff. You can play with this now... @prichodko @slightlytyler 

```sh
yarn add formik@canary
```

dan-kez:
@jaredpalmer - happy to test this out but I'm not entirely sure of the API. If you have minute could you post a small example of your intended use? E.g. just a form with a validation and submit function

seeden:
I have tried current version with just few inputs and one fieldarray. It is working but it very very slow.

jaredpalmer:
Yeah something is wrong. the alpha is very very slow.

weyert:
Interesting, why would it be slow? Maybe an idea to write some tests for it?

sshmyg:
Who needs v2, if v1 has 107 open issues?
- all form validation on every field change https://github.com/jaredpalmer/formik/issues/512
- yup transform doesn't work https://github.com/jaredpalmer/formik/pull/728
...

jaredpalmer:
@sshmyg thanks for the kind words, support, and also for all of your contributions too! in all seriousness, Yup transform doesn't make sense in the context of Formik. If we did this we'd have to wait for Yup to transform before ever single key stroke, change to values, and also run it prior to mount. Furthermore, validationSchema is just a thin wrapper around the same behavior as validate which _does_ get called on every keystroke. If you want to transform your values, you should do so in your field's handle change and render functions so that they are encapsulated and localized. As for validating one field at a time, this too has many problems. For example, this would not work if you are trying to match `email` and `confirmEmail` fields with Yup. Open suggestions and PRs!

jaredpalmer:
@sshmyg This PR also literally implements single field Yup execution using `yup.validateAt()` 😂 😂

christopherthielen:
> I am seriously considering removing field-level validation

Chiming in as a user of field-level validation.  For many our screens, we've found field level validation to be more ergonomic than schema or form-level validation.

k15a:
@jaredpalmer

> Field-level registration (need to see if this is actually worth it)

What do you mean by that? If I read the merge request correctly you handle point already?

Also, you have a small bug in the CodeSandbox. `setState` from `useState` is still asynchronous so you need to call it with the callback form. Here is a fixed version: https://codesandbox.io/s/q7ov3n0296

src/Field.tsx
```diff
@@ -82,110 +77,85 @@ export interface FieldConfig {
...
-        ...rest,
-        children,
-      });
-    }
-
-    return React.createElement(component as any, {
-      ...bag,
-      ...props,
+  if (typeof component === 'string') {
+    const { innerRef, ...rest } = props;
```
k15a: Would be nice to use React.forwardRef here.

src/Formik.tsx
```diff
@@ -1,306 +1,240 @@
...
+  const fields = React.useRef<{
+    [field: string]: {
+      validate: (value: any) => string | Promise<string> | undefined;
+    };
+  }>({});
+  React.useEffect(
+    () => {
+      initialValues.current = props.initialValues;
+    },
+    [props.initialValues]
```
TrySound: What this for? Why not just assign in render without effect?

jaredpalmer: Yeah, you are right

src/Formik.tsx
```diff
@@ -1,306 +1,240 @@
...
   setIn,
+  isEmptyChildren,
+  isPromise,
   setNestedObjectValues,
   getActiveElement,
   getIn,
   makeCancelable,
 } from './utils';
+import { FormikProvider } from './FormikContext';
+import warning from 'warning';
```
TrySound: Can this be replaced with tiny-warning? It has esm support.

weyert:
@jaredpalmer I was wondering if there is any way I could help you with this ticket?

developer239:
Why the rename of FormikContext to FormikCtx? Especially when it introduces breaking change? 🙂

jaredpalmer:
@developer239 that will be unwound in the final release. Internally there used to be a difference between FormikContext and FormikProps TS types. However, FormikContext is no longer in use as a type

src/Formik.tsx
jaredpalmer: Fixed.

src/Formik.tsx
jaredpalmer: Fixed.

nhuesmann:
@jaredpalmer 

> New canary has async `onSubmit` (so you don't need to call `setSubmitting` anymore)

I've been playing with canary build `^2.0.0-canary.effcf9b` in React Native and have found that neither `isSubmitting` nor `isValidating` are being toggled back to `false` after an aborted submission due to validation errors. I'm passing in my custom submit handler to the `onSubmit` config option when calling `useFormik`, then I'm utilizing the destructured `submitForm` function as my RN Button's `onPress` prop. I can't even figure out how to leverage `setSubmitting` to manually set it to `false` since I'm deferring my validation logic to Formik by supplying a Yup schema to `validationSchema`, so the validation failure aborts before my submit handler even gets called.

Any ideas what I could be doing wrong? Is this just a RN issue? I should mention I'm just using basic RN components, no component libs. Other than this I've been loving using Formik with hooks, it's so much cleaner. Thanks in advance, you're killing it man!

src/Field.tsx
```diff
@@ -83,110 +59,108 @@ export interface FieldConfig {
...
     );
   }
 
-  componentDidMount() {
-    // Register the Field with the parent Formik. Parent will cycle through
-    // registered Field's validate fns right prior to submit
-    this.props.formik.registerField(this.props.name, this);
-  }
+  return formik.getFieldProps(name, type);
+}
```
crosscompile: Should `useField` take an optional `validate` function and contain [register logic similar to `Field`](https://github.com/jaredpalmer/formik/blob/hooks/src/Field.tsx#L104-L114)?

jaredpalmer: Yes.

src/FormikContext.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import { FormikContext } from './types';
+
+const PrivateFormikContext = React.createContext<FormikContext<any>>({} as any);
+export const FormikProvider = PrivateFormikContext.Provider;
+export const FormikConsumer = PrivateFormikContext.Consumer;
+
+export function useFormikContext<Values>() {
+  return React.useContext<FormikContext<Values>>(PrivateFormikContext);
+}
```
crosscompile: I'm not sure if this would just be adding more API without a good use case, but I wonder if it would be worth it to have a hooks only API, something like:

```js
const App = () => {
  const [formik] = useFormik();
  const [form] = useForm(formik);
  const [email] = useField("email", formik);
  const [password] = useField("password", formik);

  return (
    <form {...form}>
      <input {...email} />
      <input {...password} />
    </form>
  );
}
```
And then `FormikContext` just puts the values from `useFormik` on context for that use case.

jaredpalmer: I don’t see this how this api is useful. You will end up having to prop drill.

jaredpalmer: useForm is not a useful hook either IMHO

jaredpalmer: 99.999% of our `<Form>` components have 0 additional props, which to me means it’s a very good abstraction.

jaredpalmer: my goal for useField is to make building your own Fieldset the goal. Eventually, perhaps even moving `<Field>` into its own package a la prop-types

jaredpalmer: it goes without saying that your API would be faster and more perform at because it does not hook into context. however, what is the point of using hooks here aside from useFormik??? just make prop getters and spread?

jaredpalmer: along those lines... this PR already has prop getters. So we technically already support this API by exporting useFormik()

src/Formik.tsx
```diff
@@ -1,312 +1,232 @@
...
-    // Cancel validation on unmount.
-    if (this.validator) {
-      this.validator();
+  function registerField(name: string, { validate }: any) {
+    if (fields.current !== null) {
+      fields.current[name] = {
+        validate,
+      };
     }
   }
```
crosscompile: I think we might want to wrap all these functions defined in `useFormik` with `useCallback` so that we're not creating new functions on every render.

src/FormikContext.tsx
crosscompile: Yeah I agree with everything you said, it makes the API less usable.

src/Formik.tsx
crosscompile: But also with ~26 functions that could be a real pain to manage.

I wonder if it's possible to separate the internals of `useFormik` into logical parts like validation and values.  I might experiment with that and report back.

jaredpalmer: Let’s start with useCallback. Let’s discuss how to slice it up. I agree it’s a lot.

jaredpalmer: Will cut a new release in the meantime

jaredpalmer:
Just cut a new release.

sijad:
with `gamma.1` in react native I get this warning:

```
useField() / <Field /> must be used underneath a <Formik> component or withFormik() higher order component
```

I tried render function as `render` props and `children`, I'm calling `useFormikContext` in my component to have access to `handleChange` and it works.

src/Field.tsx
```diff
@@ -83,110 +59,108 @@ export interface FieldConfig {
...
-    );
-
-    warning(
-      !(component && children && isFunction(children)),
-      'You should not use <Field component> and <Field children> as a function in the same <Field> component; <Field component> will be ignored.'
-    );
+export function useField(name: string, type?: string) {
+  const formik = useFormikContext();
 
+  if (process.env.NODE_ENV !== 'production') {
```
sijad: it redundant with [tiny-warning](https://github.com/alexreardon/tiny-warning/blob/dc65095d4e27290869781c778950ed82a91de866/src/index.js#L2)

src/Field.tsx
```diff
@@ -83,110 +59,108 @@ export interface FieldConfig {
...
     warning(
-      !(render && children && !isEmptyChildren(children)),
-      'You should not use <Field render> and <Field children> in the same <Field> component; <Field children> will be ignored'
+      !formik,
```
sijad: it should be `formik` see:
https://github.com/alexreardon/tiny-warning/blob/dc65095d4e27290869781c778950ed82a91de866/src/index.js#L9

buzinas:
> with `gamma.1` in react native I get this warning:
> 
> ```
> useField() / <Field /> must be used underneath a <Formik> component or withFormik() higher order component
> ```
> I tried render function as `render` props and `children`, I'm calling `useFormikContext` in my component to have access to `handleChange` and it works.

Getting the same warning here, but I didn't understand. What was your solution exactly?

sijad:
@buzinas I've send an PR #1348

max-mykhailenko:
Does this changes compatible with flow types?

otorrillas:
@jaredpalmer I've been using `2.0.1-alpha` (not in prod! 🙄) and the experience has been fantastic.

Is there something we can help with to release a stable v2? Thank you!

src/Formik.tsx
```diff
@@ -418,235 +314,415 @@ export class Formik<Values = FormikValues> extends React.Component<
...
-      setFieldTouched: this.setFieldTouched,
-      setFieldValue: this.setFieldValue,
-      setStatus: this.setStatus,
-      setSubmitting: this.setSubmitting,
-      setTouched: this.setTouched,
-      setValues: this.setValues,
-      setFormikState: this.setFormikState,
+  function getFieldProps(
+    name: string,
+    type: string
```
cdoe: Couldn't `type` be optional here?
```suggestion
    type?: string
```

buzinas:
How do I create a form with `useFormik` and `useField`? From my understanding, `useFormik` replaces `<Formik />`, since it receives the same props and return the same stuff that the current `render` prop passes down to the `<form />`. The logic seems fine, but I get this error:

```
TypeError: formik.getFieldProps is not a function

 862 |     process.env.NODE_ENV !== "production" ? warning(formik, 'useField() / <Field /> must be used underneath a <Formik> component or withFormik() higher order component') : void 0;
  863 |   }
  864 | 
> 865 |   return formik.getFieldProps(name, type);
  866 | }
```

I already tried many different things and I'm out of ideas. The warning says we still need to use `<Formik>`, so I'm wondering what is `useFormik` is used for?

Thanks in advance!

cdoe:
@buzinas I'm just recently getting up to speed with the hooks version, so someone please correct me if I'm wrong...

To my understanding, `useFormik` isn't meant as a drop-in replacement for `<Formik />`.  In fact, `useFormik` is an altogether new and rather independent thing.  It isn't intended to coordinate with `useField` at all.  I stumbled onto [this doc](https://github.com/jaredpalmer/formik/blob/14ac079d9dada09292496ab42d2ad10915fc2715/docs/api/useField.md) which may help shed some light.  Hope that helps!

buzinas:
After reading [this tweet](https://twitter.com/jaredpalmer/status/1074736177356390406), I finally could understand what was going on. So now I could solve my problem by using the `FormikProvider`. Now it's finally pretty easy to create forms in React:

```jsx
// components/Form/index.js
import React from 'react'
import { useFormik, FormikProvider } from 'formik'
import Fieldset from '../Fieldset'

const Form = ({
  initialValues,
  validationSchema,
  onSubmit,
  children,
  ...props
}) => {
  const formik = useFormik({ initialValues, validationSchema, onSubmit })

  return (
    <FormikProvider value={formik}>
      <form onSubmit={formik.handleSubmit} {...props}>
        <Fieldset disabled={formik.isSubmitting}>{children}</Fieldset>
      </form>
    </FormikProvider>
  )
}

export default Form
```

```jsx
// components/InputField/index.js
import React from 'react'
import { useField } from 'formik'

import Field from '../Field'
import Input from '../Input'

const InputField = ({ label, name, onChange, errorText, ...props }) => {
  const [field, meta] = useField(name)

  return (
    <Field label={label} errorText={meta.touch && meta.error}>
      <Input
        name={name}
        {...field}
        {...props}
        error={!!(meta.touch && meta.error)}
      />
    </Field>
  )
}

export default InputField
```

And then to use it:

```jsx
import React from 'react'

import Form from '../components/Form'
import InputField from '../components/InputField'
import Button from '../components/Button'
import { signInSchema } from '../schemas/authentication'

const SignInForm = () => (
    <Form
      initialValues={{ email: '', password: '' }}
      validationSchema={signInSchema}
      onSubmit={(values, { setSubmitting }) => {
        // fetch('my-api/authenticate') and SUCCESS 🎉 
      }}
    >
      <InputField name="email" label="email" />
      <InputField type="password" name="password" label="password" />
      <Button type="submit" primary>
        Login
      </Button>
    </Form>
)

export default SignInForm
```

tz5514:
@jaredpalmer Thanks for this awesome works! 
I already use `v2.0.1-alpha.0` for few days, and I found that `enableReinitialize` prop is missing in `Formik.tsx` at v2.0, since it has been rewrited by hooks?

stunaz:
any documentation for v2?

quochuytlbk:
@buzinas OMG, that was pure magic.

bopfer:
@buzinas's example is awesome and really all I needed to start using Formik hooks now.  My project is months away from a production version.  So, we are jumping in now and we really love how elegant the hooks solution is for our forms.

The docs in the [hooks](https://github.com/jaredpalmer/formik/tree/hooks/docs/api) branch are partially updated, but definitely still a WIP.

tz5514:
> @jaredpalmer Thanks for this awesome works!
> I already use `v2.0.1-alpha.0` for few days, and I found that `enableReinitialize` feature is missing in `Formik.tsx` at v2.0, since it has been rewrited by hooks? Thanks!

@jaredpalmer I make a PR (https://github.com/jaredpalmer/formik/pull/1399) to fix this issue, 
for bringing back `enableReinitialize` feature in `useFormik` hook.  Hope this will help. 😃

clementoriol:
For anyone trying out the v2 alpha and getting confused about the form submitting part, note that onSubmit is now async and is expected to return a promise. So no more `setSubmitting(false)`.

```
onSubmit={(values, _actions) => {
  console.log('submitting...', values);
  return new Promise((resolve, reject) => {
    setTimeout(() => {
      console.log('submitted!');
      resolve();
    }, 3000);
  });
}}
```

It's specified in the top comment but it's an easy to miss it right now. :)

ematipico:
Is there a reason why `useField` doesn't return the formikbag? Unless it's accessible using another API...

jaredpalmer:
@ematipico `useFormikContext`

sijad:
is there a way to submit form programmatically? and accessing to `isSubmitting`?

src/Field.tsx
```diff
@@ -84,110 +66,108 @@ export interface FieldConfig {
...
-      'You should not use <Field component> and <Field render> in the same <Field> component; <Field component> will be ignored'
+export function useField<Val = any>(name: string, type?: string) {
+  const formik = useFormikContext();
+
+  invariant(
+    formik,
+    'useField() / <Field /> must be used underneath a <Formik> component or withFormik() higher order component'
+  );
+
+  return formik.getFieldProps<Val>(name, type);
```
p10ns11y: with this change one can't use `useField` to get `form` bag where the imperative methods such as `setXXXX` available. Those methods needed to use with third-party libs such as `react-select`.  Would you recommend using `Field` in that case?

bopfer: See the 3 comments right above this :)

src/FormikContext.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import { FormikContext } from './types';
+
+const PrivateFormikContext = React.createContext<FormikContext<any>>({} as any);
```
mfbx9da4: This question might be a bit of a naive question but how will formik support multiple forms? Won't the context always be the same and therefore rendering two forms will cause data to bleed?

CodeWitchBella: No, context only applies to subtree which is wrapped with the provider.

mfbx9da4: Thanks

clementoriol:
Hi! I'm having a lot a fun with the new hook proposals :)

Just curious, is there any way to pass a validate function to the useField hook ? https://github.com/jaredpalmer/formik/issues/1524 current behavior exemple made me think so but after testing that it doesn't look like it's already implemented.

Andreyco:
It is not implemented, yet

src/Formik.tsx
```diff
@@ -1,661 +1,847 @@
...
+    touched: props.initialTouched || {},
+    status: props.initialStatus,
+    isSubmitting: false,
+    isValidating: false,
+    submitCount: 0,
+  });
+  const prevState = usePrevious(state);
+
+  const runValidateHandler = React.useCallback(
+    (values: Values, field?: string): Promise<FormikErrors<Values>> => {
```
johnrom: This isn't particular to this method, but I do have a question about how these callbacks are implemented. And it's a big one (sorry).

I think that `useFormik`'s sole job is basically to gather the hooks, get the data, and hook up the context. But the majority of this function is actually taken up with defining other functions. In addition, the testability of these functions is now non-existent. Finally, it's nearly impossible to determine the control flow of the function itself without collapsing every function bit by bit. So I propose one general strategy with two possible methods:

Strategy: define helpers elsewhere. It could be in this file outside of the function, or a helpers file.

```
/** @private */
function runValidations(validate, values, field) { /* the function */ }

export function useFormik<Values = object>() {
    // etc etc
    // Strategy 1: Use bind (please don't hurt me). 
    const runValidateHandler = React.useCallback(runValidations.bind(null, props.validate), [props.validate]);
    // Strategy 2: Connect the callbacks manually. 
    // Unfortunately, I don't think you can infer values and field like I've done below, 
    // so this would require defining types twice.
    const runValidateHandler = React.useCallback((values, field) => {
        runValidations(props.validate, values, field);
    }, [props.validate]);
}
```

Now, you can unit test `runValidations` really easily without trying to "unit test React". I don't know if there is a way to spy on the function when using `bind` to use `expect(runValidations).toHaveBeenCalled()` or whatever, but if so that's even better.

The best part of this method is when you read this file, you'll just see the following, you can _read_ whether or not everything is hooked up in the correct order and let TypeScript do the rest, and finally a quick glance will allow you to compare that all of the props passed to the helper function are contained in the dependency array.

```
const state = useState(etc, etc);
const initialValues = useRef<>();
const runValidationsHandler = useCallback(callback, [state]);
const runValidationSchema = useCallback(callback, [initialValues]);
const allHandlers = [
    runValidationsHandler,
    runValidationSchema,
];

**Comment by FredyC** [comment]
So how is the validation planned to work with `useField`? Considering it currently accepts the `name` and `type` args. And why was such an important piece left out for now? Or are we supposed to just stick to `Field` component if we need validation? Doesn't sound very flexible.

File: src/Formik.tsx (hunk 12 above)
Comment (jaredpalmer) [reply src/Formik.tsx]: Testing these methods internally is not a goal. These are implementation details. There is a valid usage of `useFormik`. For folks that want to use the state only version. That's it. I think we should likely rename this function to something that is less obvious or mark it as `unstable_`

**Comment by jaredpalmer** [comment]
@FredyC `useField` is going to accept either a string name or and object with #1555

File: src/Formik.tsx (hunk 12 above)
Comment (johnrom) [reply src/Formik.tsx]: @jaredpalmer I agree with you for the most part. I re-read through the code and for the most part what we are doing in these callbacks is dealing with promises / dispatches.

However, there are some parts I think should be broken out and unit tested. I'll itemize those in a different issue.

**Comment by MatyCZ** [comment]
Will `useField` accept validation too? There is no way how to add validation function to field with `useField` hook.

**Comment by FredyC** [comment]
@MatyCZ Is it really that hard to at least read the previous comment 🙄 😆

**Comment by MatyCZ** [comment]
@FredyC Jared mentioned accepting an object with #1555 as answer to your question about validation. But #1555 is already merged and `useField` doesn`t accept validation as object prop.

**Comment by fabb** [comment]
Maybe remove the `status` functionality from the hook? Users could use the regular `useState` hook with the additional benefit that it‘s not of type `any`.

**Comment by alfred-pb** [comment]
Any progress with the docs?

**Comment by jaredpalmer** [comment]
I am working on docs this week

**Comment by GollyJer** [comment]
I realize the docs aren't ready, but does there happen to be a hooks code example sitting around somewhere? 😀

**Comment by alfred-pb** [comment]
@jaredpalmer thanks! A video will be also appreciated it 👍

**Comment by veenedu** [comment]
A demo, until we see something from @jaredpalmer 


`import React from "react";
import { Formik, useFormikContext } from "formik";

const MyForm = () => {
  const {
    handleSubmit,
    handleChange,
    values,
    errors,
    isSubmitting
  } = useFormikContext();
  console.log("isSubmitting", isSubmitting);
  return (
    <div>
      <h2>Using hooks</h2>
      <form
        onSubmit={e => {
          e.preventDefault();
          handleSubmit();
        }}
      >
        <div>
          Name <input name="name" value={values.name} onChange={handleChange} />
        </div>
        {errors.name && <div>{errors.name}</div>}
        <div>
          age
          <input name="age" value={values.age} onChange={handleChange} />
          {errors.age && <div>{errors.age}</div>}
        </div>
        <div>
          <input value="submit" type="submit" disabled={isSubmitting} />
        </div>
      </form>
    </div>
  );
};

const initialValues = {
  name: "Praveen",
  age: 35
};
const config = {
  initialValues,
  onSubmit: (values, { setSubmitting }) => {
    setTimeout(() => {
      console.log("Submit", values);
      console.log("Here");
      setSubmitting(false);
    }, 1400);
  },
  validate: values => {
    const errors = {};
    if (!values.name) {
      errors["name"] = "Name is required";
    }
    if (!values.age) {
      errors["age"] = "age is required";
    }
    if (values.age && isNaN(Number(values.age))) {
      errors["age"] = "Please enter a number";
    }
    return errors;
  }
};

const Wrapped = () => (
  <Formik {...config}>
    <MyForm />
  </Formik>
);

export default Wrapped;
`

https://codesandbox.io/s/serene-grothendieck-se2om

**Comment by fabb** [comment]
AFAIK the `Wrapped` part and especially `<Formik>` is unnecessary when you use the hook.

**Comment by fabb** [comment]
And formik handles the `preventDefault` for you when you pass the event to `handleSubmit`:
```
```

veenedu:
Then how to validation rules, initial values?

fabb:
You need to pass the config to the hook as a parameter:
https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx

tommedema:
@GoPro16 checked your code but don't see references to `useFormik` anywhere

GoPro16:
> @GoPro16 checked your code but don't see references to `useFormik` anywhere

We use the [useField](https://github.com/Availity/availity-react/blob/master/packages/form/src/Input.js#L8) hook and the [useFormikContext](https://github.com/Availity/availity-react/blob/master/packages/form/src/CheckboxGroup.js#L11) hook the most. We don't ever use `useFormik` because we don't need all of the formikbag, only just a sub selection like what `useFormikContext` returns.

septs:
@jaredpalmer release the feature to next version?

Issue #1050, opened 2018-10-26
Checkbox doesn't bind to initialValues


dem0nbruce:
### Current Behavior

When using `<Field type="checkbox />` and `initialValues` the checkbox doesn't bind.

### Expected behavior

`<Field type="checkbox" />` should bind to `initialValues`

### Reproducible example

https://codesandbox.io/s/ppy7wyrl0m

### Suggested solution(s)

See #1024

### Additional context

See #1024

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           |  1.3.1
| React            | 16.5.2

pahen:
This worked as a workaround for me.

```
const CheckboxInput = (props) => (
	<Field {...props} render={({field}) => <input {...field} type="checkbox" checked={field.value} />} />
);
```

RazerM:
I think this is still a problem.

In the following sandbox, the `initialValues` only works when `<Field />` isn't given `type="checkbox"`.

https://codesandbox.io/s/p7zlvk86lm

seunggs:
I also think this is still a problem - can't seem to set checkbox initial values using `initialValues`

madroneropaulo:
Seems that this is being fixed here #1115

tonyhallett:
This is still an issue

eenewbsauce:
This is pretty sad guys...

jaredpalmer:
It's out in v2-rc.5 

https://github.com/jaredpalmer/formik/releases/tag/v2.0.1-rc.5

jaredpalmer:
Updated the initial codesandbox to Formik 2 and React 16.8.6 (something with hooks)

prajavk:
Please suggest work around for checkbox with Field render for formik 1.5.7

KrallXZ:
> 
> 
> Please suggest work around for checkbox with Field render for formik 1.5.7

@prajavk , It works for me:
```jsx
<Field type="checkbox" name="manual-only" id="manual-only" onChange={event => setFieldValue('manualOnly', event.target.checked)} checked={values.manualOnly} />
```

kkarkos:
Workaround with react-native NativeBase

```
<CheckBox color="#000"
                      onPress={event =>
                        props.setFieldValue("terms", !props.values.terms)
                      }
                      checked={props.values.terms}
                    />
```

raj-bridgeport:
[...]

Issue #1035, opened 2018-10-22
[docs] Add guide for radio and checkbox input


jaredpalmer:
## 📖 Documentation

I looked through our Algolia DocSearch analytics and it seems like people are searching for radio and checkboxes. Let's create a guide in the docs just for this.

<img width="931" alt="screenshot 2018-10-22 09 57 56" src="https://user-images.githubusercontent.com/4060187/47296751-3ba7a300-d5e1-11e8-9fca-50eb865203b1.png">

amcvitty:
Anecdotally, I was searching for it because they don't bind to values like you'd expect.

#1050 
#1024

I'm going to have a crack at a PR for that

adamduncan:
Hi @jaredpalmer, is there any update on these docs please?

There are a few related PRs (#1064, #1115), as well as some different approaches (#363, #602) being proposed by users of Formik from time to time.

Would be super helpful to have definitive examples of: single checkbox (true/false value), checkbox group (array of values) and radio group (single value).

Thanks

maitandat1507:
Are there any official link to new docs for this? I try to find on Official page https://jaredpalmer.com/formik/ but still can't get it.
I appreciate your help. Many thanks!

Andreyco:
closing in favor of #1315

etoxin:
I would suggest documentation over examples. Just saying.

Fieel:
@Andreyco  y tho

ironbyte:
Why is v2 still not fully documented?

hungvu193:
any updates here?

Issue #1024, opened 2018-10-18
Field's value for checkboxes and radios is not being read properly


hnordt:
### Current Behavior

Field's value for checkboxes and radios is not being read properly.

### Expected behavior

Values for checkboxes and radios should be read from Formik state and not from props.

### Reproducible example

https://codesandbox.io/s/ppy7wyrl0m

### Suggested solution(s)

I think the problem lies here: https://github.com/jaredpalmer/formik/blob/master/src/Field.tsx#L155-L158

I believe the fix would be changing:

```js
      value:
        props.type === 'radio' || props.type === 'checkbox'
          ? props.value // React uses checked={} for these inputs
          : getIn(formik.values, name),
``` 

To:

```js
     [props.type === 'radio' || props.type === 'checkbox' ? "checked" : "value"]:
       getIn(formik.values, name),
``` 

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 1.3.1
| React            | 16.5.2

dem0nbruce:
I also have this issue where checkboxes aren't binding correctly.

gzzo:
What's also interesting is that it doesn't seem to be respecting the `value` passed to `props` either:

https://codesandbox.io/s/n5ozork8vm

I would expect `isEnabled` to be the value of the input when it's checked, `test`, in this case.

zzuber:
@jaredpalmer I agree. Not having the ability to get the value from a checkbox or radio limits formik awesomeness!!!

smithcoin:
I was trying to look through the docs and I can't figure out how to get the value from a radio input. Is this issue closed?

I would love to see this feature implemented.

giorgioongithub:
+1

jankalfus:
@smithcoin I believe this is the minimum working example for radio buttons: https://codesandbox.io/s/88jqoy4qvl

tonyhallett:
This is still an issue

smithcoin:
Why was this issue closed? The bug still exists.

jaredpalmer:
This has been fixed in v2.

zzuber:
@jaredpalmer when is v2 going to be released?

gregoralbrecht:
[...]

PR #1064, opened 2018-10-31
Automagically set <Field> type radio & checkbox `checked` and `value` props against form values

amcvitty:
Previously when creating a Field with `type="????"` to create a simple HTML component we were setting the `value` prop on all input components from the values object, but with this change we set `checked` for checkboxes and radio buttons, so that the state is synced with the values object (including its initial values) so the components are controlled (in React terminology).

Addresses issues #1050 #1024 and partly #1035

### DIFF
--- a/docs/api/field.md
+++ b/docs/api/field.md
@@ -34,7 +34,12 @@ const Example = () => (
   <div>
     <h1>My Form</h1>
     <Formik
-      initialValues={{ email: '', color: 'red', firstName: '' }}
+      initialValues={{
+        email: '',
+        color: 'red',
+        firstName: '',
+        lastName: '',
+        likesCats: false }}
       onSubmit={(values, actions) => {
         setTimeout(() => {
           alert(JSON.stringify(values, null, 2));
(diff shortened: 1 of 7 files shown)


amcvitty:
Note to reviewer: 
There are a bunch of `actual; injected;` in Field.test.tsx because I was seeing a lot of these noUnusedLocals warnings:
```
ERROR in [at-loader] ./test/Field.test.tsx:364:11 
    TS6133: 'actual' is declared but its value is never read.
```

The pattern there seems to be to capture the props in the component render to do assertions on it, but in a number of the tests, we don't have any assertions that make use of the variable. 
```
      let actual;
      const Component: React.SFC<FieldProps> = props =>
        (actual = props) && null;
```

I'm not usually a Typescript guy, so I'm not 100% sure what to do about this. Happy to just revert these changes

jaredpalmer:
Agree

prichodko:
I didn't want to take over your PR, but I already had most of the work done. Sorry for that. 

My proposed solution is here: https://github.com/jaredpalmer/formik/pull/1115

Andreyco:
I consider this to be to specific and should be kept in user land. This would definitely not work when accepting multiple values for one field (multiple checked boxes for one field)... And other cases, for sure.

PR #1089, opened 2018-11-07
Refactor handleSubmit, FormikActions, FormikComputedProps tests

prichodko:
For some reason I couldn't get this test - https://github.com/jaredpalmer/formik/pull/1089/files#diff-a09101aa3d073a58459f3eb56b452b29R545 passing. 

When I change `onSubmit`'s `jest.fn()` to something simple as `() => console.log('called')`, then I can see it being called in the console. Spy function however is failing.

### DIFF
--- a/test/Formik.test.tsx
+++ b/test/Formik.test.tsx
@@ -917,99 +917,131 @@ describe('<Formik> alt', () => {
       });
     });
   });
-});
 
-describe('<Formik>', () => {
   describe('FormikComputedProps', () => {
     it('should compute dirty as soon as any input is touched', () => {
-      const tree = shallow(BasicForm);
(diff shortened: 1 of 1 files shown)


prichodko:
I also think we should enhance `react-testing-library`'s `render` with our custom render to simplify test setup.

```js
const renderFormik = props => {
  let injected: any;
  return {
    get injected() {
      return injected;
    },
    ...render(
      <Formik initialValues={{ name: 'hello' }} onSubmit={noop} {...props}>
        {formikProps => (injected = formikProps) && <Form {...formikProps} />}
      </Formik>
    ),
  };
};

// and use
const { injected, getByTestId } = renderFormik()
```

jaredpalmer:
Nice!

Issue #1092, opened 2018-11-08
Array-helper insert add value to error state


marcel-devdude:
### Current Behavior

When i add an item to a FieldArray using `insert` the value is also inserted into the errors state (see example below). 

### Expected behavior

Only insert the value to the values state, not to the errors state.

### Reproducible example

https://codesandbox.io/s/v87n4p6023

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 1.3.1
| React            | 16.5.2

marcel-devdude:
I have already submitted a PR: #1094

UlyssesInvictus:
This is still an issue as of 2.1.1 and the thread for the linked PR has been stale for a year. Just commenting so that people can find it if they're wondering what's going on.

udaysai1311:
Is the issue is closed or open? If it is open i would like to work on it.

PR #1094, opened 2018-11-08
Fix array insert to not add values to error/touched state

marcel-devdude:
Related to #1092 

The `updateArrayField` function use the passed `fn` to repeat the array operations on each state (values, errors, touched). For the most cases this is fine. 

But for `insert` and `unshift` the value is also written to the `errors` and `touched` state. I solved this by giving the possibility to pass an function to alter the `error` / `touched` state separately.

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -102,8 +102,8 @@ class FieldArrayInner<Values = {}> extends React.Component<
 
   updateArrayField = (
     fn: Function,
-    alterTouched: boolean,
-    alterErrors: boolean
+    alterTouched: boolean | Function,
+    alterErrors: boolean | Function
   ) => {
     const {
       name,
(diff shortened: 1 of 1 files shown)


Magellol:
I've recently came across this bug in #1242.
Anything we can do to move forward with this PR?

Magellol:
@jaredpalmer Anything preventing having this PR merged? If the conflict is the only thing, I don't mind applying @marcel-devdude changes on top of latest master and re-submit a PR.

jaredpalmer:
Let's resolve conflicts. This LGTM

Magellol:
See https://github.com/jaredpalmer/formik/pull/1311

I wasn't sure how to proceed, so I've pulled @marcel-devdude changes into one of my branch and resolved the conflict from there.

PR #1095, opened 2018-11-08
Refactor and unify Field / FastField test suite

prichodko:
It feels like cheating.


test/Field.test.tsx
```diff
@@ -1,545 +1,352 @@
...
-import { shallow, mount } from 'enzyme';
 import { noop } from './testHelpers';
 
-interface TestFormValues {
-  name: string;
-  email: string;
+const initialValues = { name: 'jared', email: 'hello@reason.nyc' };
+type Values = typeof initialValues;
+
+function renderForm(ui: any, props?: Partial<FormikConfig<Values>>) {
```
jaredpalmer: This needs to be typed.

jaredpalmer:
```suggestion
function renderForm(ui: any, props?: Partial<FormikConfig<Values>>) {
function renderForm(ui: React.ReactNode, props?: Partial<FormikConfig<Values>>) {
```

jaredpalmer: I also think we should maybe make both optional and reverse the order

jaredpalmer: We should also share these between Field and Formik in a perfect world

test/Field.test.tsx
```diff
@@ -1,545 +1,352 @@
...
+
+function cases(
+  title: string,
+  tester: (arg: typeof renderField | typeof renderFastField) => void
+) {
+  describe(title, () => {
+    it('<Field />', async () => Promise.resolve(tester(renderField)));
+    it('<FastField />', async () => Promise.resolve(tester(renderFastField)));
+  });
+}
```
jaredpalmer: Cool

jaredpalmer: TIL!

test/Field.test.tsx
prichodko: I wanted to keep it as close as possible to original `render` from `react-testing-library`, e.g. ui as a first argument.

test/Field.test.tsx
prichodko: Yeah, I was thinking if we could use `Promise.resolve` trick in Formik instead of `isPromise` check.

prichodko:
I will definitely look into further optimization of renderers when it all gets refactored / merged. 

Regarding the order, we will probably see what is used more, if passing props or rendering ui. Because if it is `ui`, then it would be cumbersome to always pass `{}` as a first arg.

test/Field.test.tsx
```diff
@@ -1,545 +1,357 @@
...
-      const FormFields = () => <Field name="name" validate={validate} />;
-
-      const tree = mount(<TestForm component={FormFields} />);
-      tree
-        .find(FormFields)
-        .props()
-        .validateField('name');
-      await Promise.resolve()
-        .then()
-        .then(); // XXX: Waiting for validateField's promise, should be fixed :(
```
prichodko: Fixed 😆

PR #1097, opened 2018-11-09
Refactor withFormik test suite

prichodko:
At the beginning I was thinking about taking similar approach as in Field/FastField case, but then I realized that it wouldn't make sense, because Formik and withFormik are not identical components as Field and FastField, because withFormik uses Formik under the hood. 

So this refactor deletes all duplicit tests and rather tests unique functionality and whether are props/config passed down correctly.


jaredpalmer:
Reviewing meow 🐱

Issue #1103, opened 2018-11-12
RFC: Store flattened nested form state internally with dotpath: value


DavidBabel:
## 🚀 Feature request

Speed up Formik by preventing the use of **toPath** / **getIn** / **setIn** where it's not necessary.
It' simple, but not fast to do.

### Current Behavior

Each of this methods : **toPath** (lodash) / **getIn** (internal) / **setIn** (internal) are very consuming. With 50 fields in a simple form, if i type only 3 characters in a `FastField`, i got close to 12000 loop launched (!) only for the "shouldComponentUpdate" part.
![image](https://user-images.githubusercontent.com/5599375/48346480-916ce980-e67b-11e8-8060-ea53ac3f0805.png)

The problem starts with a good idea. Use the **_.toPath** (which is handfull with the result, but only for the result actually). But [toPath](https://github.com/lodash/lodash/blob/master/toPath.js) use two loops [here](https://github.com/lodash/lodash/blob/master/map.js) and [here](https://github.com/lodash/lodash/blob/master/.internal/copyArray.js) and should be used only when needed.

So the problem appear because of this implementation stategy, way too much loop are triggered on rerendering, making the whole app laggy.

### Desired Behavior

Same behaviour for user, but less laggy.

### Suggested Solution

The solution is actually quite simple. instead of managing a multi level object inside the Formik internal state (thanks to the "toPath"), why not only store the full object path as a key ?

It will be very easier to implement the **getIn** / **setIn** without any loop, and save the use of **toPath** only for the submitted result.

Replace library internal state from : 
```json
{
  "person" : {
    "name": "myname",
    "forname": "myforname"
  }
}
```

by : 
```json
{
  "person.name" : "myname",
  "person.forname": "myforname"
}
```
And **only** us a **toPath** : 
- onSubmit()
- setTouched()
- set...Anything() // the provided methods of formik

This will save a LOT of unneeded calculation loops, for `Field` and `FastField` regardless.

### Who does this impact? Who is this for?

It can be transparent for users.
Formik can become scalable.
It can save this library and the world :)

### Describe alternatives you've considered

I made a very quick change to the `FastFields` to only render if they are the last touched field only : 
here is the PR : https://github.com/jaredpalmer/formik/pull/1104
implemented here for test purposes : https://www.npmjs.com/package/@davidbabel/formik

Just implemented a "lastTouched" item which contain the name of the last touched item. And it save a looooot of loops during rendering : 
[...]

PR #1104, opened 2018-11-12
update only if i'm the touched item

DavidBabel:
see https://github.com/jaredpalmer/formik/issues/1103

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -110,20 +110,22 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
   ) {
     if (this.props.shouldUpdate) {
       return this.props.shouldUpdate(props, this.props);
-    } else if (
-      getIn(this.props.formik.values, this.props.name) !==
-        getIn(props.formik.values, this.props.name) ||
-      getIn(this.props.formik.errors, this.props.name) !==
-        getIn(props.formik.errors, this.props.name) ||
-      getIn(this.props.formik.touched, this.props.name) !==
-        getIn(props.formik.touched, this.props.name) ||
-      Object.keys(this.props).length !== Object.keys(props).length ||
-      this.props.formik.isSubmitting !== props.formik.isSubmitting
-    ) {
-      return true;
-    } else {
-      return false;
     }
+    if (props.formik.lastTouched === this.props.name) {
+      if (
+        this.props.formik.isSubmitting !== props.formik.isSubmitting ||
+        Object.keys(this.props).length !== Object.keys(props).length ||
+        getIn(this.props.formik.values, this.props.name) !==
+          getIn(props.formik.values, this.props.name) ||
+        getIn(this.props.formik.errors, this.props.name) !==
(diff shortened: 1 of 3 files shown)


src/FastField.tsx
```diff
@@ -110,20 +110,22 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
...
-      getIn(this.props.formik.touched, this.props.name) !==
-        getIn(props.formik.touched, this.props.name) ||
-      Object.keys(this.props).length !== Object.keys(props).length ||
-      this.props.formik.isSubmitting !== props.formik.isSubmitting
-    ) {
-      return true;
-    } else {
-      return false;
     }
+    if (props.formik.lastTouched === this.props.name) {
```
brodybits: Why not put this all in a single return statement, something like this?

```js
    return (props.formik.lastTouched === this.props.name) && (
        this.props.formik.isSubmitting !== props.formik.isSubmitting ||
        Object.keys(this.props).length !== Object.keys(props).length ||
        getIn(this.props.formik.values, this.props.name) !==
          getIn(props.formik.values, this.props.name) ||
        getIn(this.props.formik.errors, this.props.name) !==
          getIn(props.formik.errors, this.props.name) ||
        getIn(this.props.formik.touched, this.props.name) !==
          getIn(props.formik.touched, this.props.name)
      );
```

brodybits: Or more drastically rewrite the function body like this:

```js
    return this.props.shouldUpdate) ?
      this.props.shouldUpdate(props, this.props) :
      (props.formik.lastTouched === this.props.name) && (
        this.props.formik.isSubmitting !== props.formik.isSubmitting ||
        Object.keys(this.props).length !== Object.keys(props).length ||
        getIn(this.props.formik.values, this.props.name) !==
          getIn(props.formik.values, this.props.name) ||
        getIn(this.props.formik.errors, this.props.name) !==
          getIn(props.formik.errors, this.props.name) ||
        getIn(this.props.formik.touched, this.props.name) !==
          getIn(props.formik.touched, this.props.name)
      );
```

DavidBabel: i hardly prefer your first suggestion for reading purpose.

note: is it a fast review because it relies on fast fields ? :p

DavidBabel: changed it

tlenclos:
We have a big form in our app and this PR is improving performance 👍 
The user experience is still very bad for our user but this is a move in the right direction.

jaredpalmer:
While this would work for Formik v1's registration, i think it has some problems with registration in v2. In v1 we naively assume there is 1:1 relation ship with fields to components. This is why we didn't implement ErrorMessage with field (because that would break). However, in v2, this won't be the case. This trick could still work though, the value would just include some kind of random number or uuid.

jaredpalmer:
However, we may not even really need registration (or rather it may become hard to understand) with fiber/suspense and with upcoming `useField`. Not sure yet.

jaredpalmer:
Closing in favor of #2749

PR #1109, opened 2018-11-15
Fix wrong TS types for handleBlur (#1108)

### LINKED ISSUE #1108: Curried handleBlur type is broken in TS 3.1.x
### Current Behavior

While using React-Native, setting `TextInput`'s `onBlur` prop to `props.handleBlur('fieldName')` renders a typescript error.

image

### Expected behavior

No typescript error.

### Suggested solution(s)

The problem's source is following snippet:

https://github.com/jaredpalmer/formik/blob/82f45bf0b2af054bef8fe8e0e315b1918f08f50f/src/types.tsx#L140-L144

Because of declarations' order, using `props.onBlur('fieldName')` fits the first definition and the method's return type is computed as `any` rather than proper event handler. This behaviour is actually described in [Typescript's _Do's and Don'ts_](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#ordering)

Reversing declarations' order seems like a possible solution. However, I'm not sure whether or not current order was introduced on purpose. Also, it should first be made sure that the reversed order doesn't break web forms.

Another solution would be to actually use proper type rather than `any` for `e` argument, as with `handleSubmit` and `handleChange` in the same interface.

While asking for additional data, please keep in mind, that I've been a happy Formik user for less than two hours now, so there is still a lot I don't know 😉 

### Workaround

For now, a working workaround is to directly point typescript to the generic `handleBlur<T>` method by using `onBlur={props.handleBlur<string>('prefix')}`

### Your environment

| Formik           | 1.3.2 |
| React            | 16.5.0 |
| React-Native | 0.57 / expoSDK 31.0.1 |
| TypeScript       | 3.1.6 |
| Browser          | n/a |
| npm/Yarn         | yarn 1.12.1 |
| Operating System | Win10 x64 |

burtek:
Fixes #1108

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -137,7 +137,7 @@ export interface FormikHandlers {
   /** Reset form event handler  */
   handleReset: () => void;
   /** Classic React blur handler, keyed by input name */
-  handleBlur(e: any): void;
+  handleBlur(e: React.FocusEvent<any>): void;
   /** Preact-like linkState. Will return a handleBlur function. */
   handleBlur<T = string | any>(
     fieldOrEvent: T


jaredpalmer:
This works with React Native?

burtek:
Works for me at least :p

jaredpalmer:
@ferrannp is this kosher?

PR #1115, opened 2018-11-16
Fix input type="checkbox" & type="radio"

prichodko:
I believe there is a bigger problem than what is covered here https://github.com/jaredpalmer/formik/pull/1064 and I had this PR half-done, so I finished that. ¯\_(ツ)_/¯

What (I think) we want to achieve:
- keep API of render/children/component (not string) same
- handle `checked` attribute for string components, where we have 3 scenarios

First of all, we always have to take value form `Formik` state, no matter the type

For `type="radio`:
```js
<Field type="radio" name="console" value="ps4" />
<Field type="radio" name="console" value="xbox" />
```
the `checked` property is a result of `props.value === formik.values.console`.

For `type="checkbox"`: 
 ```js
<Field type="checkbox" name="some-name"  />
```
the `checked` property is a boolean handled during `onChange` by Formik [here](https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L323), therefore `checked` is just `formik.values["some-name"]`.

...aaand everything else is `value`.

Update for `FastField` and tests are done as well.

Fixes https://github.com/jaredpalmer/formik/issues/1024
Fixes https://github.com/jaredpalmer/formik/issues/1050

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -166,11 +166,9 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
       validationSchema: _validationSchema,
       ...restOfFormik
     } = formik;
+
     const field = {
-      value:
-        props.type === 'radio' || props.type === 'checkbox'
-          ? props.value // React uses checked={} for these inputs
-          : getIn(formik.values, name),
+      value: getIn(formik.values, name),
       name,
       onChange: formik.handleChange,
       onBlur: formik.handleBlur,
@@ -186,11 +184,23 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
     }
 
     if (typeof component === 'string') {
-      const { innerRef, ...rest } = props;
+      const { innerRef, ...restProps } = props;
+      const { value, ...restField } = field;
+      const valueProp: { checked?: boolean; value?: typeof value } = {};
+
+      if (props.type === 'radio') {
+        valueProp.checked = value === props.value;
+      } else if (props.type === 'checkbox') {
+        valueProp.checked = value;
+      } else {
+        valueProp.value = value;
+      }
+
       return React.createElement(component as any, {
         ref: innerRef,
-        ...field,
-        ...rest,
+        ...restField,
+        ...restProps,
+        ...valueProp,
         children,
       });
     }
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -184,23 +184,18 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
     }
 
     if (typeof component === 'string') {
-      const { innerRef, ...restProps } = props;
+      const { innerRef, type, ...restProps } = props;
       const { value, ...restField } = field;
-      const valueProp: { checked?: boolean; value?: typeof value } = {};
-
-      if (props.type === 'radio') {
-        valueProp.checked = value === props.value;
-      } else if (props.type === 'checkbox') {
-        valueProp.checked = value;
-      } else {
-        valueProp.value = value;
-      }
 
       return React.createElement(component as any, {
+        type,
         ref: innerRef,
         ...restField,
         ...restProps,
-        ...valueProp,
+        ...{
+          [type === 'radio' || type === 'checkbox' ? 'checked' : 'value']:
+            type === 'radio' ? props.value === value : value,
+        },
         children,
       });
     }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by prichodko** [comment]
The last problem I see here, is that something like:

```js
const Input = styled.input``
...
<Field type="checkbox" component={Input} />
```

is not possible now. We could potentially modify `props` if `type` is present. However I am not sure if it isn't too much magic.

src/Field.tsx
```diff
@@ -171,11 +169,23 @@ class FieldInner<Values = {}, Props = {}> extends React.Component<
     }
 
     if (typeof component === 'string') {
-      const { innerRef, ...rest } = props;
+      const { innerRef, ...restProps } = props;
+      const { value, ...restField } = field;
+      const valueProp: { checked?: boolean; value?: typeof value } = {};
+
+      if (props.type === 'radio') {
```
jaredpalmer: Probably ways to simplify with an es6 computed property.

{ [type === radio ? ‘checked’ : ...]: getIn... }

prichodko: Full implementation then would look something like:
```js
const useChecked = type === 'radio' || type === 'checkbox'
{ [useChecked ? 'checked' : 'value']: type === 'radio' ? props.value === value : value
```

do you think it's better?

prichodko:
Updated.

jaredpalmer:
@prichodko `<Field component>` prop was meant to mimic redux-form. However, this has not aged very well. Good news is that when hooks land we will encourage people to write their own `<Fieldset />`.

jaredpalmer:
@prichodko not sure modifying props based on `type` would work, it would break Field. The better solution would have been to spread `field` in component from day 1.

src/FastField.tsx
```diff
@@ -186,11 +184,18 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
...
+      const { value, ...restField } = field;
+
       return React.createElement(component as any, {
+        type,
         ref: innerRef,
-        ...field,
-        ...rest,
+        ...restField,
+        ...restProps,
+        ...{
```
jaredpalmer: you don't need to spread here.

zzuber:
Great solution @prichodko and @jaredpalmer. What's the status on this?

jaredpalmer:
Let’s fix snapshots and then I want to test this out against existing checkboxes in our apps and make sure that it is backwards compat.

Ciberusps:
Merge?

Andreyco:
I consider this to be to specific and should be kept in user land. This would definitely not work when accepting multiple values for one field (multiple checked boxes for one field)... And other cases, for sure.

src/FastField.tsx
```diff
@@ -186,11 +184,18 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
...
+          [type === 'radio' || type === 'checkbox' ? 'checked' : 'value']:
```
KITSEPAMIgorSergiienko:
```js
['radio', 'checkbox'].includes('type')

File: src/Field.tsx [hunk 4]
```diff
@@ -171,11 +169,18 @@ class FieldInner<Values = {}, Props = {}> extends React.Component<
...
       return React.createElement(component as any, {
+        type,
         ref: innerRef,
-        ...field,
-        ...rest,
+        ...restField,
+        ...restProps,
+        ...{
+          [type === 'radio' || type === 'checkbox' ? 'checked' : 'value']:
+            type === 'radio' ? props.value === value : value,
```
johnrom: if `props.value` is an array (multiple checkboxes), we should check if `props.value.includes(value)`. Also since this exact comparison is used twice, I think it should be refactored like `{ ...mapFieldValue(type, value, props.value) }`: 

```
// might exist somewhere already, further reading [https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html](on TypeScript website)
type RemoveArray<ValueType> = ValueType extends (infer ValueTypeWithoutArray)[] ? ValueTypeWithoutArray : ValueType;

export function mapFieldValue<ValueType = any>(type: string, value: RemoveArray<ValueType>, currentValue: ValueType) {
    const key = ['radio', 'checkbox'].includes(type) ? 'checked' : 'value';

    return {
        [key]: type === 'radio'
            ? isArray(props.value) 
                ? props.value.includes(value) 
                : props.value === value 
            : value,
    }
}
```

jaredpalmer:
See #1555

jaredpalmer:
Can we close this?

Andreyco:
@jaredpalmer I would close this. Truth is, this could solve part of problem but not whole problem. There would be other issues after merge.

PR #1143, opened 2018-11-30
Spread extra Field props to render prop and children-as-function

### LINKED ISSUE #942: Field ownProps are not passed into render function

christianchown:
Closes #942 (which is marked as _wontfix_), and is a near-duplicate of #957 from @adsingh (which was closed without merge). 

So given that, I better make a good pitch for you not to hit the insta-close button! 

There's been discussion in #957, and here are some reasons why we'd like this PR merged.

1. **Consistency between the 3 ways to render using `Field`**
Of the three ways to render things with `Field`, additional props ([as per the docs](https://jaredpalmer.com/formik/docs/api/field)) 
get passed through to 1 of them - `component`, which gets `...bag` and `...props` below

https://github.com/jaredpalmer/formik/blob/5cba633bbb3d06193fccf8b0e54fa91b2c0b6457/src/Field.tsx#L183-L187

but the other 2 don't. _render_ prop and _children-as-function_ just get `bag`, no `props`:

https://github.com/jaredpalmer/formik/blob/5cba633bbb3d06193fccf8b0e54fa91b2c0b6457/src/Field.tsx#L165-L171

2. **A composite component can't decorate `Field`**.
[The main reason given behind closing #957](https://github.com/jaredpalmer/formik/pull/957#issuecomment-427029872) was that you could pass addditional props directly to a render function via a closure. This does not hold if you want to use `Field` compositionally. I have a case - https://github.com/jaredpalmer/formik/pull/957#issuecomment-430533580 - where I have a `DecoratedField` component that renders a `Field` where the render prop is injected by a consumer... however I also want to pass some of `DecoratedField`'s state to this render prop function. It cannot be done, because all it gets is `bag`

3. **Adding and removing extra props to a `FastField` causes it to rerender, but those props can't then be used _in_ the render**
Here's the line in `shouldComponentUpdate` that forces the rerender, as per the docs. It seems anachronistic not to be able to use the values that cause the rerender, unless they're `Field` ones.
https://github.com/jaredpalmer/formik/blob/4cabec1881f47ee603ae6ab97b8eb92652c8bcc7/src/FastField.tsx#L120

Anyway, hope you'll consider the above

ps loving your work

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -178,11 +178,14 @@ class FastFieldInner<Values = {}, Props = {}> extends React.Component<
     const bag = { field, form: restOfFormik };
 
     if (render) {
-      return (render as any)(bag);
+      return (render as any)({ ...bag, ...props });
     }
 
     if (isFunction(children)) {
-      return (children as (props: FastFieldProps<any>) => React.ReactNode)(bag);
+      return (children as (props: FastFieldProps<any>) => React.ReactNode)({
+        ...bag,
+        ...props,
+      });
     }
 
     if (typeof component === 'string') {
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -163,11 +163,14 @@ class FieldInner<Values = {}, Props = {}> extends React.Component<
     const bag = { field, form: restOfFormik };
 
     if (render) {
-      return (render as any)(bag);
+      return (render as any)({ ...bag, ...props });
     }
 
     if (isFunction(children)) {
-      return (children as (props: FieldProps<any>) => React.ReactNode)(bag);
+      return (children as (props: FieldProps<any>) => React.ReactNode)({
+        ...bag,
+        ...props,
+      });
     }
 
     if (typeof component === 'string') {


jaredpalmer:
Formik’s render props work exactly like React Router’s work.  In the future, we will deprecate render entirely.

christianchown:
Just dug through React Router's `Route`, and that doesn't pass props through, but does have a mechanism for proving decoration by [spreading its `staticContext`](https://github.com/ReactTraining/react-router/blob/9d278b4b110f38479725fe5a1104990c2f3e76c2/packages/react-router/modules/Route.js#L17-L31)

v2 (and hooks?) will likely make all this moot, of course, but until then...

jaredpalmer:
Again, this was a design decision. You can write your own version of Field with this behavior if you’d like with connect().

christianchown:
Hi - at the moment, I am using `component`, as these get `props` as well as `bag`. Should the other two ways to render - render props and children-as-function - not get the same `props` along with `bag`?

jaredpalmer:
This introduces redundancy. All props can be accessed through a closure in the case of render and children as they are already in scope. There is no need to send them.

PR #1150, opened 2018-12-03
Allow callbacks to be passed to formik functions that set state

### LINKED ISSUE #529: setFieldValue should return a Promise or receive a callback

raq929:
Affects:
setErrors, setValues, setTouched
setFieldErrors, setFieldValues, setFieldTouched

This allows much greater flexibility when using these functions and resolves #529

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,28 +1,28 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 176109,
-    "minified": 47667,
-    "gzipped": 13847
+    "bundled": 176101,
+    "minified": 47655,
+    "gzipped": 13834
   },
   "./dist/formik.cjs.production.js": {
-    "bundled": 40832,
-    "minified": 20783,
-    "gzipped": 5229
+    "bundled": 40820,
+    "minified": 20771,
+    "gzipped": 5214
   },
   "./dist/formik.cjs.development.js": {
-    "bundled": 41720,
-    "minified": 21666,
-    "gzipped": 5575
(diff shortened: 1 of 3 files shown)


jaredpalmer:
I want to merge something like this, but it would not be possible to implement in v2 because Hooks does not yet support this kind of behavior (see my issue here https://github.com/facebook/react/issues/14221). 

🤷‍♂️

raq929:
Right now, I need dispatch an action in Redux after setTouched runs. The alternative seems to be setting a timeout, or watching for the change to occur in getDerivedStateFromProps, both of which I would really like to avoid. Are there any other solutions for this? I'm happy to make the PR if there's a proposed alternative.
Out of curiosity, when is v2.0 coming out?

hrgui:
@raq929 I wonder if https://github.com/jaredpalmer/formik-effect would help? 

```tsx
<Effect onChange={(currentFormikState, nextFormikState) => {
               const {touched} = nextFormikState;
               this.props.dispatch({touched});
            }} 
          />
```

PR #1169, opened 2018-12-08
Only use isValidating during submit

jaredpalmer:
This removes the toggling of `isValidating` everywhere (change/blur) except:

- During submission
- When someone calls `validateForm` directly

Most users do not care about `isValidating` outside of the submission process, plus this constant flickering is the root of the recent performance regressions #671 #1026.

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -112,7 +112,7 @@ Returns `true` if the there are no `errors`, or the result of
 
 #### `isValidating: boolean`
 
-Returns `true` if Formik is running any validation function, `false` otherwise. To learn more about what happens with `isValidating` during the submission process, see [Form Submission](guides/form-submission.md).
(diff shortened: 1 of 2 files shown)


royipressburger:
hi @jaredpalmer

How do you suggest to handle this case:

I have an input that has indicator on it (that displays V or X if the input is valid or not).

I could use `isValidating` to determine if to show that indicator incase I have async validations:
```
const inputValidation = yup.string().test('is-valid', 'Input is not valid', (v) => {
 return new Promise((res) => setTimeout(() => res(false), 1000));
})

{
   !form.isValidating && !form[errors][name] && (<SuccessIndication />)
   !form.isValidating && form[errors][name] && (<ErrorIndication />)
}
```

Now since I don't have `form.isValidating` I have a split second where the input is touched, has value (that is invalid) and the error object is still not up to date since the validation is async.
As you can see the validation takes at least 1 second at that time the error object is empty.

any ideas how to handle this case?

Issue #1119, opened 2018-11-20
Formik automatically converts input type number value to float


TLadd:
### Current Behavior

When a form field is an `<input type="number" />`, Formik automatically parses the value to a float. It does so [here](https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L322):
```
val = /number|range/.test(type)
          ? ((parsed = parseFloat(value)), isNaN(parsed) ? '' : parsed)
          : /checkbox/.test(type) ? checked : value;
```
This behavior was unexpected for me. It ended up causing an issue in validation where I was doing a truthiness check on the value and was expecting "0" to be the value, but it ended up being the number 0. This also opens you up for type errors even when using Typescript. I have the form value typed like `{ value: string }`, but it would really need to be `{ value: string | number }`, since the type of the value will change depending on if the input is empty of not. It seems like a pretty easy mistake to make.

### Expected behavior

By default, I would expect Formik to not interfere with the default behavior of `<input type="number />` and keep the value as a string. The value automatically parsing the value doesn't seem to be worth the confusion it causes by differing from the default React behavior. The answer to [this issue](https://github.com/jaredpalmer/formik/issues/961#issuecomment-427242274), where the suggestion to achieve such behavior was to build your own `NumberInput` component, would be what I expect.

### Reproducible example

https://codesandbox.io/s/40kq34q69. Typing numbers into the `Formik Number Input` results in the type of the value being a number, whereas the `Normal Number Input` remains a string.

### Suggested solution(s)

No longer call parseFloat for `<input type="number" />`. I imagine if we did end up making this change, it would make sense to be consistent and also not do it for `<input type="range" />` as well. It looks like this behavior has been a part of Formik a long ways going back, so it would definitely be a breaking change. 

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | v1.3.2
| React            | v16.6.3
| TypeScript       | v3.1.6
| Browser          | Chrome
| npm/Yarn         | Yarn 1.12.3
| Operating System | Mac OS X

vojtechportes:
Another issue with calling parseFloat is that if you try to type eg 0.01, once you get to 0.0, it will round the value to 0 (because parseFloat(0.0) is 0)

jaredpalmer:
Can you submit a PR?

TLadd:
Sure, I'll get a PR together.

ufuji1984:
Same here.
[...]

PR #1171, opened 2018-12-09
Remove special number parsing logic for input of type number and range

TLadd:
Resolves https://github.com/jaredpalmer/formik/issues/1119.

Removes the special number parsing behavior for number and range inputs, and just maintains the value as-is.

It looks like this was the only use of the utility isNaN method, so may want to take that out at well.
```
/** @private is the given object a NaN? */
export const isNaN = (obj: any): boolean => obj !== obj;
```

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -499,7 +499,6 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       // provides just the value of the input.
       let field = maybePath;
       let val = eventOrTextValue;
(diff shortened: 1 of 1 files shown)


TLadd:
@jaredpalmer @eonwhite @crosscompile @slightlytyler Is this still on the table? I would think a major release would be the logical time to make a change like this, so imagine the window of opportunity to do it is closing.

dontepsu:
When is this being merged?

PR #1178, opened 2018-12-12
Fix handling array operations when errors is an array-like object

### LINKED ISSUE #1158: TypeError: `array.splice is not a function` with non-array errors
### Current Behavior

* Create a form with a `FieldArray`.
* `setErrors` to be of the form `{"0": ['invalid']}` (note that we're not using an array here).
* When you remove the first item in the field array (using `arrayHelpers.remove`, it crashes with `TypeError: a.splice is not a function`

### Expected behavior

No crash, should just work. Note that while I was testing, it seems it was working fine with 1.2.0.

### Reproducible example

https://codesandbox.io/s/mjoknk1q2y

```jsx
// Helper styles for demo
import "./helper.css";
import { MoreResources, DisplayFormikState } from "./helper";

import React from "react";
import { render } from "react-dom";
import { Formik, Form, FieldArray, Field, getIn } from "formik";
import * as Yup from "yup";

// To reproduce:
// 1. Click submit
// 2. Try to remove the first item
// 3. Observe the crash `array.slice is not a function`

const App = () => (
  <div>
    <h1>Friend List</h1>
    <Formik
      initialValues={{ friends: [{ name: "jared" }] }}
        // HERE, note that we set the error using an object - not an array.
        actions.setErrors({ friends: { "0": { name: ["invalid"] } } });
      }}
      render={({ values }) => (
        <Form>
          <FieldArray
            name="friends"
            render={arrayHelpers => (
              <div>
                {values.friends && values.friends.length > 0 ? (
                  values.friends.map((friend, index) => (
                    <div key={index}>
                      <Field name={`friends.${index}.name`}>
                        {({ field, form }) => {
                          const error = getIn(form.errors, field.name);
                          console.log(error);
                          console.log(field);
                          console.log(form.touched[field.name]);
                          return (
                            <div>
                              <input
                                type="text"
                                {...field}
                                placeholder="First Name"
                              />
                              {error && <div className="error">{error}</div>}
                            </div>
                          );
                        }}
                      </Field>
                      <button
                        type="button"
                        onClick={() => arrayHelpers.remove(index)} // remove a friend from the list
                      >
                        -
                      </button>
                      <button
                        type="button"
                        onClick={() => arrayHelpers.insert(index, "")} // insert an empty string at a position
                      >
                        +
                      </button>
                    </div>
                  ))
                ) : (
                  <button type="button" onClick={() => arrayHelpers.push("")}>
                    {/* show this when user has removed all friends from the list */}
                    Add a friend
                  </button>
                )}
                <div>
                </div>
              </div>
            )}
          />
        </Form>
      )}
    />
  </div>
);

render(<App />, document.getElementById("root"));

```

### Suggested solution(s)

In my case, the back-end returns nested object errors using an object with numerical keys. This makes sense, because it allows it to report an error for the 125th object by just returning `{friends: {124: ["invalid"]}}`.

This looks like a regression - `formik` should allow that too.

### Your environment

| Formik           | 1.3.2
| React            | 16.6.3
| TypeScript       | N/A
| Browser          | N/A
| npm/Yarn         | N/A
| Operating System | N/A

charlax:
Attempt to fix #1158 

Note: this is not finished yet - the logic has only been applied to `remove`, it needs to be applied to the other array operations as well (`pop`, etc.).

There are numerous ways to fix the bug, I chose to go with transforming the array-like object (`{1: "invalid"}`) into a sparse array. I don't think that's a great solution because it changes the structure of the developer-controlled `errors` object... The only advantage is that it is less code changes, because then the logic can stay the same once the object has been transformed into an array.

The other solution would be to have a different handling of array and object for all array helpers. That is a much more extensive change, I wanted to have your opinion on the direction before moving further ahead.

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -85,6 +85,23 @@ export const replace = (array: any[], index: number, value: any) => {
   copy[index] = value;
   return copy;
 };
+
+const copyArray = (array?: any[]) => {
+  if (!array) {
+    return [];
+  }
+  if (isFunction(array.slice)) {
+    return [...array];
+  }
+
+  // It's an array-like object, e.g. {2: "hello"}, let's create a sparse array from it.
+  const copy = new Array();
+  Object.keys(array).forEach(k => {
+    copy[parseInt(k)] = array[parseInt(k)];
+  });
+  return copy;
+};
+
 class FieldArrayInner<Values = {}> extends React.Component<
   FieldArrayConfig & { formik: FormikContext<Values> },
   {}
@@ -200,7 +217,7 @@ class FieldArrayInner<Values = {}> extends React.Component<
     this.updateArrayField(
       // so this gets call 3 times
       (array?: any[]) => {
-        const copy = array ? [...array] : [];
+        const copy = copyArray(array);
         if (!result) {
           result = copy[index];
         }
--- a/test/FieldArray.test.tsx
+++ b/test/FieldArray.test.tsx
@@ -14,6 +14,14 @@ const TestForm: React.SFC<any> = p => (
   />
 );
 
+const TestObjectsForm: React.SFC<any> = p => (
+  <Formik
+    onSubmit={noop}
+    initialValues={{ friends: [{ name: 'jared' }] }}
+    {...p}
+  />
+);
+
 describe('<FieldArray />', () => {
   const node = document.createElement('div');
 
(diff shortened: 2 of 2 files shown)


charlax:
Note that this also adds tests for array of objects (as opposed to just array of strings). I actually don't need them so I might remove them.

charlax:

src/FieldArray.tsx
```diff
@@ -85,6 +85,23 @@ export const replace = (array: any[], index: number, value: any) => {
...
+  if (!array) {
+    return [];
+  }
+  if (isFunction(array.slice)) {
+    return [...array];
+  }
+
+  // It's an array-like object, e.g. {2: "hello"}, let's create a sparse array from it.
+  const copy = new Array();
+  Object.keys(array).forEach(k => {
```
adnasa: Although this is fine, you can use [`Array.From(arrayLike, mapFn)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)

austinbutler:
Seems this PR is obsoleted by https://github.com/jaredpalmer/formik/pull/1830.

PR #1192, opened 2018-12-18
Replace warning

jaredpalmer:
Warning is not treeshakable :shrug:

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,35 +1,35 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 141428,
-    "minified": 39046,
-    "gzipped": 11848
+    "bundled": 141508,
+    "minified": 39072,
+    "gzipped": 11859
   },
   "./dist/formik.umd.development.js": {
-    "bundled": 171423,
-    "minified": 49396,
-    "gzipped": 14922
+    "bundled": 171632,
+    "minified": 49502,
+    "gzipped": 14960
   },
   "./dist/formik.cjs.production.js": {
-    "bundled": 42378,
-    "minified": 21478,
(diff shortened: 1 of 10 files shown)


jaredpalmer:
Errrrrr. Actually, I'm not sure this will work. the warning module provides the call stack, which is kinda nice. 

Also tiny-warning, uses `console.warn` while warning actually uses `console.error`

TrySound:
Why error instead of warning? Error always frighten me when I see it in console.

TrySound:
I can add call stack
https://github.com/alexreardon/tiny-warning/issues/2

TrySound:
https://github.com/alexreardon/tiny-warning/pull/6

TrySound:
Could you point me how throwing error helps with debugging? I see stack trace in both cases
https://codesandbox.io/s/l34rvm5r8z

jaredpalmer:
Warning uses console.error. there is a full discussion about it in React repo somewhere. Not sure why we would change this behavior now.

TrySound:
I found it. I understand why try catch there. But `console.error` reason seems like a legacy. I see stack trace with warning too. I just think that semantically error is too aggressive.

https://github.com/facebook/react/issues/4216

jaredpalmer:
FWIW React still uses console.error https://github.com/facebook/react/blob/master/packages/shared/warningWithoutStack.js#L40

TrySound:
Should we? The decision was made because of features lack in old browsers. Now it's probably a time to follow semantic. We may even rise an issue in react so they could migrate in v17.

jaredpalmer:
Fair enough. We just need to update test harness.

jaredpalmer:
This should work now

PR #1201, opened 2018-12-19
fix(formik): validation upon form submission in case of thrown errors

### LINKED ISSUE #1198: Form is submitted if there are uncaught errors in validate function
### Current Behavior

At the moment, if within the `validate` function an error is thrown, Formik does not handle it, causing the form to be submitted as it thinks there are no validation errors.

### Expected behavior

If an error is thrown, Formik should handle it and do not allow to submit the form.

### Reproducible example

https://codesandbox.io/s/5klqjpl00k

### Suggested solution(s)

~~I think the problem is in the `runValidateHandler` function~~

https://github.com/jaredpalmer/formik/blob/13954fdd10070f727a0259dbbf98824cb980fa92/src/Formik.tsx#L223-L241

~~Given that the `this.props.validate` is wrapped within the `new Promise`, if an error is thrown it will be swallowed.~~

~~The fix is to properly handle those errors.~~

Update: after digging a bit more into the code, the problem is in the `submitForm` function

https://github.com/jaredpalmer/formik/blob/13954fdd10070f727a0259dbbf98824cb980fa92/src/Formik.tsx#L446

If `combinedErrors` is an instance of an `Error`, the check passes

image

### Your environment

See codesandbox dependencies.

emmenko:
Fixes #1198

As described in the issue, the main problem was in the `submitForm` function

https://github.com/jaredpalmer/formik/blob/13954fdd10070f727a0259dbbf98824cb980fa92/src/Formik.tsx#L446

If `combinedErrors` is an instance of an `Error`, the check passes, **causing the form to be submitted**.

image

I also extended the types to return either an `Error | FormikErrors<Values>`.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -187,7 +187,7 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
 
   runFieldLevelValidations(
     values: FormikValues
-  ): Promise<FormikErrors<Values>> {
+  ): Promise<Error | FormikErrors<Values>> {
     const fieldKeysWithValidation: string[] = Object.keys(this.fields).filter(
       f =>
         this.fields &&
@@ -215,23 +215,29 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
           }
           return prev;
         },
-        {} as FormikErrors<Values>
+        {} as Error | FormikErrors<Values>
       )
     );
   }
 
-  runValidateHandler(values: FormikValues): Promise<FormikErrors<Values>> {
-    return new Promise(resolve => {
+  runValidateHandler(
+    values: FormikValues
+  ): Promise<Error | FormikErrors<Values>> {
+    return new Promise((resolve, reject) => {
       const maybePromisedErrors = (this.props.validate as any)(values);
       if (maybePromisedErrors === undefined) {
         resolve({});
       } else if (isPromise(maybePromisedErrors)) {
         (maybePromisedErrors as Promise<any>).then(
-          () => {
-            resolve({});
+          maybeErrors => {
+            resolve(maybeErrors);
           },
           errors => {
-            resolve(errors);
+            if (errors instanceof Error) {
+              reject(errors);
+            } else {
+              resolve(errors);
+            }
           }
         );
       } else {
@@ -265,7 +271,7 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
    */
   runValidations = (
     values: FormikValues = this.state.values
-  ): Promise<FormikErrors<Values>> => {
+  ): Promise<Error | FormikErrors<Values>> => {
     if (this.validator) {
       this.validator();
     }
@@ -293,7 +299,7 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
         });
         return errors;
       })
-      .catch(x => x);
+      .catch(error => error);
   };
 
   handleChange = (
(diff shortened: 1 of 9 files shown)


src/Formik.tsx
```diff
@@ -441,16 +443,39 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
...
+    return this.runValidations(this.state.values).then(
+      combinedErrors => {
+        this.setState({ isValidating: false });
+        // In case an error was thrown and passed to the resolved Promise,
+        // `combinedErrors` can be an instance of an Error. We need to check
+        // that and abort the submit.
+        // If we don't do that, calling `Object.keys(new Error())` yields an
+        // empty array, which causes the validation to pass and the form
+        // to be submitted.
+        const isInstanceOfError = combinedErrors instanceof Error;
```
emmenko: This is the actual fix

src/Formik.tsx
```diff
@@ -441,16 +443,39 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
...
+        const isValid =
+          !isInstanceOfError && Object.keys(combinedErrors).length === 0;
+        if (isValid) {
+          this.executeSubmit();
+        } else if (this.didMount) {
+          // ^^^ Make sure Formik is still mounted before calling setState
+          this.setState({ isSubmitting: false });
+        }
+      },
+      error => {
```
emmenko: I'm not sure if this is really necessary but I felt like we could gracefully handle this and show a warning.

If you think it's not necessary, I can remove it.

test/Formik.test.tsx
```diff
@@ -442,13 +442,24 @@ describe('<Formik>', () => {
...
       });
 
-      it('should not submit the form if invalid', () => {
+      it('should not submit the form if invalid', async () => {
         const onSubmit = jest.fn();
         const validate = jest.fn(() => ({ name: 'Error!' }));
         const { getByTestId } = renderFormik({ onSubmit, validate });
 
         fireEvent.submit(getByTestId('form'));
-        expect(onSubmit).not.toBeCalled();
```
emmenko: This assertion was not really working, as we need to `wait` to let the promise resolve

test/Formik.test.tsx
```diff
@@ -442,13 +442,24 @@ describe('<Formik>', () => {
...
+        await wait(() => expect(onSubmit).not.toBeCalled());
+      });
+
+      it('should not submit the form if validate function throws an error', async () => {
+        const onSubmit = jest.fn();
+        const validate = jest.fn(() => {
+          throw new Error('Oops');
```
emmenko: Here we check that if an error is thrown, the form is not submitted

test/Formik.test.tsx
```diff
@@ -470,13 +481,24 @@ describe('<Formik>', () => {
         await wait(() => expect(onSubmit).toBeCalled());
       });
 
-      it('should not submit the form if invalid', () => {
+      it('should not submit the form if invalid', async () => {
         const onSubmit = jest.fn();
-        const validate = jest.fn(() => Promise.resolve({ name: 'Error!' }));
+        const validate = jest.fn(() => Promise.reject({ name: 'Error!' }));
```
emmenko: This was also "wrong", we want to check that the form is not submitted when `validate` rejects the promise

src/Formik.tsx
```diff
@@ -215,23 +215,29 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
...
+    return new Promise((resolve, reject) => {
       const maybePromisedErrors = (this.props.validate as any)(values);
       if (maybePromisedErrors === undefined) {
         resolve({});
       } else if (isPromise(maybePromisedErrors)) {
         (maybePromisedErrors as Promise<any>).then(
-          () => {
-            resolve({});
+          maybeErrors => {
+            resolve(maybeErrors);
```
emmenko: This case assumes that I write the `validate` function like this:

```js
const validate = values => new Promise(resolve => {
  const errors = {};
  if (!values.name) errors.name = "Missing required value";
  resolve(errors);
});
```

src/Formik.tsx
```diff
@@ -215,23 +215,29 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
...
           },
           errors => {
-            resolve(errors);
+            if (errors instanceof Error) {
```
emmenko: In this case, an error was thrown within the promise in the `validate` function.


```js
const validate = values => new Promise(resolve => {
  const errors = {};
  if (!values.name) errors.name = "Missing required value";
  throw new Error('Oops');
  resolve(errors);
});
```

We need to **reject** this promise, otherwise the `deepmerge.all` in the `runValidations` will swallow the error.

https://github.com/jaredpalmer/formik/blob/13954fdd10070f727a0259dbbf98824cb980fa92/src/Formik.tsx#L273-L284

image

test/Formik.test.tsx
```diff
@@ -470,13 +481,27 @@ describe('<Formik>', () => {
...
+        await wait(() => expect(onSubmit).not.toBeCalled());
+      });
+
+      it('should not submit the form if validate function rejects with an error', async () => {
         const onSubmit = jest.fn();
-        const validate = jest.fn(() => Promise.resolve({ name: 'Error!' }));
+        const validate = jest.fn(
+          () =>
+            new Promise(() => {
+              throw new Error('Oops');
```
emmenko: This test covers the case described here: https://github.com/jaredpalmer/formik/pull/1201/files#r243037849

emmenko:
I might need some help with compiling the types in the `connect.tsx`.

emmenko:
> I might need some help with compiling the types in the `connect.tsx`.

Ok, I managed somehow to work around the compiler to cast `FormikContext<Values>` (https://github.com/jaredpalmer/formik/pull/1201/commits/c7fc561d04624c23aa98e2c14cc0fe2d32050419). Please check if that's ok.

probablyup:
@jaredpalmer Can you please take a look at this? I've had to put a try/catch around my validation logic because of this bug.

emmenko:
@jaredpalmer have you had a chance to look at this? How do you wish to proceed here?

jaredpalmer:
Thanks for this. There 2 warts in validation right now. First, validate swallows errors. We need to fix that and the above case will not happen because a real runtime error will be thrown. 

Next, I want to implement async onSubmit so folks don't need to write `setSubmitting(false)`. Some people have suggested that we should automagically gather up any submit errors into a `submitError` or `apiError` prop. The way to do that is to make a rule that anything resolved from async `onSubmit` will be transformed into `apiError`. Rejections will thus bubble up normally. This then puts the error catching work in userland and away from formik.

jaredpalmer:
See #1575. I think it solves the problem.

PR #1211, opened 2018-12-26
Force validation if called during submission

### LINKED ISSUE #1209: [1.4.X] submitForm() isn't hitting handler because combinedErrors = {isCanceled: true}
### Current Behavior

I'm using [Component Component](https://ui.reach.tech/component-component) to automatically submit my form when a select field changes. The sandbox below works if you use Formik 1.3.2 but doesn't hit the handleSubmit with 1.4.1. Debugging into Formik from my real app shows that in submitForm() it does this:

```
      return _this.runValidations(_this.state.values).then(function (combinedErrors) {
        _this.setState({
          isValidating: false
        });
```

In the code above `combinedErrors = {isCanceled: true}` but I don't know why and I'm assuming this is why my handleSubmit() isn't being called. 

### Expected behavior

My handleSubmit() to be called. 

### Reproducible example

[Edit Formik Effect vs Component^2](https://codesandbox.io/s/p30y4qw2mq)

The console will show when submitForm() and handleSubmit() get called. Typing in the name field will cause submitForm() to get called. If you change the Formik version to 1.3.2 in the sandbox it calls handleSubmit()

### Suggested solution(s)

I'm not sure this is a bug or something changed from 1.3.X to 1.4.X. If you have any suggestions on why isCanceled is being set please let me know. 

### Your environment

| Formik           | 1.4.1
| React            | 16.7.0
| Browser          | Chrome
| Yarn         |  1.12.3
| Operating System | Windows 10

jaredpalmer:
Closes #1209

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,28 +1,28 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 141423,
-    "minified": 39046,
-    "gzipped": 11848
+    "bundled": 141343,
+    "minified": 39020,
+    "gzipped": 11840
   },
   "./dist/formik.umd.development.js": {
-    "bundled": 171603,
-    "minified": 49483,
-    "gzipped": 14950
+    "bundled": 171523,
+    "minified": 49457,
+    "gzipped": 14940
   },
   "./dist/formik.cjs.production.js": {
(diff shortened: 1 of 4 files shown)


jaredpalmer:
Not needed

PR #1216, opened 2018-12-30
Improve handling of non event values in handleChange and handleBlur

MrLoh:
This PR improves `handleChange` and `handleBlur` for cases where they are not passed an event, for usage with React Native (Web) and other UI libraries, where the library currently struggles. 

- `handleBlur` can now be called with undefined instead of an event.
- `handleChange` can now no only be called with a string value, but also any other value

The gist of this PR is checking for `isEvent` instead of checking for `!isString`. This makes the handling much more straight forward. I also took the liberty to refactor some of the code to be less dense (no confusing ternary assignment returns) and improve the types for `handleBlur`.


MrLoh:
Circle CI only fails because credentials are missing. The tests all pass.

krizka:
@jaredpalmer Please approve this!

src/utils.ts
```diff
@@ -126,6 +126,10 @@ export const isEmptyChildren = (children: any): boolean =>
 export const isPromise = (value: any): value is PromiseLike<any> =>
   isObject(value) && isFunction(value.then);
 
+/** @private is the given object/value a type of synthetic event? */
+export const isEvent = (value: any): value is React.SyntheticEvent<any> =>
```
jaredpalmer: Rename to isInputEvent?

src/Formik.tsx
```diff
@@ -298,37 +299,21 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
...
-          id,
-          value,
-          checked,
-          outerHTML,
-        } = (eventOrTextValue as React.ChangeEvent<any>).target;
+      let value: any;
+      if (isEvent(eventOrValue)) {
+        const event = eventOrValue as React.ChangeEvent<any>;
+        // If we can, persist the event, https://reactjs.org/docs/events.html#event-pooling
+        if (event.persist) event.persist();
```
jaredpalmer: Add curly + new line

src/Formik.tsx
```diff
@@ -298,37 +299,21 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
...
-  ): void | ((eventOrTextValue: string | React.ChangeEvent<any>) => void) => {
-    // @todo someone make this less disgusting.
-    //
-    // executeChange is the core of handleChange, we'll use it cache change
-    // handlers like Preact's linkState.
+  ): void | ((eventOrValue: any | React.ChangeEvent<any>) => void) => {
+    // this function actually handles the change
     const executeChange = (
-      eventOrTextValue: string | React.ChangeEvent<any>,
+      eventOrValue: any | React.ChangeEvent<any>,
```
jaredpalmer: Is it any? Or a string?

MrLoh: A use case I had in my project, were custom input elements that for-example return a number in onChange or even an array (for a tag input). That's why I change the check, not relying on detecting wether the value is a string, but instead wether the value is an event. This way the default case is that onChange has any kind of value and the special case that it is an event like for default html input elements.

src/Formik.tsx
MrLoh: sorry, fixed it

src/utils.ts
MrLoh: good idea, done

MrLoh:
@jaredpalmer thanks for reviewing this quickly! This PR addresses more than only issues with RN. Checking the value for `isInputEvent` instead of checking for `isString` allows usage of Formik with any ui library that doesn't return string values. Basically if the field returns an event, Formik will handle converting it to the right type. Otherwise the input component can give back any type (Number, Date, Array of Strings, ...) and Formik will take the value. A check of the execution environment wouldn't be able to help with that.

src/Formik.tsx
jaredpalmer: Maybe we use unkown then.

MrLoh: Sure. I didn’t know that type. Should I change that?

jaredpalmer: yes please

MrLoh:
Sorry for the delay, I was traveling, changed the type from any to unknown where relevant

bsr203:
hi I too like to have this functionality as I am not sure how to use a custom component otherwise. Mine is a Select component from an external library which provides the value at onChange callback. using string value is not a great option as the field value may be an id, but select component may display other attributes. so need to pass an object as value. Please let me know if it is possible now. thanks

MrLoh:
@bsr203 your usecase sounds like mine. You can install my fork branch from github and try it out.

bsr203:
thanks @MrLoh Meanwhile I found `form.setFieldValue` API allow any value to set to the field.

src/Formik.tsx
```diff
@@ -337,42 +324,52 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
             handlerName: 'handleChange',
           });
         }
-        val = /number|range/.test(type)
-          ? ((parsed = parseFloat(value)), isNaN(parsed) ? '' : parsed)
-          : /checkbox/.test(type) ? checked : value;
+        value = event.target.value;
+        if (/number|range/.test(type)) {
+          const parsed = parseFloat(event.target.value);
+          value = isNaN(parsed) ? '' : parsed;
```
Andreyco: How about intermediate values, e.g.
```
var value = "12."; // user is still typing the value
var num = parseFloat(value); // 12
```
With this code, 12 would be set, losing "." and impossible to type it. Or am I wrong?

This kind of type conversion should be done on blur event.

It is tricky field

MrLoh: @Andreyco This is completely unrelated to this PR, as this code was not changed. It might be a separate issue that needs fixing

MrLoh: I only refactored the unnecessarily terse and confusing javascript syntax here.

krizka: still need it. please update

MrLoh: @Andreyco what’s going on, can you please resolve this issue or provide some feedback.

MrLoh:
@jaredpalmer can we please get this moving forward, it is starting to get a bit frustrating, I’ve made all the effort I can and after 2 months this has still not been merged 😞. I understand that time is limited for this open source project but it’s frustrating for contributors.

MrLoh:
@Andreyco @jaredpalmer any updates on this?

src/utils.ts
```diff
@@ -126,6 +126,10 @@ export const isEmptyChildren = (children: any): boolean =>
 export const isPromise = (value: any): value is PromiseLike<any> =>
   isObject(value) && isFunction(value.then);
 
+/** @private is the given object/value a type of synthetic event? */
+export const isInputEvent = (value: any): value is React.SyntheticEvent<any> =>
+  value && isObject(value) && isObject(value.target);
```
jaredpalmer: Does this work in React Native?

jaredpalmer: I think we should name it `isDomInputEvent`

jaredpalmer: or `isSyntheticDomEvent`

jaredpalmer: Oh i see in React Native e.target will will be a number.

src/Formik.tsx
```diff
@@ -298,37 +299,23 @@ export class Formik<Values = object, ExtraProps = {}> extends React.Component<
 
   handleChange = (
     eventOrPath: string | React.ChangeEvent<any>
-  ): void | ((eventOrTextValue: string | React.ChangeEvent<any>) => void) => {
-    // @todo someone make this less disgusting.
-    //
-    // executeChange is the core of handleChange, we'll use it cache change
-    // handlers like Preact's linkState.
+  ): void | ((eventOrValue: unkown | React.ChangeEvent<any>) => void) => {
```
jaredpalmer: typo: unknown

jaredpalmer:
I think this will work meow

jhoch:
@MrLoh @jaredpalmer how was this intended to work with `name` or `id`?  [`field` is not set](https://github.com/jaredpalmer/formik/pull/1216/files#diff-18d1f112a814aca08509a74bded25fbfR322) in the case of a non-event first-argument for `handleChange`.  Wondering if this is intentional or an oversight.

I've just started looking at Formik and happen to want to use it with [react-number-format](https://github.com/s-yadav/react-number-format) which exposes `onValueChange`, which passes a value directly, instead of an event.  I could see `handleChange` taking two arguments, (1) `value` and (2) `name`.  Perhaps there's an appropriate "advanced" API in Formik already that I haven't found yet though!

PR #1225, opened 2019-01-06
issue number 1158 solution

TechnologyGeek12:


jaredpalmer:
This data structure is not supported for errors though.

Issue #905, opened 2018-09-13
Warnings from Draft.js (Immutable.js) with Formik 1.3.0 (iterable.length has been deprecated)


jnv22:
## Current Behavior

When using formik 1.3.0 and draft.js, I'm getting hundreds of warnings thrown by immutable, when trying to type in the RichEditorExample component. The warnings say: `iterable.length has been deprecated, use iterable.size or iterable.count(). This warning will become a silent error in a future version`.  This is only happening when I'm calling `setFieldValue`, from draft.js, using a lodash-like path syntax field, to modify a value that's nested more than one-level deep.

## Steps to Reproduce

See example attached (cloned from: https://github.com/jaredpalmer/formik/issues/93).  All that I did was upgrade to formik 1.3.0 and make the initial structure nested. 
The draft.js component is now using the updated path: `editorState.edit` instead of `editorState`.

## Expected behavior

No Errors thrown

## Suggested solution(s)

I was able to track this bug down to `lodash.cloneDeep`. Since Draft.js uses Immutable.js, we shouldn't be using lodash.  Short of bringing in Immutable to deep clone this structure, are there any lightweight alternatives?

## Additional context

**CodeSandbox Link**:
https://codesandbox.io/embed/v4vo4yny5
---

- Formik Version: 1.3.0
- React Version: 15.3.0
- TypeScript Version: n/a
- Browser and Version: Chrome 69
- OS: MacOS High Sierra
- Node Version: v10.4.1
- Package Manager and Version: npm v6.1.0

jnv22:
Bump:  Just wanted to see if anyone has thought about this.  I'd love to bounce ideas off of people!

jamiedavenport:
I'm also having this problem with Formik 1.4.2 and Draft 0.10.5. My use case includes having a `FieldArray` of Draft editors.

The performance of my form makes it unusable so it'd be nice to get this issue reopened.

Does anyone know if this issue also exists for Slate?

davidbarker:
I managed to stop the warnings by following the steps here: https://github.com/facebook/draft-js/issues/950#issuecomment-381911759

I hope it works for you too!

PR #1235, opened 2019-01-11
Change setIn to use clone instead of cloneDeep.

danifus:
Improve performance of ``setIn`` by shallow copying unchanged siblings
along the path to be updated.

Could enable significant performance improvements when implementing
``shouldComponentUpdate`` methods for nested structures as it allows direct
comparison of object / arrays to see if there has been changes to any of
their child values (because unchanged objects retain the same js
reference).


danifus:
Hi,

This is useful when ``fieldArray`` items are expensive to render. It allows the component which renders the ``fieldArray`` item to compare the new and old item by reference, simplifying the implementation of ``shouldComponentUpdate`` for that component.

From looking at the code (and passing tests) I think it is ok but I'd like to get your input on whether you think this change may have unintended side effects.

jaredpalmer:
I need to look up the issue. However, I think it was changed to deepClone for a reason.

jaredpalmer:
We should add a comment either way as to why or why not since I can’t remember off the top of my head

danifus:
That is a good idea.

How about a comment like:
```
/**
 * Deeply set a value from in object via it's path.
 * @see https://github.com/developit/linkstate
 * If the value at `path` has changed, return a shallow copy of obj with
 * `value` set at `path`. If `value` has not changed, return the original
 * `obj`.
 *
 * Existing objects / arrays along `path` are also shallow copied. Sibling
 * objects along path retain the same internal js reference. Because new
 * objects / arrays are only created along `path`, we can test if anything
 * changed in a nested structure by comparing the object's reference in
 * the old and new object, similar to how russian doll cache invalidation
 * works.
 */
export function setIn(obj: any, path: string, value: any): any {
...
}
```

jamiedavenport:
Does this have implications for issues such as #905?

danifus:
Depends how deeply the incompatible Immutable.js object is nested within the Draft.js object that gets put into Formik's `values`. `clone` and `cloneDeep` both use `baseClone` underneath which [still accesses `.length`](https://github.com/lodash/lodash/blob/master/.internal/arrayLikeKeys.js#L23), so the only way this change would help is if the Draft.js object in `values` is not itself the incompatible object and can be shallow copied (avoiding the incompatible Immutable.js object completely).

I've never used Draft.js so I can't tell you how likely that is.

jamiedavenport:
I've run my project using `danifus:master` and I don't have the performance issues when using custom Draft.js inputs.

kenta88:
Hi guys, there are any update about this? 
There are any intention to merge this one? 

I have the same issue with draft.js and this will help. In case this will not be merged, i'll have a possible workaround for that. (basically avoiding the use of FieldArray)

Thanks

danifus:
Hi, I've been using this in my private projects for a while and haven't noticed any issues so far.

Are there any parts of the code base that you suspect could be affected by this change that you would like me to write some tests for?

danifus:
Was this the previous issue you were thinking of: https://github.com/jaredpalmer/formik/issues/364? That issue was setting a nested value would mutate the parent instead of creating a new object.

`clone` avoids that bug making a shallow copy of the objects along the update path so no object is mutated in place.

The test added for that issue is still passing: https://github.com/jaredpalmer/formik/blob/master/test/utils.test.tsx#L180

dzg:
Is there documentation for `setIn`?

How would I set a value for a single property of a single object of an array of objects?
I.e.,

jaredpalmer:
@danifus this is great. Thanks! Pushing out a new release now

PR #1236, opened 2019-01-12
Removing is/are info in the readme

luisrudge:
No need to overcomplicate 😬


jaredpalmer:
Can you make this change in the other docs versions too?

luisrudge:
done ❤️

PR #1258, opened 2019-01-22
Do not forward shouldUpdate to component

### LINKED ISSUE #1132: FastField: React does not recognize the `shouldUpdate` prop on a DOM element
Formik 1.3.2

charlax:
Fixes #1132

### DIFF
--- a/test/Field.test.tsx
+++ b/test/Field.test.tsx
@@ -6,6 +6,7 @@ import {
   FastField,
   FieldProps,
   FieldConfig,
+  FastFieldConfig,
   FormikProps,
   FormikConfig,
   FastFieldProps,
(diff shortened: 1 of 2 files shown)


test/Field.test.tsx
```diff
@@ -355,3 +356,14 @@ describe('Field / FastField', () => {
     expect(getProps().field.value).toBe('Binding');
   });
 });
+
+describe('<FastField />', () => {
```
charlax: @jaredpalmer Created its own section because this test is only for `FastField`. Also, I did not create `FastField.test.tsx` for a single test. Let me know what you think.

charlax:

PR #1260, opened 2019-01-22
Replace uglify with terser plugin

jaredpalmer:
- Replace `rollup-plugin-uglify` with `rollup-plugin-terser`

cc @TrySound

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,7 +3,7 @@ import replace from 'rollup-plugin-replace';
 import resolve from 'rollup-plugin-node-resolve';
 import sourceMaps from 'rollup-plugin-sourcemaps';
 import babel from 'rollup-plugin-babel';
-import { uglify } from 'rollup-plugin-uglify';
+import { terser } from 'rollup-plugin-terser';
 import { sizeSnapshot } from 'rollup-plugin-size-snapshot';
 import pkg from './package.json';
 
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -60,14 +60,17 @@ const buildUmd = ({ env }) => ({
...
+        sourcemap: true,
         output: { comments: false },
         compress: {
           keep_infinity: true,
           pure_getters: true,
         },
         warnings: true,
+        ecma: 5,
         toplevel: false,
+        mangle: false,
```
TrySound: Why not mangle? This is expected from production umd I think.

jaredpalmer: What options should we use for mangle?

TrySound: No options. Just remove `mangle: false`. Or is there a reason for it?

jaredpalmer: No idea. Just wanted to check.

rollup.config.js
```diff
@@ -60,13 +60,15 @@ const buildUmd = ({ env }) => ({
...
+      terser({
+        sourcemap: true,
         output: { comments: false },
         compress: {
           keep_infinity: true,
           pure_getters: true,
         },
         warnings: true,
+        ecma: 5,
         toplevel: false,
```
jaredpalmer: @TrySound should we run terser on CJS production build as well? I see that it can accept a `format` option.

TrySound: Yeah, probably

jaredpalmer:
Moved to #1261

Issue #812, opened 2018-08-06
RFC: Controlled Formik (a.k.a top-level onChange prop)


jaredpalmer:
## Current Behavior

Formik is an uncontrolled component. You can set default values through initialValues/mapPropsToValues and then it's a black box until submit happens. You can use FomikEffect to trigger side effects underneath.

## Desired Behavior

Give up. Let people get silly. Allow them to store their own parts of Formik state. 

```jsx
class ControlFormikValues extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      name: '',
    };
  }

  render() {
    return (
      <Formik
        values={this.state}
        onChange={({ values }, state) => this.setState(values)}
      >
        <Field name="name" />
      </Formik>
    );
  }
}
```

## Suggested Solution

```jsx
import React from 'react';
import { FormikContext } from './FormikContext';

function cbToCb(cb) {
  return typeof cb === 'function' ? cb : noop;
}

function noop() {}

export class Formik extends React.Component {
  handleChange = e => {
    if (e.persist) {
      e.persist();
    }
    this.internalSetState(state => ({
      values: { ...state.values, [e.target.name]: e.target.value },
    }));
  };

  handleBlur = e => {
    if (e.persist) {
      e.persist();
    }
    this.internalSetState(state => ({
      touched: { ...state.touched, [e.target.name]: true },
    }));
  };

  state = this.getState({
    values: this.props.initialValues,
    errors: this.props.initialErrors,
    touched: this.props.initialTouched,
    isSubmitting: false,
    isValidating: false,
    submitCount: 0,
    handleBlur: this.handleBlur,
    handleChange: this.handleChange,
  });

  getState(stateToMerge = this.state) {
    return Object.keys(stateToMerge).reduce((state, key) => {
      state[key] = this.isControlledProp(key)
        ? this.props[key]
        : stateToMerge[key];
      return state;
    }, {});
  }

  isControlledProp(key) {
    return this.props[key] !== undefined;
  }

  internalSetState = (stateToSet, cb) => {
    let onStateChangeArg = {};

    const isStateToSetFunction = typeof stateToSet === 'function';
    return this.setState(
      prevState => {
        prevState = this.getState(prevState);
        let newStateToSet = isStateToSetFunction
          ? stateToSet(prevState)
          : stateToSet;
        let nextState = {};
        Object.keys(newStateToSet).forEach(key => {
          if (prevState[key] !== newStateToSet[key]) {
[...]

PR #1263, opened 2019-01-22
Add initialStatus

### LINKED ISSUE #471: Initialization of status
## Bug, Feature, or Question?
Bug

## Current Behavior
When including a reference to `this.status.property` in a Formik inner component, where `property` is determined after `handleSubmit` is called, initially referencing `this.status.property` results in error: Cannot find property 'property' of undefined

## Desired Behavior
Reference to `this.status.property` should return `undefined` on component initialization.

## Suggested Solutions
After a few hours of digging, and looking at [this](https://codesandbox.io/s/ppnkqv78wq) CodeSandbox, the current solution is to append `this.status &&` before `this.status.property` so the conjunction short-circuits to false.

Given this is a basic JS fact of how objects evaluate, it may seem like the solution is obvious, but I found the documentation regarding how `status` is handled a little sparse. 

Maybe there could be a `initializeStatus` property for the Formik component which allows you to specify `status` on page initialization?

- Formik Version: 0.11.11
- React Version: 16.0
- TypeScript Version: N/A
- CodeSandbox Link: https://codesandbox.io/s/ppnkqv78wq
- OS: Windows 10
- Node Version: 9.5
- Package Manager and Version: npm 5.6.0

charlax:
Fixes #471 (proof of concept, lacking tests)

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -194,7 +194,7 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
   /**
    * Initial status
    */
-  initialStatus: any;
+  initialStatus?: any;
 
   /**
    * Reset handler
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -192,6 +192,11 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
   initialValues: Values;
 
   /**
+   * Initial status
+   */
+  initialStatus: any;
+
+  /**
    * Reset handler
    */
   onReset?: (values: Values, formikActions: FormikActions<Values>) => void;
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -59,6 +59,7 @@ export class Formik<Values = FormikValues> extends React.Component<
       isSubmitting: false,
       isValidating: false,
       submitCount: 0,
+      status: props.initialStatus,
     };
     this.didMount = false;
     this.fields = {};
@@ -531,7 +532,7 @@ export class Formik<Values = FormikValues> extends React.Component<
       errors: {},
       touched: {},
       error: undefined,
-      status: undefined,
+      status: this.props.initialStatus,
       values,
       submitCount: 0,
     });


charlax:

jaredpalmer:
I think it initialStatus makes sense. However, I think we need to see some real examples though of exactly what this solves for and then decide if it is still necessary against alternatives.

jaredpalmer:
I reread 471 and I’m still unconvinced that status preferred in these situations vs. react state. 

Cc @eonwhite @slightlytyler have you run into the need for this? Recall that status replaced the singular error and setError which were initially meant for a form level error message. It initialized to undefined because we don’t use null at The Palmer Group. It then became a kind of store whatever you want escape hatch. As hooks arrive, I think that status might be a bit of an anti-pattern. Perhaps the only lingering value being that status gets reset with the rest of the form in resetForm(), which I guess is legit, but idk.

charlax:
You're right that this is kind of using `status` for something else than what was originally intended. I could achieve the same thing with building my own context instead of putting this in `status`.

That being said, as soon as you provide `status`, I think you need to allow setting `initialStatus`, just like Downshift provides `initialX` for all of its `X` state. And you provide `initialValues` for `values` so...

jaredpalmer:
I agree. This would also solve for the monstrosity of isInitialInvalid and isValid. However, then we should not stop there and we ought to bring back up the controlled Formik RFC #812 and #837

eonwhite:
I barely use status honestly -- but I agree that as long as it exists, it makes sense to have an option to populate its initial value, just for consistency of API.

src/types.tsx
```diff
@@ -191,6 +191,11 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
    */
   initialValues: Values;
 
+  /**
+   * Initial status
+   */
+  initialStatus: any;
```
jaredpalmer: Make this optional.
```suggestion
  initialStatus?: any;
```

jaredpalmer:
I'm going to finish this all up this afternoon

jaredpalmer:
See #1321

charlax:
Thanks @jaredpalmer !

PR #1267, opened 2019-01-24
Add shouldValidate argument to setTouched and setValues methods

martinhorsky:
To be consistent with `setFieldTouched`. Also I found use-case where it is needed:

When you validate part of the form using custom validate logic, you need to set errors and touched fields without validating whole form.

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -161,9 +161,11 @@ Set `touched` imperatively.
 
 You can also explicitly prevent/skip validation by passing a second argument as `false`.
 
-#### `setValues: (fields: { [field: string]: any }) => void`
+#### `setValues: (fields: { [field: string]: any }, shouldValidate?: boolean) => void`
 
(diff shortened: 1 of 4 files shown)


jaredpalmer:
I want to merge this, but this will be impossible in v2 at the moment

martinhorsky:
So there is no blocker to merge it to `v1`. Or is there?

I only thinking if I should add same logic to `setValues`, but I have not found use-case for that.

jaredpalmer:
For later reference, this is blocked in v2 by https://reactjs.org/docs/hooks-intro.html

jaredpalmer:
Yeah for consistency I would.

I am going to wait to merge this PR. This has nothing to do with the code or you. I am just waiting to see the final hooks API in 16.8 before I make a decision on the current hooks branch.

martinhorsky:
I added `shouldValidate` argument  also to `setValues` method as we agreed.

PR #1274, opened 2019-01-25
update CombinedValidations.js example

DerejeKitaw:
not needed "validate={isRequired('This field is required')}" for zip and email. Formik will validate based on validationSchema={Schema} and validate={validate} by default


jaredpalmer:
The example is meant to show how to use all methods of validation at once

DerejeKitaw:
The example is not using all methods. All of the field are validated by "isRequired" method. The update i suggested shows we can use all methods of validation at once.
"username" => is validated by "isRequired" method
"email" => is validated by "Yup"
"zip" is validated by "validate" method.

In current example all of the fields are validated by "isRequired".
Regards

PR #1282, opened 2019-01-28
We need to use cross platform cp command to support windows developers

anton-nagornyi:


anton-nagornyi:
What is wrong cp-cli?

package.json
```diff
@@ -41,6 +41,7 @@
     "size": "size-limit"
   },
   "dependencies": {
+    "cp-cli": "^1.1.2",
```
jaredpalmer: move to dev dependency

Issue #1195, opened 2018-12-18
Post render callback


brokarth:
## ❓Question
Hello, I got a question, I need to know if there is a post-render callback on formik, I got a button that is outside of the form, so to enable or disable the button, I send a event from the form saying if the form is valid.

I tried using validate function, but it is executed before the forms knows if is valid or not
so i put it on the render function, but react warning me that i execute a state change inside render function (which is right since I'm changing the disable prop of the submit button with a event sent from the form render

I tried putting the function on the `componentDidUpdate`, but it is not called when something change on the form

Why i cannot use the validate function? because each field knows how to validate it self, I need this because the form is built dynamically

jaredpalmer:
Can you create a codesandbox demonstrating what you are trying to do? Or share some code?

brokarth:
Sure, ill create the repo and tell you when is done

garyng2000:
throw away the rulebook of 'pure' , something like below(may need setTimeout but you get the idea)
the key is do use this.setState(), use instance variable. if you must use setState, probably have to it via setTimeout to get around React's protection. Treat setTimeout as external event not trigger by click but timer(no different than manually click a button to disable something).

stealErrorInfo=(errors)=>{this.buttonDisable = errors.isempty}
<Formik
render={({errors})=>
(
<Form>

{this.stealErrorInfo(errors)}
</Form>
)} />
<button disable={this.buttonDisabled}/>

brokarth:
Thanks for the answer but i dont think is good solution.

Here is a repo with the warning for you to see:
https://gitlab.com/brokarth/formik-post-validation-callback

I n the README is the instructions for replicate the problem, if you got any question, tell me :)

garyng2000:
that is exactly what I meant 'throw away the 'pure' rulebook' :-)
you can only get the errors list of Formik during rendering so unless your button is under its context, you cannot see it. what I suggest is to 'steal' it(which is effectively your post render callback)
whether it is a good solution or not is up to you to decide. just make sure that if you must setState rather than using simple instance variable, it needs to be stopped or else it becomes
render -> delay change state(which trigger render)->render ....


the setTimeout would trick React thinking you are doing it as 'pure' even though you are not.

brokarth:
I understand what you say, but since we have already a workaround, I don't think that an `setTimeOut` is a good solution. I would like a definitive solution.

So, it appear that formik doesn't have a post render callback to know the form state at that point... so we can may build it.

So i think in two options:

[...]

PR #1284, opened 2019-01-28
Added new callback for formik's componentDidUpdate function

brokarth:
Added new callback for formik's componentDidUpdate lifecycle function.

More info in this question:
https://github.com/jaredpalmer/formik/issues/1195

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -116,6 +116,10 @@ export class Formik<Values = FormikValues> extends React.Component<
       // @todo refactor to use getDerivedStateFromProps?
       this.resetForm(this.props.initialValues);
     }
+
+    if (this.props.formikDidUpdate) {
+      this.props.formikDidUpdate();
+    }
   }
 
   setErrors = (errors: FormikErrors<Values>) => {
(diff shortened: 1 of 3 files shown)


jaredpalmer:
formik-effect does exactly this. When hooks land users can do this ease with just useEffect(). However, I strongly believe this is a footgun as it would encourage people to duplicate/store formik state improperly. A better solution would be to bring back the controlled Formik RFC as described in #812 and prototyped in #837.

brokarth:
Thanks for your answer. We dont need to know the formik state, only when it is updated, but `formik-effect` looks like it does the work.

I will close this PR and test `formik-effect`, if it doesnt works i will re-open this PR.

Thank you again 😄

brokarth:
We use `formik-effect`, but as it execute `componentWillReceiveProps` inside Formik, it trigger the problem anyway.

@jaredpalmer  as you can see in our PR we dont want manage the formik state, formik do it incredibly. We want the ability to fire a post validate callback.

brokarth:
By the way, `formik-effect` is not working with formik as the last comment in https://github.com/jaredpalmer/formik/issues/812#issuecomment-459780274 says

PR #1287, opened 2019-01-29
Update FieldProps to mimic formik handlers

lukaskral:
This change fixes the type signature for the onChange handler, which can be called with a string as a 1st argument.

Example code:

```
import {Field, FieldProps} from 'formik';

import {TextAreaField} from 'my-own-ui';

interface IProps {
    id: string;
    label: string;
}

function TextArea({id, label}: IProps) {
    return (
        <Field
            name={id}
            render={({field}: FieldProps) => {
                return (
                      <TextAreaField
                          label={component.label || ''}
                          value={field.value}
                          onChange={(val: string) => field.onChange(val)}
                      />
                );
            }}
        />
    );
}
```
Up until now this wasn't possible in Typescript because field.onChange only accepts ```React.ChangeEvent<any>```

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -5,6 +5,7 @@ import {
   FormikProps,
   GenericFieldHTMLAttributes,
   FormikContext,
+  FormikHandlers,
 } from './types';
 import warning from 'tiny-warning';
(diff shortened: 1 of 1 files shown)


jaredpalmer:
LGTM.

PR #1291, opened 2019-01-30
Feature/improve validation

mohamedmagdy17593:
Added a second argument (context) to "**validate**" method to help us run validation based on "context" object. This will help us in cases like async validation that takes time and run on every onChange. After the change, We can now run the validation onChange of a target field only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ You can play with Formik in your web browser with these live online playgrounds.
 * [Sync Validation](https://codesandbox.io/s/q8yRqQMp)
 * [Building your own input primitives](https://codesandbox.io/s/qJR4ykJk)
 * Working with 3rd-party inputs:
-  - [react-select-v1](https://codesandbox.io/s/jRzE53pqR)
-  - [react-select-v2](https://codesandbox.io/s/73jj9zom96)
-  - [Draft.js](https://codesandbox.io/s/QW1rqjBLl)
+  * [react-select-v1](https://codesandbox.io/s/jRzE53pqR)
+  * [react-select-v2](https://codesandbox.io/s/73jj9zom96)
+  * [Draft.js](https://codesandbox.io/s/QW1rqjBLl)
 * [Accessing React lifecycle functions](https://codesandbox.io/s/pgD4DLypy)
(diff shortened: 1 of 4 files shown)


jaredpalmer:
Can you give some code samples of situations where this enables new functionality?

mohamedmagdy17593:
Ok that is the use case that encourage me to think about this is using async validation
[codesandbox](https://codesandbox.io/s/p7wpv66rqj)
async validation run on every change
try to write a valid email then write a password and remove it, it will take time to show "Require" text on password because async validation run on every input change of a whole form

using context argument let me know the context that causes validate function to run and decide which validation to run 
so it could be something like this
``` javascript
async function validate(values, { field, on }) {
  let errors = {};
  if (!values.email) {
    errors.email = "Required";
  } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i.test(values.email)) {
    errors.email = "Invalid email address";
  } else {
    if (on === "submit" || field === "email") {
      errors.email = await isValidEmail(values.email);
    }
  }

  if (!values.password) {
    errors.password = "Required";
  }
  console.log(errors);
  throw errors;
}
```

jaredpalmer:
IMHO this is `isValidEmail` call belongs in `onSubmit` and not in validate. It's presumably hitting your remote API.

mohamedmagdy17593:
and i need also to run `isValidEmail` on `onChange` and `onBlur` but when the changed `field` is "email"

i know that `field` will be the value of the field that cases `validate` to run

jaredpalmer:
I don't think this form is a good candidate for Formik then. FWIW I almost never use Formik on any email/password form.

jaredpalmer:
Just us React here. It will be far easier than bending Formik for this use case.

mohamedmagdy17593:
I am with you and this is just a simple example. Imagine a bigger form with a lot of async validation and other use cases that may make the developer needs to know what causes the `validate` function to run.

Virgomax:
This is a nice feature. I also need field-level async validation.

mohamedmagdy17593:
Yes, i love formik and i wish to be able to do something like this. It will make form validation to be more controlled

Issue #1107, opened 2018-11-13
Formik's onSubmit expects (values: FormValues, actions: FormikActions<FormikValues>)


dfee:
### Current Behavior
Documentation says that the second argument to `onSubmit` is `formikBag`.

### Expected behavior
Actual implementation has the second argument to 'onSubmit` as `formikActions`.

### Suggested solution(s)
Update the documentation (or if it's the code that's erraneous, update the code to pass `formikBag` instead of `formikActions`).

### Additional context
Here seems to be a discussion of this in the past: https://github.com/jaredpalmer/formik/issues/234
Here is the actual code snippet that returns formikActions:
https://github.com/jaredpalmer/formik/blob/6a1f9825534942a40bf1c4bb3ba6dce3a2ffff6a/src/Formik.tsx#L435-L437

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           |  1.3.1
| React            | 16.5.2
| TypeScript       | 3.1.3
| Browser          | Firefox 63.0.1
| npm/Yarn         | 6.4.1
| Operating System | MacOS 10.14

jaredpalmer:
This is fixed in v2.

beausmith:
In case anyone else is looking for the correct typings when creating a onSubmit handler method in a react class:

```js
interface FormValues {
  myString: string,
  myNumber: number
}

...

import {
  FormikActions,
} from 'formik'

...

public onSubmitHandler = (
  values: FormValues,
  formikActions: FormikActions<FormValues>
) => {
  console.log({ values, formikActions })
}
```

EfstathiadisDimitris:
Quick question guys. Is there a way to pass multiple FormValues, in `handleSubmit`, `handleSubmitSuccess` methods? I have a conditional rendered form, and it needs 3 sets of formValues.

I currently have them as optional values inside the IValues, which is kind of wrong. See below:
```
interface IFullValues extends FormikValues {
  username?: string;
  email?: string;
  password?: string;
  group?: string;
}

  handleSubmitSuccess = (
    { username, email, password, group }: IFullValues,
    formikActions: FormikActions<IFullValues>
  ) => {
    const { onSave, mode } = this.props;
    const { setSubmitting } = formikActions;

    setSubmitting(false);
    if (mode === ActionMode.EDIT_INFO) {
      onSave({ username, email, group });
    } else if (mode === ActionMode.EDIT_PASSWORD) {
      onSave({ password });
    } else {
[...]

PR #1305, opened 2019-02-06
Add example usage of the second parameter of the onSubmit method.

beausmith:
More docs = more knowledge!

Closes: https://github.com/jaredpalmer/formik/issues/1107

Related: https://github.com/jaredpalmer/formik/issues/234

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -26,7 +26,11 @@ export const MyApp: React.SFC<{}> = () => {
       <h1>My Example</h1>
       <Formik
         initialValues={{ firstName: '' }}
-        onSubmit={(values: MyFormValues, actions: FormikActions<MyFormValues>) => console.log({ values, actions })}
+        onSubmit={(values: MyFormValues, actions: FormikActions<MyFormValues>) => {
+            console.log({ values, actions });
+            alert(JSON.stringify(values, null, 2));
+            actions.setSubmitting(false)
+         }}
         render={(formikBag: FormikProps<MyFormValues>) => (
           <Form>
             <Field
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Cool. Just a heads up that `FormikActions` type will be renamed to `FormikHelpers` in v2.

docs/guides/typescript.md
```diff
@@ -26,7 +26,7 @@ export const MyApp: React.SFC<{}> = () => {
       <h1>My Example</h1>
       <Formik
         initialValues={{ firstName: '' }}
-        onSubmit={(values: MyFormValues) => alert(JSON.stringify(values))}
+        onSubmit={(values: MyFormValues, actions: FormikActions<MyFormValues>) => console.log({ values, actions })}
```
jaredpalmer:
```suggestion
        onSubmit={(values: MyFormValues, actions: FormikActions<MyFormValues>) => {
            console.log({ values, actions });
            alert(JSON.stringify(values, null, 2));
            actions.setSubmitting(false)
         }}
```

jaredpalmer: We should actually use the actions. Also, as of Formik 1.4.3 all of these types get inferred so you actually don't need these at all. 🤷‍♂️

beausmith:
@jaredpalmer - I'm stoked to hear that the types are supposed to be inferred in Formik 1.4.3. 

I am using 1.4.3 with VS Code (with the built-in `ms-vscode.vscode-typescript-tslint-plugin` plugin) and it's complaining…

image

image

I'd love to be able to have the types be inferred! Any tips to fix this?

jaredpalmer:
Ahh I meant they are inferred if you declare the function inline

PR #1306, opened 2019-02-06
Add MultistepWizard example using React Router

timburgess:
This example extends the MultistepWizard by adding React Router support.

Page locations are simply `/step/1` and `/step/2`

Key benefit is that the UI then intuitively allows page navigation using back & forward buttons for a multipage form.

I'm also suggesting some Debug style tweaks. Reading 8 point on a 4K monitor is nigh impossible.

### DIFF
--- a/examples/Debug.js
+++ b/examples/Debug.js
@@ -4,10 +4,9 @@ import { FormikConsumer } from 'formik';
 export const Debug = () => (
   <div
     style={{
-      margin: '3rem 0',
+      margin: '3rem 1rem',
       borderRadius: 4,
       background: '#f6f8fa',
-
       boxShadow: '0 0 1px  #eee inset',
     }}
   >
@@ -30,7 +29,7 @@ export const Debug = () => (
       {({ validationSchema, validate, onSubmit, ...rest }) => (
         <pre
           style={{
-            fontSize: '.65rem',
+            fontSize: '.85rem',
             padding: '.25rem .5rem',
             overflowX: 'scroll',
           }}
--- a/examples/RoutedMultistepWizard.js
+++ b/examples/RoutedMultistepWizard.js
@@ -1,0 +1,207 @@
+import React from "react";
+import { BrowserRouter as Router, Route, withRouter } from "react-router-dom";
+import { Formik, Field, ErrorMessage } from "formik";
+import { Debug } from "./Debug";
+
+const locations = ["/step/1", "/step/2"];
+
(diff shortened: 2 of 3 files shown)


charlestati:
I like the idea of routes for the wizard steps!

Maybe they should be dynamically generated from the children count rather than hardcoded though: `const locations = ["/step/1", "/step/2"];`

Also, what happens when the user goes to step 2 by typing the URL directly? Should it redirect to step 1?

I'll use this but with fragment identifiers instead as I believe they're more semantically appropriate:
`site.com/form#step1`

Thanks!

timburgess:
Cheers! I think the page paths are really going to be project-dependent. Some folks may want an explicit numbering, others may prefer to have a path that describes the page form. The code at present obviously caters more to the former than to the latter. I like your idea of using fragments.

I think the question about what happens when someone goes directly to /step/2 is actually in part a design question. In some form designs I've been working on recently, the design presents some required fields on each page and so key form values can be captured at almost any stage.

timburgess:
I've added some comments and also tried to make it more explict that the Wizard renders on all paths that match the root path.

As the example is just a two page form, using a React Router Link component would replace the existing Button. Maybe the "previous" can be made a Link on the last page.

timburgess:
I've added a `RoutedMultiStep2` as an intention to replace the current `RoutedMultiStep` example originally proposed in this PR. After working with the original example in a project, it's become clear it wasn't a great example as:
- buttons to handle each step are in a separate component to each page layout. That makes it somewhat harder to write/understand the css than it otherwise would be and dealing with situations where the page button has to be disabled become tricky
- the state management in `WizardBase` is complicated and hard to grok

The newer code is:
- simpler
- each page including submit button is a separate component
- values are not managed in a separate state
- uses `withFormik`

I'm not yet sure about the validation process for each step however. I suspect that that was what caused the existing example to be so complicated.

gl0gl0:
Hey guys,
Are there any new updates on this? I have currently a multistep wizard on my project with react router, but I wanted to integrate Formik to hopefully make it easier to handle the validations and default values; so I'm quite interested on this discussion/example :)

gl0gl0:
How would this (RoutedMultistepWizard2) approach work if you need to validate each step before going to the next one separately? 🤔

Also for this RoutedMultistepWizard2 you will always need to render the previous/next button inside the step right?

I would like to do it in the BaseForm but I don't see any other way of doing it apart from using `ref`? Any suggestions/recommendations?

Thanks 😊

t2ca:
Is it possible to create a solution that would work with Gatsby?

PR #1311, opened 2019-02-07
Fix arrayHelper `insert` and alterTouched falsy behavior

Magellol:
- Apply changes from @marcel-devdude in #1092 
- Fix a bug from which `state.touched` would have been altered even if `alterTouched` was falsy.


jaredpalmer:
There has to be cleaner way to do this.

Magellol:
Do you mean the merge or the actual logic? I haven't looked at the logic too much to be honest. I've just applied whichever changes were in #1092.

jaredpalmer:
I'm fine with this actually. I think we should rename the slices of state like `(array) =>` to like `(toucheArray) =>`, but that can be done in another PR

Issue #445, opened 2018-02-19
Auto touch on submit


fkrauthan:
## Bug, Feature, or Question?
Feature/Question

## Current Behavior
I only show error messages when a field is touched and has a error message. The problem when clicking submit it should show error messages for all fields but since they are not getting marked as touched no error shows up

## Desired Behavior
When emitting the handleSubmit event all fields in the form should be touched/there should be a flag that indicates if the form was submitted and not reseted.

- Formik Version: 0.11.10
- OS: Mac OS
- Node Version: v9.5.0
- Package Manager and version: Yarn 1.3.2

jeffbski:
I agree. I just ran into this as well. 

When we are submitting, we want the invalid fields to display their errors, so they should automatically be touched.

jeffbski:
Seems like even if I manually have called setFieldsTouched on the fields, then when handleSubmit is called it is clearing out the touched object so then the errors don't show.

jeffbski:
It looks like the problem occurs when initialValues are not specified. 

So when handleSubmit calls setNestedObjectValues with the values, they are {} and so does an Object.keys(values) and comes up with none. 

So touched is returned as {}

When using yup schema it should get the fields to touch from the schema rather than the initialValues object.

fkrauthan:
I think solutions could be as follow:

1. Use the error object keys to populate touched
1. Use the submitted value object keys to populate touched
1. Add a separate field to specify form fields (with API methods to dynamically add or remove fields)
1. Add a flag that is true when the form is submitted and gets reseted to false when form gets reseted

All 4 solutions would fix this issue. The last one is probably the easiest to implement.

jaredpalmer:
This behavior exists in formik. All fields are set to touched before submit.

jaredpalmer:
initialValues are required and should always be specified

jaredpalmer:
It cannot depend on schema because not al users use schema validation

jeffbski:
@jaredpalmer when a developer does use validationSchema then obviously that can allow formik to know the fields, so if that is supplied it could be used by formik. No it is not required, but frankly I appreciate the value of yup and even the special optional support by formik, so it wouldn't be wrong to use the schema if it is provided. 

formik + yup = fantastic

The fact that you provide this as an option gives it an advantage over the other competitors.

fkrauthan:
[...]

PR #1313, opened 2019-02-07
Update documentation according to the issue #445

ozalexo:
#445 Auto touch on submit


jaredpalmer:
This may change in v2 but I’m fine with this for now. Thanks!

PR #1316, opened 2019-02-07
Example of FieldArray component prop in TypeScript

beausmith:
This is an example using similar code/content from elsewhere in the docs. 

This PR also fixes a bug in the `src/types.ts` file where when using a component prop (such as `<FieldArray name="foo" component={FooFieldArray} />` would cause the following error:

> Type '{ name: string; component: (arrayHelpers: FieldArrayRenderProps) => Element; }' is not assignable to type 'IntrinsicAttributes & { name: string; validateOnChange?: boolean | undefined; } & SharedRenderProps<FieldArrayRenderProps> & { children?: ReactNode; }'.
  Property 'component' does not exist on type 'IntrinsicAttributes & { name: string; validateOnChange?: boolean | undefined; } & SharedRenderProps<FieldArrayRenderProps> & { children?: ReactNode; }'.

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -52,6 +52,66 @@ export const MyApp: React.SFC<{}> = () => {
 };
 ```
 
+#### Component props (`<Formik />` and `<FieldArray />`)
+
+```typescript
+import * as React from 'react'
+import { Field, FieldArray, Form, Formik, FormikActions, getIn } from 'formik'
+
+interface MyFormValues {
+  friends: string[]
+}
+
+export const FriendList = () => (
+  <div>
+    <h1>Friend List</h1>
+    <Formik
+      initialValues={{
+        friends: ['jared', 'ian', 'brent', 'beau'],
+      }}
+      onSubmit={(
+        values: MyFormValues,
+        actions: FormikActions<MyFormValues>
+      ) => {
+        console.log({ values, actions })
+        alert(JSON.stringify(values, null, 2))
+        actions.setSubmitting(false)
+      }}
+    >
+      {({ isSubmitting }) => (
+        <Form>
+          <FieldArray name="friends" component={FriendsFieldArray} />
+          <button type="submit" disabled={isSubmitting}>
+            Save
+          </button>
+        </Form>
+      )}
+    </Formik>
+  </div>
+)
+
+const FriendsFieldArray = (arrayHelpers: FieldArrayRenderProps) => {
+  const values = getIn(arrayHelpers.form.values, arrayHelpers.name);
+  return (
+    <div>
+      {values.map(
+        (firstName: string, i: number) => (
+          <div key={i}>
+            <label htmlFor={`${arrayHelpers.name}.${i}`}>Friend Name</label>
+            <Field
+              id={`${arrayHelpers.name}.${i}`}
+              name={`${arrayHelpers.name}.${i}`}
+            />
+            <button onClick={() => arrayHelpers.remove(i)}>Remove</button>
+          </div>
+        )
+      )}
+      <button onClick={() => arrayHelpers.push('')}>Add Friend</button>
+    </div>
+  )
+}
+```
+
 #### `withFormik()`
 
 ```typescript
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -242,21 +242,21 @@ export interface FormikRegistration {
 export type FormikContext<Values> = FormikProps<Values> &
   Pick<FormikConfig<Values>, 'validate' | 'validationSchema'>;
 
-export interface SharedRenderProps<T> {
+export interface SharedRenderProps<P> {
   /**
    * Field component to render. Can either be a string like 'select' or a component.
    */
-  component?: string | React.ComponentType<T | void>;
+  component?: string | React.ComponentType<P>;
 
   /**
    * Render prop (works like React router's <Route render={props =>} />)
    */
-  render?: ((props: T) => React.ReactNode);
+  render?: ((props: P) => React.ReactNode);
 
   /**
    * Children render function <Field name>{props => ...}</Field>)
    */
-  children?: ((props: T) => React.ReactNode);
+  children?: ((props: P) => React.ReactNode);
 }
 
 export type GenericFieldHTMLAttributes =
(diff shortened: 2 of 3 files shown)


Andreyco:
> * Replace "any" in `(arrayHelpers: any)` with the correct type for the `ArrayField` component props.

it is `FieldArrayRenderProps` type exported from `FieldArray.tsx` file

beausmith:
@Andreyco - Thanks for the tip. I was also convinced that `FieldArrayRenderProps` is correct. But when I used it I still got errors. So I went digging a little deeper and took a look at `src/types.ts` and it looks like the issue was with the types.

No more Typescript errors! 🎉 

I've forced pushed to update the commit. Take a look and LMK if you have any feedback.

cc @jaredpalmer - This PR is ready to roll!

beausmith:
🛎 Anything I can do to make this PR more mergeable @jaredpalmer? Or anything I don't know about the roadmap which makes this PR moot? Thanks!

src/types.tsx
```diff
@@ -242,21 +242,21 @@ export interface FormikRegistration {
 export type FormikContext<Values> = FormikProps<Values> &
   Pick<FormikConfig<Values>, 'validate' | 'validationSchema'>;
 
-export interface SharedRenderProps<T> {
+export interface SharedRenderProps<P> {
   /**
```
Andreyco: I would not alter type name. It is a common among TS devs (and others) to reference to generic by `T` type. Less changes, more likely PR gets merged ;)

src/types.tsx
```diff
@@ -242,21 +242,21 @@ export interface FormikRegistration {
...
    * Field component to render. Can either be a string like 'select' or a component.
    */
-  component?: string | React.ComponentType<T | void>;
```
Andreyco: Are there any downside/upsides of removing default `void` type?

docs/guides/typescript.md
```diff
@@ -52,6 +52,63 @@ export const MyApp: React.SFC<{}> = () => {
...
+          </button>
+        </Form>
+      )}
+    </Formik>
+  </div>
+)
+
+const FriendsFieldArray = (arrayHelpers: FieldArrayRenderProps) => (
+  <div>
+    {arrayHelpers.form.values[arrayHelpers.name].map(
```
Andreyco: Using arrow function is 👍 , personally I would store values in readable var instead. No big deal.

```
const FriendsFieldArray = (arrayHelpers: FieldArrayRenderProps) => {
  const values = getIn(arrayHelpers.form.values, arrayHelpers.name);
  return values.map(...)
}
```

Using `getIn` allows you use "dot" patch, like `user.firstName`.

src/types.tsx
beausmith: True. I'll change it.

beausmith: On second thought… type types which are passed to `SharedRenderProps` are props which are used on `React.ComponentType`. This was the reason that I used "P" instead of "T". All the examples I see where types are passed to `React.ComponentType` they are passed using "P" or some longer variable with the word "props" in it somewhere.

src/types.tsx
beausmith: The use of `void` is what was causing the bug I was running into. When I was looking around to see how I don't see how types were passed to `React.ComponentType`, I didn't see any using `void` which is what lead to me try removing it. When removed it, I no longer had any errors.

docs/guides/typescript.md
beausmith: Nice! Thanks for the tip about `getIn()` method. I hadn't seen this before. Will update PR.

src/types.tsx
beausmith: Also when I opened up `src/types.d.ts`, the following typescript errors were shown:

image

Again, removing the `| void` also resolved all these errors.

beausmith:
Thank you for the review @Andreyco!

I have updated the PR to contain two commits: 
- Fix the bug and update test to test for bug.
- Add the documentation for using FieldArray with component prop. Updated to use `getIn()`!

src/types.tsx
beausmith: Here's the actual `ComponentType` type from DefinitelyTyped:

```ts
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L63

Note that it uses `P` not `T`

src/types.tsx
beausmith: Here's the actual `ComponentType` type from DefinitelyTyped:

```ts
type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L63

Note that `P` defaults to `{}` if unspecified… and thus `void` is not used AFAIK.

beausmith:
Is this useful @jaredpalmer?

docs/guides/typescript.md
```diff
@@ -52,6 +52,66 @@ export const MyApp: React.SFC<{}> = () => {
...
+          <button type="submit" disabled={isSubmitting}>
+            Save
+          </button>
+        </Form>
+      )}
+    </Formik>
+  </div>
+)
+
+const FriendsFieldArray = (arrayHelpers: FieldArrayRenderProps) => {
```
Yankovsky: `FieldArrayRenderProps` should be `FieldArrayRenderProps<Values extends FormikValues>` instead. Currently it just erase info about types.

ayhid:
Does anyone know why this PR was closed please ?
I am struggling with my FieldArray component typing as well. It's a good idea to have some examples in the documenbtation for noobs like me

ayhid:
Does anyone know why this PR was closed please ?
I am struggling with my FieldArray component typing as well. It's a good idea to have some examples in the documenbtation for noobs like me

Philipp91:
+1 for having an example, as this is a tricky situation. And also I agree with this:

> Again, removing the `| void` also resolved all these errors.

Why is it [there](https://github.com/formium/formik/blob/b9cc2536a1edb9f2d69c4cd20ecf4fa0f8059ade/packages/formik/src/types.tsx#L261) to begin with?

Issue #1238, opened 2019-01-14
FastField: fallback to default shouldComponentUpdate implementation even if custom shouldUpdate implementation returns false


Verbon:
### Current Behavior

Default `FastField`'s `shouldComponentUpdate` implementation is not triggered, if custom `shouldUpdate` implementation returns `false`.

### Desired Behavior

If `FastField`'s custom `shouldUpdate` implementation returns `false`, then `FastField`'s `shouldComponentUpdate` default implementation should be triggered.

### Suggested Solution

**formik/src/FastField.tsx**

Current `FastField`'s `shouldComponentUpdate` implementation:
```
shouldComponentUpdate(
    props: FastFieldAttributes<Props> & { formik: FormikContext<Values> }
  ) {
    if (this.props.shouldUpdate) {
      return this.props.shouldUpdate(props, this.props);
    } else if (
      getIn(this.props.formik.values, this.props.name) !==
        getIn(props.formik.values, this.props.name) ||
      getIn(this.props.formik.errors, this.props.name) !==
        getIn(props.formik.errors, this.props.name) ||
      getIn(this.props.formik.touched, this.props.name) !==
        getIn(props.formik.touched, this.props.name) ||
      Object.keys(this.props).length !== Object.keys(props).length ||
      this.props.formik.isSubmitting !== props.formik.isSubmitting
    ) {
      return true;
    } else {
      return false;
    }
  }
```
should be changed to:
```
shouldComponentUpdate(
    props: FastFieldAttributes<Props> & { formik: FormikContext<Values> }
  ) {
    if (this.props.shouldUpdate && this.props.shouldUpdate(props, this.props)) {
      return true;
    } else if (
      getIn(this.props.formik.values, this.props.name) !==
        getIn(props.formik.values, this.props.name) ||
      getIn(this.props.formik.errors, this.props.name) !==
        getIn(props.formik.errors, this.props.name) ||
      getIn(this.props.formik.touched, this.props.name) !==
        getIn(props.formik.touched, this.props.name) ||
      Object.keys(this.props).length !== Object.keys(props).length ||
      this.props.formik.isSubmitting !== props.formik.isSubmitting
    ) {
      return true;
    } else {
      return false;
    }
  }
```

### Who does this impact? Who is this for?

All the people!

### Describe alternatives you've considered

[...]

PR #1317, opened 2019-02-07
FastField: fallback to default shouldComponentUpdate implementation even if custom shouldUpdate implementation returns false (#1238)

Verbon:
Fixes https://github.com/jaredpalmer/formik/issues/1238


Verbon:
@jaredpalmer Have you had a chance to look at it yet?

PR #1319, opened 2019-02-08
Call preventDefault in handleReset

maxdavidson:
This aligns the behavior of `handleReset` with `handleSubmit`, calling `event.preventDefault()` if an event is supplied. I noticed the form's reset event had unexpected side effects otherwise, resetting `select` elements to the first option and not the controlled value.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -537,7 +537,11 @@ export class Formik<Values = FormikValues> extends React.Component<
     });
   };
 
-  handleReset = () => {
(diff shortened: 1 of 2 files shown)


jaredpalmer:
hrmmm let me try this out.

maxdavidson:
Here's an example that demonstrates the issue https://codesandbox.io/s/502pxwow9x

jaredpalmer:
Continued in #1629

PR #1348, opened 2019-02-21
Fix warnings

sijad:
first argument passed to `tiny-warning` should be falsy to warn.

https://github.com/alexreardon/tiny-warning/blob/dc65095d4e27290869781c778950ed82a91de866/src/index.js#L9

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -116,8 +116,7 @@ export function useFormik<Values = object>({
     };
   }>({});
   const [state, dispatch] = React.useReducer<
-    FormikState<Values>,
-    FormikMessage<Values>
+    React.Reducer<FormikState<Values>, FormikMessage<Values>>
   >(formikReducer, {
     values: props.initialValues,
     errors: {},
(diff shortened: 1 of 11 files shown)


jaredpalmer:
Test failed because of my other type errors. I just upgraded TS and @types/react in the hooks branch

jaredpalmer:
Merging with latest hooks

jaredpalmer:
Thanks for this! Going to merge and cut a release

src/Formik.tsx
```diff
@@ -314,10 +314,8 @@ export function useFormik<Values = object>({
         activeElement instanceof HTMLButtonElement
       ) {
         warning(
-          !!(
```
buzinas: If it was `!!`, shouldn't it be `!` now? It seems that this will start triggering false positives.

sijad: nice catch, thanks

jaredpalmer:
I went down a rabbit hole on this one. We need to basically rearchitect our tests to use `act()`, but really need Sunil's `await act()` PR to land in the next React release. Until then, our tests will have warnings

src/Formik.tsx
jaredpalmer: The test fails if you make this change

PR #1353, opened 2019-02-21
Fix for Windows Paths

### LINKED ISSUE #1352: Rollup config doesn't detect Windows Paths
### Current Behavior

On Windows, running `npm start`, `yarn start`, etc does not work. The `external()` function in rollup.config.js doesn't detect Windows Paths, so it thinks `C:\path\to\formik\compiled\Field.js` is an external node module.

### Expected behavior

The project should build on Windows for any Windows Contributors.

johnrom:
This adds a check for `path.isAbsolute()` instead of `path.startsWith('/')`, since absolute Windows paths don't start with `/`.

Closes #1352

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,3 +1,4 @@
+import path from 'path';
 import commonjs from 'rollup-plugin-commonjs';
 import replace from 'rollup-plugin-replace';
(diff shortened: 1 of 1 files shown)


Andreyco:
Looks good. cc @jaredpalmer

PR #1361, opened 2019-03-01
Accessors for FieldArray

wenkepaul:
The goal of this PR is to expose `ArrayHelpers` outside of FieldArray's `render` method.  It would be convenient to be able to just pass around FormikProps and call `myForm.getFieldArrayHelpers(myFieldName)` when necessary.  Similarly to form values and errors, the array helpers could also be accessed via `myForm.fieldArrayHelpers`.  What are your thoughts on these two API modifications?  Development is not complete, I'm looking for feedback on the initial concept and the API modifications.

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -299,6 +299,8 @@ class FieldArrayInner<Values = {}> extends React.Component<
       name,
     };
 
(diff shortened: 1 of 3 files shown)


wenkepaul:
@jaredpalmer . Just circling back to this.  Is this a feature you'd guys be interested in?  If so, I'll continue development.

PR #1382, opened 2019-03-14
Let user drop props through FieldArray, like Field

dalevfenton:
I've read #561, #693 and while you can user render and a closure to accomplish it, it would be nice if `<FieldArray />` would pass all props down to a `component` prop.

This PR implements it pretty naively, but please do consider adding it!

I'm happy to make edits or build the typings out better if it helps.

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -12,6 +12,7 @@ import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
 export type FieldArrayRenderProps = ArrayHelpers & {
   form: FormikProps<any>;
   name: string;
+  [key: string]: any;
 };
 
 export type FieldArrayConfig = {
@@ -19,7 +20,10 @@ export type FieldArrayConfig = {
   name: string;
   /** Should field array validate the form AFTER array updates/changes? */
   validateOnChange?: boolean;
+  /** Allow users to pass props through to component/render/children */
+  [key: string]: any;
 } & SharedRenderProps<FieldArrayRenderProps>;
+
 export interface ArrayHelpers {
   /** Imperatively add a value to the end of an array */
   push: (obj: any) => void;
@@ -202,12 +206,16 @@ class FieldArrayInner<Values = {}> extends React.Component<
       },
       (array: any[]) => {
         const arr = array ? [null, ...array] : [null];
-        if (length < 0) length = arr.length;
+        if (length < 0) {
+          length = arr.length;
+        }
         return arr;
       },
       (array: any[]) => {
         const arr = array ? [null, ...array] : [null];
-        if (length < 0) length = arr.length;
+        if (length < 0) {
+          length = arr.length;
+        }
         return arr;
       }
     );
@@ -291,9 +299,12 @@ class FieldArrayInner<Values = {}> extends React.Component<
         validationSchema: _validationSchema,
         ...restOfFormik
       },
+      validateOnChange,
+      ...passedProps // grab any props dropped in by a user
     } = this.props;
 
     const props: FieldArrayRenderProps = {
+      ...passedProps,
       ...arrayHelpers,
       form: restOfFormik,
       name,
--- a/test/FieldArray.test.tsx
+++ b/test/FieldArray.test.tsx
@@ -37,6 +37,26 @@ describe('<FieldArray />', () => {
     );
   });
 
+  it('renders component and passes exra props down', () => {
+    const TestComponent = (props: any) => {
+      expect(props.userProps === 'test');
+      return null;
+    };
+
+    ReactDOM.render(
+      <TestForm
+        component={() => (
+          <FieldArray
+            name="friends"
+            userProp="test"
+            component={TestComponent}
+          />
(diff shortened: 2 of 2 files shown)


test/FieldArray.test.tsx
```diff
@@ -37,6 +37,26 @@ describe('<FieldArray />', () => {
     );
   });
 
+  it('renders component and passes exra props down', () => {
```
hugonasciutti: Just a little misspelling.
```suggestion
  it('renders component and passes extra props down', () => {
```

dalevfenton: good catch 👍

patrickgordon:
This would be a great addition!

src/FieldArray.tsx
```diff
@@ -12,14 +12,18 @@ import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
 export type FieldArrayRenderProps = ArrayHelpers & {
   form: FormikProps<any>;
   name: string;
+  [key: string]: any;
```
johnrom: This isn't typed strongly enough. Instead it should be something like `FieldArrayRenderProps<ExtraProps> = ArrayHelpers & {form, name} & ExtraProps;`

src/FieldArray.tsx
```diff
@@ -12,14 +12,18 @@ import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
...
 };
 
 export type FieldArrayConfig = {
```
johnrom: `FieldArrayConfig<ExtraProps>`

src/FieldArray.tsx
```diff
@@ -12,14 +12,18 @@ import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
...
   /** Really the path to the array field to be updated */
   name: string;
   /** Should field array validate the form AFTER array updates/changes? */
   validateOnChange?: boolean;
+  /** Allow users to pass props through to component/render/children */
+  [key: string]: any;
 } & SharedRenderProps<FieldArrayRenderProps>;
```
johnrom: `SharedRenderProps<FieldArrayRenderProps<ExtraProps>>` or something

test/FieldArray.test.tsx
```diff
@@ -37,6 +37,26 @@ describe('<FieldArray />', () => {
...
+  it('renders component and passes extra props down', () => {
+    const TestComponent = (props: any) => {
+      expect(props.userProps === 'test');
+      return null;
+    };
+
+    ReactDOM.render(
+      <TestForm
+        component={() => (
+          <FieldArray
```
johnrom: should be `<FieldArray<ExtraProps> />` which uses all the ExtraProps declarations I commented above.

dalevfenton:
I'll try to get the typings updated for this PR this week.

test/FieldArray.test.tsx
dalevfenton: I was able to write the types above pretty easily, but `FieldArray`, and `Field` do not accept generic parameters right now so I'm going to see if I can figure that part out tonight maybe

johnrom: I have a PR where I implemented `Field` strong types and am in the process of updating it for v2. I didn't touch FieldArray though. Another individual also has a Field typing PR

#1336 
#1786

soulchainer:
How is this going, @dalevfenton? Are typings the only thing which is holding this back? Are you planning to adding them to this PR?
It seems like this is paused since September :(. I'm having right now an issue which would be easily solved just being able to pass props through `FieldArray`, instead of doing some workaround to get the props passed somehow.

dalevfenton:
> How is this going, @dalevfenton? Are typings the only thing which is holding this back? Are you planning to adding them to this PR?
> It seems like this is paused since September :(. I'm having right now an issue which would be easily solved just being able to pass props through `FieldArray`, instead of doing some workaround to get the props passed somehow.

I forgot all about this to be honest. I tried to implement the generic typings but wasn't able to get it to work correctly. If you want to try and take this over that would be great. I don't expect that I'll have time to work on this in the near future myself.

johnrom:
There's an issue called Strongly Typed Fields with a PR where I did the generic work on Field w/ ExtraProps. Might be a useful example.

It's based on v1 currently.

soulchainer:
@dalevfenton @johnrom I will give it a try then, thanks :). «Soon» (because this weekend in kinda packed already :sweat_smile:).

soulchainer:
> @dalevfenton @johnrom I will give it a try then, thanks :). «Soon» (because this weekend is kinda packed already sweat_smile).

I already took a look into this (yesterday and today) and... Couldn't' get it working :sweat_smile:
I still have to learn TypeScript properly (we use Formik but no TypeScript) and this escapes to my understanding.
I will leave it to someone well versed in both, Formik and TypeScript, sorry :þ.

johnrom:
@soulchainer I think it took me an entire day to figure out the types on my branch, especially in a way that would be backwards compatible. No worries!

If I end up strengthening types I'll circle back to this item as well.

PR #1391, opened 2019-03-24
Fix missing displayName for Form component

yoursdearboy:
`Form` component has no `displayName`. Current code changes `displayName` of connected component, not `Form` component.

See screenshots of dev tools.

Before
After


jaredpalmer:
Fix tests

jaredpalmer:
This has been fixed in v2. Closing.

Issue #485, opened 2018-03-05
Add optional onChange property to formik which will be called when fields change


jeffbski:
## Feature

New feature, allow for an optional onChange property to be supplied to formik that would allow external notification of changes before things are submitted.

For instance, we often start out not needing to know about form values until they are submitted, but occasionally we do need to do additional things based on the values as they change. So Formik could allow us to pass in an additional onChange property which will be called whenever the values change. This would allow external systems to be called as needed.

## Current Behavior

There is no way to notify about field changes until submitted.

## Desired Behavior

Optionally it would be good to have the ability to notify about changes. This would allow certain fields to be monitored and enable updates based on changes.

## Suggested Solutions

Add optional property onChange to formik. If a function is provided then it will be called with the latest values and formik bag every time any field changes. 

onChange: (values: Values, formikBag: FormikBag) {
  // I can be notified of the updated values and can use them
}

## Additional Information

---

- Formik Version: v1.0.0-alpha.2
- React Version: any
- TypeScript Version: any
- CodeSandbox Link:
- OS: any
- Node Version: any
- Package Manager and Version: any

jeffbski:
yes, #401 would be a nice way to handle that. It would make it easy to know when a specific value changed, etc. Thanks.

wachunga:
@jaredpalmer The linked proposal was abandoned, but there are plenty of use cases for `onChange`. I think it's worth re-opening this. Here's one use case:

### Filter form

A typical filter form doesn't have a submit button, but every interaction with selects, checkboxes, radios etc immediately update the UI. It's convenient to have an `onChange` as proposed above to update the UI.

Issue #271, opened 2017-11-27
onChange for entire form


timoxley:
I have a form where some field's values depend on other fields values. I have been using redux-form and have been able to calculate these values using an [`onChange` handler on the top-level form object](https://redux-form.com/7.1.2/docs/api/reduxform.md/#-code-onchange-function-code-optional-).

I don't see an equivalent using formik, is this right? I thought I could hijack the validation function to do this, but it's not passed the formik bag, only the values to validate.

Another usecase for a form-level `onChange` is automatic persistence without needing to hit a submit button e.g. something like:

```js
<Formik
  initialValues={JSON.parse(localStorage.getItem('formvalues'))}
  onChange={(values) => {
    localStorage.setItem('formvalues', JSON.stringify(values))
  }}
>
{/* … */}
</Formik>
```

Another option could be a "submitOnChange` flag.

prichodko:
You will always receive latest `values` in render method. So you can perform anything you need there. I also recommend having a look at https://github.com/jaredpalmer/formik-persist how such functionality could be implemented. Take a note that `<FormikPersist />` takes values from the context, you can of course pass it as prop instead.

jaredpalmer:
@prichodko is exactly correct.

benvan:
Sorry to reopen a closed thread, but this seemed like the most appropriate place to follow up on this.

If I understand correctly, the suggestion is to use another component to simulate onChange responses. I can see how that might make sense for something extremely agnostic like FormikPersist, but for the general situation where you wish to cause some effect based on form changes, this doesn't make sense to me.

Consider a simple form which contains dropdowns for the ingredients of a burger. You wish to render a preview of the burger _outside_ the form somewhere. It would seem fairly simple (and conventional) to do this by passing the form an "onBurgerIngredientsChanged" handler, and hooking that up into the form's onChange handler.

Am I correct in understand you would suggest instead creating a `<BurgerIngredientsChangedDetector />` component, complete with constructor, componentDidMount, componentWillReceiveProps methods ... 

It seems to me this would very naturally lead people to creating their own `<OnFormUpdate hander={someHandler}/>` component. In this case... why the extra complexity? Would it not be better to be able to pass Formik an onChange handler?

Just my two cents - I'm facing this problem now with a few forms that are slightly more dynamic than their siblings. Any thoughts would be much appreciated.

Thanks

ramitos:
[...]

PR #1397, opened 2019-03-26
Adding notifyOnChange as a side effect to succesful handleChange

rohanBagchi:
#### Description
This PR adds a side-effect to formik's `handleChange` which can be used to react to any update to formik form state.
The effect is similar to [formik-effect](https://github.com/jaredpalmer/formik-effect) except built into formik itself without the need to install any additional library or retrofitting formik's validate hook to perform the callback.

#### Usage
In my project I had to broadcast a form dirty signal whenever user begins typing on it. This I implemented by triggering said signal from validate.
While this works for my use case, I cannot now make the form validate on blur as this would delay triggering of dirty signal.

If this change gets accepted, the users can add their own callback to react to formik state update like this:

```
<Formik
    {...}
    notifyOnChange={this.notifyOnChange}
>
    {...}
</Formik>
```

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -356,6 +356,10 @@ export class Formik<Values = FormikValues> extends React.Component<
             if (this.props.validateOnChange) {
               this.runValidations(setIn(this.state.values, field!, val));
             }
+
+            if (this.props.notifyOnChange) {
+              this.props.notifyOnChange(this.state.values);
+            }
           }
         );
       }
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -217,6 +217,11 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
   validate?: ((
     values: Values
   ) => void | object | Promise<FormikErrors<Values>>);
+
+  /**
+   * gets called after every formik state update
+   */
+  notifyOnChange?: (values: Values) => void;
 }
 
 /**
--- a/test/Formik.test.tsx
+++ b/test/Formik.test.tsx
@@ -223,6 +223,26 @@ describe('<Formik>', () => {
     });
   });
 
+  describe('notifyOnChange', () => {
+    it('should trigger notifyOnChange on succesful state update', () => {
+      const notifyOnChange = jest.fn();
+      const { getProps, getByTestId } = renderFormik({
+        notifyOnChange,
+      });
+
+      const input = getByTestId('name-input');
+      fireEvent.change(input, {
+        persist: noop,
+        target: {
+          name: 'name',
+          value: 'ian',
+        },
+      });
+
+      expect(notifyOnChange).toHaveBeenCalledWith(getProps().values);
+    });
+  });
+
   describe('handleBlur', () => {
     it('sets touched state', () => {
       const { getProps, getByTestId } = renderFormik();


tristanmarsh:
Yes please!

rohanBagchi:
@jaredpalmer can we merge this? or if you have any other solution in mind, please revert back.

animify:
Need this! 👍

ducdev:
Need something like this. Please consider merging.

ruiz-inpt:
@jaredpalmer after jumping around a few issues referencing the same problem I realized there are a lot of people facing it. The form level `onChange` seems such a common use case that with little effort would make `formik` an even better form library... however the latest threads seem ignored by the project maintainers. Please, consider responding (even if the answer is *won't do*) to the latest comments on these threads:
https://github.com/jaredpalmer/formik/issues/485
https://github.com/jaredpalmer/formik/issues/271
https://github.com/jaredpalmer/formik/issues/401
https://github.com/jaredpalmer/formik/pull/1397
I'm just trying to solve this problem on my project on day one of adopting **formik** (which I liked a lot) and after reading all those threads I still don't know  what should I do. Having to install a 3rd party package to solve such a common use case doesn't feel good to me.

ebrearley:
> @jaredpalmer after jumping around a few issues referencing the same problem I realized there are a lot of people facing it. The form level `onChange` seems such a common use case that with little effort would make `formik` an even better form library... however the latest threads seem ignored by the project maintainers. Please, consider responding (even if the answer is _won't do_) to the latest comments on these threads:
> #485
> #271
> #401
> #1397
> I'm trying to solve this problem on my project on day one of adopting **formik** (which I liked a lot) and after reading all those threads I still don't know what should I do. Having to install a 3rd party package to solve such a common use case doesn't feel good to me.

Maybe this PR isn't required since it's possible to achieve with the existing API. See my comment in https://github.com/jaredpalmer/formik/issues/271#issuecomment-518679767

jaredpalmer:
This is possible with the existing API. There is no need for this addition.

serusko:
hi @jaredpalmer culd you please add some ref to this solution, i think a lot of people end here with some googleing :) so we could prevent reopens or duplicates :)

tom-wagner:
Boo!!! no fun!!!

Cyclodex:
For me, this article here helped: https://plainenglish.io/blog/how-to-listen-to-formik-onchange-event-in-react

wchargin:
An important advantage of the approach proposed in this PR (and the other linked threads), compared to the `useEffect`-based "Formik observer" approaches, is that the change notification takes effect in the same render cycle.

When you use a `useEffect` to extract the change event, the flow is:

 1. User types in a text field, causing a React `onChange` event.
 2. `onChange` handler calls into Formik `handleChange`, which updates state.
 3. Component re-renders with new state, and the `useEffect` dependency changes.
 4. After rendering, the `useEffect` callback is called, and itself calls some React state setter (e.g., to smuggle the Formik state out to a different component to render a preview).
 5. Affected components re-render *again*, with the same Formik state but new state that was changed due to the Formik observer.

This is unnecessary! It creates visible lag and jank, in addition to useless renders. The React docs ["You Might Not Need an Effect"][ymmn] are all about how to remove instances of this general pattern of using effects to observe changes and call state setters. Effects are for subscribing to external stores or managing event listeners on imperative components, not communicating React state around.

If Formik exposed an `onChange` handler, as proposed in this PR, this would all be done in a single render cycle. I am actually *not* aware of a way to achieve that with the existing Formik APIs.

[ymmn]: https://react.dev/learn/you-might-not-need-an-effect

PR #1398, opened 2019-03-28
Change setFieldValue to return a Promise #529

nzyoni:
setFieldValue should return a Promise or receive a callback #529

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -397,6 +397,8 @@ export class Formik<Values = FormikValues> extends React.Component<
               this.runValidations(this.state.values)
                 .then(resolve)
                 .catch(reject);
+            } else {
+              resolve();
             }
           }
         );
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -384,20 +384,26 @@ export class Formik<Values = FormikValues> extends React.Component<
     value: any,
     shouldValidate: boolean = true
   ) => {
-    if (this.didMount) {
-      // Set form field by name
-      this.setState(
-        prevState => ({
-          ...prevState,
-          values: setIn(prevState.values, field, value),
-        }),
-        () => {
-          if (this.props.validateOnChange && shouldValidate) {
-            this.runValidations(this.state.values);
+    return new Promise((resolve, reject) => {
+      if (this.didMount) {
+        // Set form field by name
+        this.setState(
+          prevState => ({
+            ...prevState,
+            values: setIn(prevState.values, field, value),
+          }),
+          () => {
+            if (this.props.validateOnChange && shouldValidate) {
+              this.runValidations(this.state.values)
+                .then(resolve)
+                .catch(reject);
+            }
           }
-        }
-      );
-    }
+        );
+      } else {
+        resolve();
+      }    
+    })
   };
 
   handleSubmit = (e: React.FormEvent<HTMLFormElement> | undefined) => {


RigoTheDev:
@jaredpalmer

src/Formik.tsx
```diff
@@ -384,20 +384,28 @@ export class Formik<Values = FormikValues> extends React.Component<
...
+        // Set form field by name
+        this.setState(
+          prevState => ({
+            ...prevState,
+            values: setIn(prevState.values, field, value),
+          }),
+          () => {
+            if (this.props.validateOnChange && shouldValidate) {
+              this.runValidations(this.state.values)
+                .then(resolve)
```
hertzg:
```suggestion
                .then(() => resolve())
```
To make sure the promise returned resolves with the same result in all cases.
This introduces inconsistency in resolved values as the `runValidations` will resolve to `FormikErrors<Values>` which will be passed as a result of our promise while in other cases below our promise is resolved with no value (just `resolve()`).

src/Formik.tsx
```diff
@@ -384,20 +384,28 @@ export class Formik<Values = FormikValues> extends React.Component<
     value: any,
     shouldValidate: boolean = true
   ) => {
```
hertzg: Would be nice to have a return type defined, this would have helped detect the inconsistency issue I've mentioned below
```suggestion
  ) : Promise<void> => {
```

eboutin:
I'm really looking forward to this PR being merged!
Yesterday I realised I needed setFieldValue to return a Promise to fix a bug in my app. This morning, before starting working on it I realised nzyoni already did it! Thanks!
Please let me know if I can help move this forward.

LukeAskew:
Any update on this?

Sutikshan:
Can handleChange also be modified to return `executeChange`, which can return the `setFieldValue`. This will make whoel chain promisify.

src/Formik.tsx
```diff
@@ -384,20 +384,28 @@ export class Formik<Values = FormikValues> extends React.Component<
...
-      // Set form field by name
-      this.setState(
-        prevState => ({
-          ...prevState,
-          values: setIn(prevState.values, field, value),
-        }),
-        () => {
-          if (this.props.validateOnChange && shouldValidate) {
-            this.runValidations(this.state.values);
+    return new Promise((resolve, reject) => {
```
konojunya: The code looks a bit complicated. I think the reason is that nesting is deep. So you should use early return. The following code looks good.

```tsx
return new Promise((resolve, reject) => {
  if(!this.didMount) {
    resolve();
    return;
  }
  
  this.setState(prevState =>({
    ...prevState,
    values: setIn(prevState.values, field, value)
  }),
  () => {
    if(!this.props.validateOnChange || !shouldValidate) {
       resolve();
       return;
    }

    this.runValidations(this.state.values)
      .then(() => resolve())
      .catch(reject);
  })
})
```

ajorlando:
Any update on this?

inkiltie:
Any update on this?

gauthamchandra:
Hi 👋🏼  @jaredpalmer @nzyoni @konojunya @hertzg 

Sorry for all the `@username`s.

I see that this PR was closed. If the only blockers are just a matter of addressing your comments to the PR above, I can happily take this over and reopen the PR with the proper fixes made to address them. 😄 

Please let me know! I know a few people on our team who would love this change to make it into Formik!

johnrom:
@gauthamchandra there isn't really a way to theoretically return the result of a setFieldValue to your code. Because Formik v1-2 uses a reducer + dispatches, the only way to get the result of the setFieldValue call is technically to wait for an effect and React itself to commit the changes.

In v3 #3231 I have adopted a redux-like API where we keep Formik's state separate from React commits, and React commits replay changes to Formik. At this point, setFieldValue could thus be awaited and the end of setFieldValue could be the resulting getState() -- but the caveat of that would be that other changes such as validation or even future state changes could theoretically be applied by the time the Promise is returned.

When using React setState callbacks, other changes can also be applied during the render commit phase so I don't think this would be an issue.

PR #1404, opened 2019-03-30
[v2] Add handling backend errors with async submit

LKay:
Small improvement to add custom error wrapper to handle validation errors from rejected submit promise. A little more details is in the added docs file: [docs/api/FormikSubmitError.md](https://github.com/jaredpalmer/formik/compare/hooks...LKay:lkay/async-submit-errors?expand=1#diff-cead737648972a5caf76605302da6084)

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -21,6 +21,7 @@ import {
   makeCancelable,
 } from './utils';
 import { FormikProvider } from './FormikContext';
+import { FormikSubmitError } from './FormikSubmitError';
 import warning from 'tiny-warning';
 
 // We already used FormikActions. So we'll go all Elm-y, and use Message.
@@ -598,8 +599,11 @@ export function useFormik<Values = object>({
               dispatch({ type: 'SUBMIT_SUCCESS' });
             }
           })
-          .catch(_errors => {
+          .catch(error => {
             if (isMounted.current) {
+              if (error instanceof FormikSubmitError) {
+                dispatch({ type: 'SET_ERRORS', payload: error.errors });
+              }
               dispatch({ type: 'SUBMIT_FAILURE' });
             }
           });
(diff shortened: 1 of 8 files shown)


src/Formik.tsx
```diff
@@ -598,8 +599,11 @@ export function useFormik<Values = object>({
               dispatch({ type: 'SUBMIT_SUCCESS' });
             }
           })
-          .catch(_errors => {
+          .catch(error => {
             if (isMounted.current) {
+              if (error instanceof FormikSubmitError) {
+                dispatch({ type: 'SET_ERRORS', payload: error.errors });
```
jaredpalmer: So this will get overwritten on the next key stroke. We should either create a new state key just for this error or use `status.submitError`

LKay: My idea is to provide a simple way to handle server side validation errors. Since submit handler is a `Promise` now, we could leverage rejected promise state.  I'm all into improving this.

Can you elaborate what do you mean by using `status.submitError`? 

About the first proposal I don't think having another key makes much sense as we already have `SET_ERRORS` and `SUBMIT_FAILURE` and having ie. `SUBMIT_FAILURE_ERRORS` would be redundant. My other idea was to add optional payload (with errors) to `SUBMIT_FAILURE`, bu I'd rather separate setting form errors and submission state. What do you think?

jaredpalmer: My point is just that `errors` gets overwritten on the very next key stroke because validation will run. If you, the dev, want to persist these API errors in your UI until the next submission, then `errors` can't be used to store this data for the aforementioned reason.

DennisSnijder:
Hmm, no point in doing this right? with the `onSubmit` you can simply use the passed in property like this: 

```js
function handleMySubmission(values, {setErrors})
```

johnrom:
Related issue: https://github.com/jaredpalmer/formik/issues/150

Issue #1371, opened 2019-03-08
RFC: Improving Field


jaredpalmer:
I want to take folk's temperature on some warts that I want to change related to `<Field>` as well as discuss how best to move forward in a way that doesn't force people to rewrite stuff immediately.

## `<Field component>`

This prop is the source of a ton of confusion because it does not simply pass through `onChange`, `onBlur`, and `value` but rather `field` and `form`. This means that there is some inconsistent behavior with respect to `<Field component="select">` vs. `<Field component={MyInput}>`. I think a decent solution is to create a new prop called `is` or `as` and make that prop just pass through `onBlur`, `onChange`, `value`/`checked` as expected. We then would put a deprecation warning on `component` instructing people to use `as`/`is` or `children` and eventually remove `component` entirely in future versions. Unfortunately, this will force folks to change every single `"select"` and `"textarea"` over time, but this at least avoids breaking changes

## `<Field render>`

This is just redundant at this point. We should put up a deprecation warning suggesting to use `children` as a function. 

## `<Field children>`

We should merge #343 and add a `meta` to `<Field children>`.

```diff
<Field>
-  {({field, form }) => ... }
+  {({field, form, meta }) => ... }
</Field>
```

## Next Steps

In a perfect world, we make these changes gradually in a way that prevents people from having to rewrite stuff. I definitely don't want to have to rewrite all my inputs, but we need to make these changes at some point. We also need to consider that `<Field>` will have a dramatically reduced usage in the future once the `useField` hook is released. 

## Umbrella Issues

- Field `meta` (#343)
- Deprecate `render` and `component` (#1406)

VincentLanglet:
I agree passing `onChange`, `onBlur`, `value` and `name` directly instead of inside a `field` props is less confusing. It will be similar to component given as string and add more compatibility with other component, like Material-ui. Actually we have to do:
```
<Field name="email" type="text" component={(props) => <TextField {...props.field} {...props} />} />
```

But, it's not clear for me. 
Do you still consider passing the `form` props or not ? I think this one can still be useful.

WickyNilliams:
Sounds good to me. Would be simpler (or unnecessary) to adapt existing components to Formik

Andreyco:
@VincentLanglet 
> But, it's not clear for me.
> Do you still consider passing the form props or not ? I think this one can still be useful.

From original post it is most likely a NO. `onChange`, `onBlur` and `value` props only. Passing `form` is redundant.

[...]

PR #1406, opened 2019-04-01
Create <Field as> and deprecate <Field component> and <Field render>

### LINKED ISSUE #1407: Add <Field as> prop
See #1371

### LINKED ISSUE #1408: Deprecate <Field render> and <Field component>
See #1371

jaredpalmer:
- Deprecate `<Field component>` and `<Field render>`. These haven't been removed. However, they will yield a warning in the console. 
- Add `<Field as>` prop which just passes related handlers and props straight through. This should make Styled Components folks happy.

Related to #1371
Closes #1407 
Closes #1408

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -30,15 +30,15 @@ export interface FieldConfig {
   component?:
     | string
     | React.ComponentType<FieldProps<any>>
-    | React.ComponentType<void>;
+    | React.ComponentType;
 
   /**
-   * Component to render. Can either be a string like 'select' or a component.
+   * Component to render. Can either be a string e.g. 'select', 'input', or 'textarea', or a component.
    */
   as?:
-    | string
     | React.ComponentType<FieldProps<any>['field']>
-    | React.ComponentType<void>;
+    | keyof JSX.IntrinsicElements
+    | React.ComponentType;
 
   /**
    * Render prop (works like React router's <Route render={props =>} />)
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -25,10 +25,19 @@ export interface FieldProps<V = any> {
 export interface FieldConfig {
   /**
    * Field component to render. Can either be a string like 'select' or a component.
+   * @deprecated
    */
   component?:
     | string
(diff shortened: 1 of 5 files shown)


src/Field.tsx
```diff
@@ -128,20 +153,27 @@ export function Field({
...
+      return React.createElement(
+        component,
+        { ref: innerRef, ...field, ...rest },
+        children
+      );
+    }
+    return React.createElement(component, { ...bag, ...props }, children);
+  }
+
+  if (typeof is === 'string') {
```
nathanforce: Isn't this defaulted to `string` above on line 94?

```
as: is = 'input', // `as` is reserved in typescript lol
```

jaredpalmer: Yep this is desired because this will return `true` for `as="select"` but `false` for `as={StyleComponent}`

src/Field.tsx
```diff
@@ -92,17 +102,32 @@ export function Field({
...
+
+  warning(
+    !!component,
+    '<Field component> has been deprecated and will be removed in future versions of Formik. Use <Formik as> instead. Note that with the `as` prop, all props are passed directly through and not grouped in `field` object key.'
   );
 
   warning(
-    component && children && isFunction(children),
-    'You should not use <Field component> and <Field children> as a function in the same <Field> component; <Field component> will be ignored.'
+    !!is && !!children && isFunction(children),
```
hrgui: Since line 94 defaulted https://github.com/jaredpalmer/formik/pull/1406/files#diff-bc78961df770c0df68f6cd758a6f2afdR94 `as` to `input`, wouldn't this error message show up every time the dev does this?

```jsx
<Field name="name">
{({form, field}) => {
    return <MyCustomInput {...field} />
}}
</Field>
```

In the example above, as would default to `input`. In this case, would the check be something like `as !== 'input'`?

prajavk: @jaredpalmer  Why we Field render will be deprecated. 
Recently, we started using Field render in project. Formik 1.5.8
Please update API examples, if formik going to change ?

Issue #1148, opened 2018-12-03
Form is not dirtying when objects in the form change


jonathonwpowell:
## 🐛 Bug report

When a fields value is an object instead of an array / string, a change to the value will not dirty the form.  This causes the form to require another item to be changed for the form to dirty, and in the case where isInitialValid is false, for the form to even be valid.

### Current Behavior (in the demo)

Click the "Add my address" button
A new property should be added to testObj ( new value of myAddress: 222 Place)
The new property should be displayed below "Addresses"
The form should be dirtied and become valid.  The submit button should be clickable.

### Expected behavior

Clicked the "Add my address" button
A new property is added to testObj ( new value of myAddress: 222 Place)
The new property is displayed below "Addresses"
The form IS NOT dirtied and remains invalid. The submit button is not clickable.

### Reproducible example
https://codesandbox.io/s/w6z5r0nmp8

The contrived example is the user is on a form and clicks a button to add their address to the form.
I left in the e-mail section so that you can see how the form should dirty / become valid.

### Suggested solution(s)

The issue appears to be that when copying values to initial values (and back on reset), it is not being deep copied so the object is a reference copy.  This means when the value is changed, the saved initial value is changed, making it look like the form still isn't dirty when the initial values are compared to the current values to check if it is dirty.

### Additional context

I think the problem is manifesting on Formik.tsx:560, where the dirty value is calculated.  I was able to get it working as expected by adding a cloneDeep whenever value was set to initialValue or initialValue was set to value.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 1.3.1 |
| React            | 16.5.2 |
| TypeScript       | 3.1.3 |
| Browser          | Chrome 70.0.3538.110 |
| npm/Yarn         | npm 6.4.1 |
| Operating System | MacOS 10.14.1 |

zirho:
it seems using `setFieldValue` works just fine.
https://codesandbox.io/embed/625q5w7l0w

Cethy:
Had the same problem, and was using `setFieldValue`. 
The gotcha is using the same **reference** to the `values` object :

*not working*
```js
// directly updating the `values` object :
values.testObj.myAddress = "222 Place";
setFieldValue("testObj", values);
```
*working*
```js
const newValues = {...values, myAddress: "222 Place"};
// or 
const newValues = Object.assign({}, values); // clone values
newValues.myAddress = "222 Place";

setFieldValue("testObj", newValues);
```

[...]

Issue #1116, opened 2018-11-18
isValid is false although all fields are valid


burtek:
### Current Behavior

According to [validation reference FAQ](https://jaredpalmer.com/formik/docs/guides/validation) returning `undefined` as error message is considered as marking field valid. Yet returning `{ fieldName: undefined }` from validation function sets `isValid` to`false` thus preventing form from submitting. This is caused by following line in code:

https://github.com/jaredpalmer/formik/blob/85d15b302cd776eb615064a4451e1a7575b1439d/src/Formik.tsx#L425

### Expected behavior

Returning `{ fieldName: undefined }` from validation function should set `isValid` to true

### Reproducible example

https://codesandbox.io/s/py95r24n7x

Typing anything in the field changes form state, but keeps `isValid` set to `false`

If you change validate to `const validate = values => ({});`, `isValid` will be set to `true` 

### Suggested solution(s)

At least change the docs.
To make the validation work according to current docs, you'd need to check all values (incl. nested values) in object returned by validation function.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 1.3.2 |

motiazu:
The documentation is pretty weird around that line. [This](https://github.com/jaredpalmer/formik/blob/f58f85a6c229c1e4eb6e26b71af33726c0bfa036/src/Formik.tsx#L564) is the code that decides whether you have errors.
`Object.keys(this.state.errors).length` - this snippet pretty much means that any set field (even with undefined) will result in `isValid` being false.

I recommend clearing your own object of any undefined properties before returning the validation errors to formik. This could work before returning -
```
Object.keys(errors)
	.filter(key => errors[key] !== undefined)
	.reduce((res, key) => {
		res[key] = errors[key];
		return res;
	}, {})
```

webberwang:
I'm seeing this as well even after submitting a valid form, the `isValid` property stays false. It's causing some issue with validation states

barrett-vegas-com:
@webberwang I ran into the same issue, but then found https://jaredpalmer.com/formik/docs/api/formik#isinitialvalid-boolean

jp94:
@webberwang This seems to work for me.
Using @barrett-vegas-com response with `isInitialValid`,

```ts
const validate = (values: Values): FormikErrors<Values> => { ... };
const isInitialValid = (values: FormikConfig<Values> & IProps) =>
  Object.keys(validate(values.initialValues)).length === 0;

[...]

Issue #954, opened 2018-10-03
Form submission should not require a form to be dirty.


Alex-BHN:
## Current Behavior
Dirty is computed like this:

`const dirty = !isEqual(this.initialValues, this.state.values);`

Because of this and the decision to not allow not dirty forms to be submit, if a form isn't modified, it can't be submit.

## Steps to Reproduce
Don't modify a form and attempt to submit it.

## Expected behavior
An unmodified form should be able to be submit.

## Suggested solution(s)
A flag to allow this possibility or removal of the restriction.

## Additional context
Imagine a form with a single checkbox. The user should be able to submit this without modification.

**CodeSandbox Link**:

---

- Formik Version: 0.11.11
- React Version: 16.2.0
- TypeScript Version:
- Browser and Version: 
- OS: 
- Node Version:
- Package Manager and Version:

Saifadin:
`const dirty = !isEqual(this.initialValues, this.state.values);` is the definition of the dirty state.

But submitting a form without being dirty and touched is possible. There is no check for the dirty before an internal submit execution. 

In version `0.11.11` you can also pass `isInitialValid` to tell formik the form is valid.

prichodko:
@Saifadin is correct. Form is always submitted when there are no errors.

Issue #698, opened 2018-06-15
Custom isInitialValid should actually run errors so that they are visible


togakangaroo:
I have a long form that is persisted to the backend as the user works through it. This form has only a couple validations and when the user goes to the page, I need them to see any errors they might have so that they don't try to submit it.

It would seem to me that if isInitialValid is overridden, it would make sense to run `validateForm` to populate errors so that they're available. Its rather strange to have it not be valid but to have zero errors

Currently I am solving this by implementing

    isInitialValid: ({validate, data})=> !Object.keys(validate(data)).length,

and adding in a higher order component

```
const validateFormInitially = (FormComponent) =>
  class ValidateFormInitially extends Component {
    doCheck = () => {
      if(!this.props.isValid && !Object.keys(this.props.errors).length)
        this.props.validateForm(this.props.values)
    }
    componentDidUpdate = () => this.doCheck()
    constructor(props) {
      super(props)
      this.doCheck()
    }
    render = () => React.createElement(FormComponent, {...this.props})
  }
```

but that is both a pain, and the sort of behavior that I would expect from `isInitialValid`.

Additionally, I feel like `values` should be available in that function, but I don't feel super strongly about that.

mistadikay:
If validators are asynchronous, even that painful solution from @togakangaroo will not work. Would be nice to have an option to trigger validation on `componentDidMount` - whether it's synchronous or asynchronous.

mistadikay:
also, folks in #189 have a similar problem

mistadikay:
or even better: allow `isInitialValid` to return a Promise like `validate` does

togakangaroo:
I do think this hasn't been addressed and the point by @mistadikay is valid and maybe should become an up for grabs task?

DeClaessens:
Is there a certain issue to running `validateForm()` in `componentDidMount` that I am overlooking ? I came across the same issue and solved it that way.

keddisa:
If you're using Yup for validationSchema, Yup has it's own validation method that you can run in your lifecycle methods pre form render

Something like:
const isInitialFormValid = await validationSchema.isValid(initialValues)
(that returns a boolean)

and then you can assign isInitialFormValid to Formik's isInitialValid

<Formik
              initialValues={initialValues}
              validationSchema={validationSchema}
              isInitialValid={isInitialFormValid}
              .
              .
/>

PS: validationSchema.isValid() returns a promise, so you need to take that into consideration

PR #1410, opened 2019-04-02
Add initialXXXX and fix isValid

### LINKED ISSUE #1133: isValid behavior
### Current Behavior
When initializing a form with initialValues, `isValid` state is determined by `isInitialValid`.
If a user did not set isInitialValid, it will still be applied with `false`.
Even if the values have changed but still return (after multiple edits) back to the same `initialValues`, the form is deemed not dirty and the error object is ignored in deciding the validity.

### Expected behavior
If I decided not use `isInitialValid` explicitly, I expect the validity of the form to be decided by the `errors` object only.
If form is dirty or not, I want to `errors` to determine validity.

### Reproducible example

https://codesandbox.io/s/jn630ymxjv
Notice how the initial value of `id` is `GOOD_INPUT`, and how any value that isn't empty or `GOOD_INPUT` is accepted even though `GOOD_INPUT` is good input.

### Suggested solution(s)
I would suggest to not default `isInitialValid` to false, or add a prop to indicate whether I want it to determine my validation.
`isInitialValid` by itself is a pretty simple solution but it puts a lot more work on the dev. When using Formik I can just provide a Yup schema and the lib handles the rest, unless I set a possibly valid initial value ;(
In which case I will have to trigger my own validation, handle it's async nature by myself just to get the initialValue of something the lib gets right in any other case.

For our team I wrapped Formik and overriden the calculation of isValid, and triggered initial render on mount (we don't use SSR, I've seen this solution suggested in one of the issues). This helps avoid whatever boilerplate manual pre-validation we would have to do pretty much for every Formik use.
Edit: Since writing this I managed to create a better solution [here](https://github.com/jaredpalmer/formik/issues/1133#issuecomment-472760068)
```
/**
 * WrappedFormik was created for two reasons:
 * 1. Formik doesn't validate initialValues, so we trigger a validation on mount.
 * 2. Formik calculates isValid by whether there are any errors (good) or by a prop named
 * isInitialValid. This can be found in their github under Formik.tsx in
 * the getFormikComputedProps method. Since WE don't use isInitialValid, this logic assumes all
 * our initialValues are invalid by default. So we override their isValid calculation to
 * only account for found errors.
 */
const RenderWrapper = compose(
  lifecycle({
    componentDidMount() {
      this.props.data.validateForm();
    },
  }),
  withProps(props => {
    const errors = props.data.errors;
    return { isValid: Object.keys(errors).length === 0 };
  }),
)(({ render, data, isValid }) => render({ ...data, isValid }));
export const WrappedFormik = ({ render, ...props }) => (
  <Formik {...props} render={data => <RenderWrapper data={data} render={render} />} />
);
```

### Your environment

| Formik           | 1.3.1
| React            | 15.6.2
| TypeScript       | -
| Browser          | Chrome
| npm/Yarn         | Both
| Operating System | OSX

jaredpalmer:
- Deprecate `isInitialValid` with a warning. If it is specified, old 1.x `isValid` behavior is utilized.
- `isValid` is now `true` by default.
- Add `initialErrors` and `initialTouched` props to `<Formik>`
- Add `mapPropsToErrors` and `mapPropsToTouched` to `withFormik()`
- **Breaking**: `resetForm(nextValues: Values)` to `resetForm(nextInitialState: FormikState<Values>`

I'm not convinced we actually need to change `resetForm` (and thus keep all of these as refs that can be mutated on reset)

FWIW I can't think of when you would want `initialTouched`. Open to ideas and would be happy to remove.

Closes #1133 #1116 #1148 #288 #954 #698

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -107,19 +107,17 @@ Submitting state of the form. Returns `true` if submission is in progress and `f
 
 #### `isValid: boolean`
 
-Returns `true` if the there are no `errors`, or the result of
-`isInitialValid` the form if is in "pristine" condition (i.e. not `dirty`).
+Returns `true` if there are no `errors` (i.e. the `errors` object is empty) and `false` otherwise.
+
+> Note: `isInitialValid` was deprecated in 2.x. However, for backwards compatibility, if the `isInitialValid` prop is specified, `isValid` will return `true` if the there are no `errors`, or the result of `isInitialValid` of the form if it is in "pristine" condition (i.e. not `dirty`).
 
 #### `isValidating: boolean`
 
 Returns `true` if Formik is running validation during submission, or by calling [`validateForm`] directly `false` otherwise. To learn more about what happens with `isValidating` during the submission process, see [Form Submission](guides/form-submission.md).
 
-#### `resetForm: (nextValues?: Values) => void`
+#### `resetForm: (nextInitialState?: FormikState<Values>) => void`
 
-Imperatively reset the form. This will clear `errors` and `touched`, set
-`isSubmitting` to `false`, `isValidating` to `false`, and rerun `mapPropsToValues` with the current
-`WrappedComponent`'s `props` or what's passed as an argument. The latter is
-useful for calling `resetForm` within `componentWillReceiveProps`.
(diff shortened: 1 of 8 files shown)


jaredpalmer:
Does anyone actually need `initialTouched`?

ohmonster:
I *think* I need initial touched. We have a use case where users may have address data saved on the server that doesn't meet our current validation field criteria --generally missing a phone number or something. Our error display rule triggers if the field is touched. So we want the form to be as populated as it can be, validated on mount, and touched on mount.

I was using a ref to do this before, but the upgrade to Formik 2.0 gave me errors about innerRefs.

PR #1423, opened 2019-04-09
Updating documentation of submitForm

FBerthelot:
Hello,

A little first PR :)

Updating documentions according to #486.

Have a nice day

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -131,9 +131,9 @@ Set the error message of a field imperatively. `field` should match the key of
 Set the touched state of a field imperatively. `field` should match the key of
 `touched` you wish to update. Useful for creating custom input blur handlers. Calling this method will trigger validation to run if `validateOnBlur` is set to `true` (which it is by default). `isTouched` defaults to `true` if not specified. You can also explicitly prevent/skip validation by passing a third argument as `false`.
 
-#### `submitForm: () => void`
+#### `submitForm: () => Promise`
 
-Trigger a form submission.
+Trigger a form submission. The promise will be rejected if form is invalid. 
 
 #### `submitCount: number`
 


hinok:
@jaredpalmer @FBerthelot 

At the moment documentation says that `submitForm` returns a rejected promise if form is invalid but it looks that it "eats" errors and only set state `this.setState({ isSubmitting: false });`.

**Lines:** https://github.com/jaredpalmer/formik/blob/v1.5.4/src/Formik.tsx#L426-L450

This misleading information shows up on

```
- v1.5.4 (https://jaredpalmer.com/formik/docs/api/formik#submitform-promise)
- v1.5.3 (https://jaredpalmer.com/formik/docs/1.5.3/api/formik#submitform-promise)
```

v1.5.2 looks okay (https://jaredpalmer.com/formik/docs/1.5.2/api/formik#submitform-void), no information about returning promise.

Was that PR merged too early?

FBerthelot:
Hi @hinok ,

Submit form return the promise from runValidations. This Promise is here => https://github.com/jaredpalmer/formik/blob/v1.5.4/src/Formik.tsx#L267

I believed the  RunValidation reject promise when the form is invalid, it's make sense.

Sorry If I made a mistake. I was certain at the time.

hinok:
It returns a promise but it's always resolved https://github.com/jaredpalmer/formik/blob/v1.5.4/src/Formik.tsx#L299

so something like

```js
submitForm()
  then(() => {
    // this is always fired, it doesn't matter if form is valid or not
  })
  .catch((errors) => {
    // this will never be fired
  });
```

ivan-oliveira:
@FBerthelot 

Looking at the code for runValidations it will resolve the promise with the errors: 
`return promise
      .then((errors: FormikErrors<Values>) => {
        if (this.didMount) {
          this.setState(prevState => {
            if (!isEqual(prevState.errors, errors)) {
              return { errors };
            }
            return null; // abort the update
          });
        }
        return errors;
      })
      .catch(x => x);
`

In turn submitForm() will always resolve as well.

FBerthelot:
@ivan-oliveira look at #1580 ;)

Issue #1431, opened 2019-04-11
run Storybook is impossible due to at-loader configuration


andrewMuntyan:
### Current Behavior
Storybook is built, up and running on `localhost:9001`, but is unusable

![Storybook 2019-04-11 22-53-07](https://user-images.githubusercontent.com/1164826/55988484-a8584c00-5cac-11e9-8616-dcb2ab372825.png)

### Expected behavior

Storybook is up and running on `localhost:9001` without rendered at-loader errors

### Reproducible example

_Steps_:
* cd formik
* yarn 
* yarn storybook

_Result_:
```
ERROR in [at-loader] ./test/Field.test.tsx:43:41
    TS2314: Generic type 'FastFieldConfig<T>' requires 1 type argument(s).
```
and 8 more at-loader errors in terminal console.

### Suggested solution(s)

Provide additional configuration to at-loader to avoid loading `test` dir

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 1.5.2
| TypeScript       | 3.1.6
| Browser          | Google Chrome. 73.0.3683.86 (Official Build) (64-bit)
| npm/Yarn         | 1.10.1
| Operating System | macOs High Sierra

andrewMuntyan:
Shame on me. I've misspelled filename for config. I've put `storybook.tsconfig.json` instead of `tsconfig.storybook.json`. Will do one more PR

PR #1432, opened 2019-04-11
bugfix Storybook ts config added bugfix/jaredpalmer/formik#1431

andrewMuntyan:
this is for https://github.com/jaredpalmer/formik/issues/1431 `run Storybook is impossible due to at-loader configuration`


jaredpalmer:
Nice.

PR #1445, opened 2019-04-17
Expose form submission errors

Granipouss:
Hi people,

I have been super excited about formik hooks, it's awesome. But I also noticed that I was often using a pattern that I think can be part of the default formik API.

I very often write code that looks like that to handle errors:
```js
const [error, setError] = useState<Error | null>(null);
const { /* ... */ } = useFormik({
  // ...
  onSubmit: async values => {
    try {
      await doSomeAsyncStuff(values) // Mostly API calls
    } catch (err) {
      setError(err)
    }
  },
});
```

I am pretty sure there could be a better way.

I just provided a quick POC, but it is really a quick draft to start the conversation.

Cheers,

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -28,7 +28,7 @@ import invariant from 'tiny-warning';
 // We already used FormikActions. So we'll go all Elm-y, and use Message.
 type FormikMessage<Values> =
   | { type: 'SUBMIT_ATTEMPT' }
-  | { type: 'SUBMIT_FAILURE' }
+  | { type: 'SUBMIT_FAILURE'; payload: any }
   | { type: 'SUBMIT_SUCCESS' }
   | { type: 'SET_ISVALIDATING'; payload: boolean }
   | { type: 'SET_ISSUBMITTING'; payload: boolean }
@@ -92,6 +92,7 @@ function formikReducer<Values>(
       return {
         ...state,
         isSubmitting: false,
+        formError: msg.payload,
       };
     case 'SUBMIT_SUCCESS':
       return {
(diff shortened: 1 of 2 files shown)


Satyam:
I like your solution.

I think that separating `formError` from the `errors` object (i.e.: adding `errors._form = "some submit error`) is wise, as `isValid` would never be true thereafter since no change on any field would clear an error that doesn't belong to any named field, and the form would never be able to re-submit.

I was wondering if a property like `onSubmitResolved` could be added, which would receive a function with `formError` and `values` as its only arguments (a signature not unlike many NodeJS async functions).  It would be called after the actions for `SUBMIT_SUCCESS` or `SUBMIT_FAILURE` have been processed and no further changes of state would happen.   Separate events for `onSubmitSuccess` and `onSubmitFailure` would be just as good.

It often happens that you want to switch to another view or navigate away from a form once it has been submitted and, by lack of any alternative, you append it to the async action in `onSubmit`, something like:

```jsx
onSubmit={(formikbag) => fetch(....)
    .then((result) => {
         // .....
         history.goBack();
     })
}
```
The problem with this code is that Formik will do things when the Promise returned by `fetch` is returned from `onSubmit` but, since part of that Promise chain is navigating away (`history.goBack()`), hence unmounting, the form still has things to do with the state, which gets React to complain that you can't call `setState` on an unmounted component.

The reason for having so few arguments passed to the function is that you shouldn't be doing anything else with the form at this point, no calling any `setXxxx` method, and most of the state of the form should be irrelevant by then.  

Additionally, most of the checks for `isMounted.current` might go away if developers used `onSubmitResolved` to do whatever action gets the form unmounted.  This check had to be there because Formik didn't give any clue when it was safe for the form to be unmounted so you just have to risk it.  With a clear signal on when it is safe, any remaining errors on the forms using Formik, which are currenty masked,  should really be dealt with.

Thanks

Issue #1467, opened 2019-04-26
Warn consumers if Formik context is undefined


jaredpalmer:
Continuation of #1368

gaspardip:
Would this be enough?

```
  const C: React.SFC<OuterProps> = (props: OuterProps) => (
    <FormikConsumer>
      {formik => {
        warning(
          isEmpty(formik),
          'Formik context is undefined, please verify you are rendering <Field>, <FastField> or <FieldArray> inside <Formik>'
        );

        return <Comp {...props} formik={formik} />;
      }}
    </FormikConsumer>
  );
```

zsid:
How can I test that field outside formik? We create components that we use in multiple formik forms and rather than creating them every time we have abstracted them on their own.

So how can I have unit tests for those fields since I am getting the errors:
```
Warning: Formik context is undefined, please verify you are rendering <Form>, <Field>, <FastField>, <FieldArray>, or your custom context-using component as a child of a <Formik> component. Component name: FieldArrayInner
```

```
    TypeError: Cannot read property 'touched' of undefined
```

cinemanja:
@zsid Did you figure out on how to do this?

zsid:
@cinemanja import `Formik` and use it as normally by providing the form etc. I am also mounting it.

Issue #1368, opened 2019-03-06
React 16.8 | this.props.formik.registerField is not a function


KevinMongiello:
After upgrading from React 16.0.0 to 16.3.0, I'm getting the following breaking changes when trying to render my `<Formik><Form /></Formik>` component.

```
Uncaught TypeError: this.props.formik.registerField is not a function
 at FieldInner.componentDidMount
```

Formik Version: 1.5.1
React Version: >= 16.3.0

KevinMongiello:
I found a work-around.

By using [connect()](https://jaredpalmer.com/formik/docs/api/connect) on my Field components, they were able to receive the required props.

pvijeh:
seeing this too with react v 16.5.1

amelieMa:
same with react v16.8.4

russpitre:
Not sure if this help you all but I was seeing the same error. I noticed i had an errant import statement (es5) like this:  

```
import {
  Form,
  Formik
} from 'formik/dist/index';
```

I changed it to and now all is well:

```
import {
  Form,
  Formik
} from 'formik'
```

Hope that helps.

ColadaFF:
I run in the same issue and solved it by moving the `Field` component as a direct child of the `Formik` component.

mikemclin:
Updating to Formik `1.5.2` fixed this for me

phil-lgr:
Still have the issue with `1.5.2`

vget98:
Any update with this issue?

KevinMongiello:
For whatever reason it started working for me.
i didn't change a thing except the react packages.

Formik: 1.51
react: 16.8.6
react-dom: 16.8.6

gaspardip:
Having the same problem on `1.5.2`, any workarounds?

jaredpalmer:
Can you share code that is problematic? I have tested with latest versions of React and everything works fine. That error can only happen if your Field is not being rendered underneath a `<Formik>` component.

jaredpalmer:
My guess is that it has to do with the way that `connect` is including the types for create-react-context. I believe that `import<>` could be rpoblematic for those on a typescript version less that 2.9.

```
import * as React from 'react';
import { FormikContext } from './types';
export declare const FormikProvider: React.ComponentClass<import("create-react-context").ProviderProps<FormikContext<any>>, any>, FormikConsumer: React.ComponentClass<import("create-react-context").ConsumerProps<FormikContext<any>>, any>;
[...]

PR #1470, opened 2019-04-27
Add warning when formik ctx is empty in connect

gaspardip:
Related to #1368 and then #1467, I added the warn when `formik` is empty in `connect`. Also added some checks on `Field` to avoid crashes, but maybe we should let it crash and warn, not one or another.


gaspardip:
Made the requested changes!

PR #1483, opened 2019-05-03
Rerender FastField when non-formik context props change

### LINKED ISSUE #1188: Rerender FastField when non-formik context props change
### Current Behavior

Non-formik context props only trigger a `FastField` rerender if they are [added or removed](https://github.com/jaredpalmer/formik/blob/master/src/FastField.tsx#L120).

### Desired Behavior

Allow `FastField` to rerender when non-formik context props change value. This doesn't strike me as something that is likely to produce a performance hit.

### Suggested Solution

Do a shallow comparison of non-formik context props in `FastField`'s `shouldComponentUpdate`

All the people!

martingtheodo:
Resolves #1188

Hello everyone,

Thank you for this project, really useful and easy to use.

We had the same problem as in this issue, so I proposed this PR to implement the suggested solution.

I followed the doc with yarn link and tested it on my project. It worked fine.
I chose react-fast-compare as I saw it was already used elsewhere in formik.

Hope this helps and I welcome any feedback you could have.

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import isEqual from 'react-fast-compare';
 
 import { connect } from './connect';
 import {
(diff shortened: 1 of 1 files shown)


martingtheodo:
Ping @jaredpalmer I implemented this feature request that you said it seemed a good idea.
Do you think you could have time to review it ?
Thanks in advance

chrisswordfish:
Keen to see this implemented! My use case is where the label needs to change using react-intl. It wasn't a problem when the component was wrapped in a HOC, but now I've converted to the useIntl hook and re-renders are not being triggered when the locale changes.

PR #1490, opened 2019-05-08
Provide better typing for isInitialValid method when using withFormik

martynsmith:
We want to use the props passed to our form to determine if the initial state should be valid.

At the moment because the signature is `(props: object) => boolean | undefined` we can't access our props to calculate the boolean value.

This patch is a minimal change to make `FormikSharedConfig` a generic with a default that matches the existing behaviour. We then alter `WithFormikConfig` to pass its `Props` through when extending it.

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -155,13 +155,13 @@ export interface FormikHandlers {
 /**
  * Base formik configuration/props shared between the HoC and Component.
  */
-export interface FormikSharedConfig {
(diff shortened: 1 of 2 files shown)


Andreyco:
This looks good.
cc @jaredpalmer

jaredpalmer:
We are deprecating this, but sure.

Andreyco:
Lol, sorry :)

PR #1494, opened 2019-05-08
[v2]:sparkles: field level initial value

kimochg:
### What

make field-level initial value

### Why

it provides more flexible ways to set the initial values from the field-level

### Done

- [x] initial values impl
- [x] update example in the storybook
- [x] docs updated
- [x] pass test
  I tried to make this pr pass test in `hooks` branch. but it seems the code doesn't 100% pass test in `origin/hooks`. I think the related failure causing by this pr has been resolved.


FredyC:
You might perhaps elaborate why do you think it's "flexible". I used to be declaring initial value per field too (with other form libraries), but on the contrary, it got fairly messy. It's so much easier to initials for the whole format one place.

Besides if you really need to change "initial" value of a specific field, it should be as easy as this.

```js
const formik = useFormikContext()
React.useEffect(() => {
  formik.setValue("field", value)
}, [])
```

You can make a custom hook of it to make it a one-liner.

jaredpalmer:
@kimochg Can you demonstrate a use-case where this will be used? I cannot think of one off the top of my head.

kimochg:
@FredyC Thanks for your reply. But `setFieldValue` make the form `dirty`, while the `initialValues` not. This is the main reason that the `initialValues` should be used in my case.

@jaredpalmer In our project, the process of rendering form is totally dynamic, we fetch the meta information from backend, and turn that info to React component. Things become complex when "subform" is introduced, (a subform is a form inside a top form, so it renders as FieldArray in Formik). The initialValues comes from backend meta information as well. So, in this scenario, if I can set the initialValue in field-level, this might be a more flexible way to implement our system. Right now, I did all kind of stuff in top-level formik, and it works well. If you think, the field should not set initial value, You can close this PR. Thanks

FredyC:
@kimochg And why is such a problem to dynamically create initialValues for a whole form? If you can generate dynamically the whole form, it would make sense to have initial values on the same level nonetheless.

kimochg:
@FredyC I agree with your concern about making things worse. It's more like a choice to manage the logic inside `formik` or lift this part to developers.

jaredpalmer:
I don't think that fields should set initial values. Going to close.

PR #1502, opened 2019-05-09
Upgrade hoist-non-react-statics to v3.3.0

matomesc:
This pull request is meant to fix issues arising from older versions of `hoist-non-react-statics` hoisting the `contextType` static property when using the `connect()` or `withFormik()` HOCs:

```
Warning: FormikConnect(ComponentInner): Function components do not support contextType.
```

This issue was fixed in `hoist-non-react-statics@3.1.0` in https://github.com/mridgway/hoist-non-react-statics/pull/62

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "dependencies": {
     "create-react-context": "^0.2.2",
     "deepmerge": "^2.1.1",
-    "hoist-non-react-statics": "^2.5.5",
+    "hoist-non-react-statics": "^3.3.0",
     "lodash": "^4.17.11",
     "lodash-es": "^4.17.11",
     "prop-types": "^15.6.1",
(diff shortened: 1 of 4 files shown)


bopfer:
I am getting this error when upgrading to the latest alpha (2.0.1-alpha.51):

```
node_modules/formik/dist/ErrorMessage.d.ts:1:23 - error TS2688: Cannot find type definition file for 'hoist-non-react-statics'.

1 /// <reference types="hoist-non-react-statics" />

node_modules/formik/dist/ErrorMessage.d.ts:11:88 - error TS7016: Could not find a declaration file for module 'hoist-non-react-statics'. '/myproject/node_modules/hoist-non-react-statics/dist/hoist-non-react-statics.cjs.js' implicitly has an 'any' type.
   Try `npm install @types/hoist-non-react-statics` if it exists or add a new declaration (.d.ts) file containing `declare module 'hoist-non-react-statics';`

11 export declare const ErrorMessage: React.FunctionComponent<ErrorMessageProps> & import("hoist-non-react-statics").NonReactStatics<React.ComponentClass<ErrorMessageProps & {
```

Is that something I would need to install separately from Formik? or is it something that needs to be added to Formik's dependencies?

jaredpalmer:
Will investigate

PR #1508, opened 2019-05-13
Add `validationContext` prop and pass to `validationSchema`

DaddyWarbucks:
This PR attempts to address the requested feature here: https://github.com/jaredpalmer/formik/issues/503 of passing some prop to the underlying Yup schema's `context` option.

I started to make `validationContext` more similar to `validationSchema` in that the prop could either be the object or a function that returns the object, similar to how `validationSchema` is either a Yup schema or a function that returns a yup schema. But, this required quite a bit more work to execute that function (if it is a function), rather than just assuming it is the context object already. It would also require stripping that function off as I have seen done with other props that are functions, such as `examples/Debug` and other places. I am willing to update this, but I wanted to get your temperature on the changes before diving into that. It seems like making that change to allowing a function increases the footprint of this PR quite a bit.

I am also not great with TS or jest, so I suspect you will have some input ons ome of those types of things as well.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,28 +1,28 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 144255,
-    "minified": 39643,
-    "gzipped": 12000
+    "bundled": 144485,
+    "minified": 39756,
+    "gzipped": 12026
   },
   "./dist/formik.umd.development.js": {
-    "bundled": 174435,
-    "minified": 50097,
-    "gzipped": 15119
+    "bundled": 174665,
+    "minified": 50210,
+    "gzipped": 15144
   },
   "./dist/formik.cjs.production.js": {
-    "bundled": 44141,
-    "minified": 22101,
-    "gzipped": 5521
+    "bundled": 44638,
+    "minified": 22214,
+    "gzipped": 5548
   },
   "./dist/formik.cjs.development.js": {
-    "bundled": 45029,
-    "minified": 23001,
-    "gzipped": 5862
+    "bundled": 45526,
+    "minified": 23114,
+    "gzipped": 5888
   },
   "dist/formik.esm.js": {
-    "bundled": 40594,
-    "minified": 22365,
-    "gzipped": 5736,
+    "bundled": 40808,
+    "minified": 22478,
+    "gzipped": 5763,
     "treeshaked": {
       "rollup": {
         "code": 802,
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -133,7 +133,7 @@ Set the touched state of a field imperatively. `field` should match the key of
 
 #### `submitForm: () => Promise`
 
-Trigger a form submission. The promise will be rejected if form is invalid. 
+Trigger a form submission. The promise will be rejected if form is invalid.
 
 #### `submitCount: number`
 
@@ -380,3 +380,7 @@ events and `change`-related methods. More specifically, when either
 [A Yup schema](https://github.com/jquense/yup) or a function that returns a Yup
 schema. This is used for validation. Errors are mapped by key to the inner
 component's `errors`. Its keys should match those of `values`.
+
+### `validationContext?: any`
+
+[A Yup context object](https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror). Pass an arbitrary set of key value pairs which will be accesible to the `validationSchema` or `validate` props.
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -246,11 +246,11 @@ export class Formik<Values = FormikValues> extends React.Component<
    */
   runValidationSchema = (values: FormikValues) => {
     return new Promise(resolve => {
-      const { validationSchema } = this.props;
+      const { validationSchema, validationContext } = this.props;
       const schema = isFunction(validationSchema)
         ? validationSchema()
         : validationSchema;
-      validateYupSchema(values, schema).then(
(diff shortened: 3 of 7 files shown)


DaddyWarbucks:
I had not noticed the work here: https://github.com/jaredpalmer/formik/pull/522. But it appears my work and the work done on that PR are quite similar

andercard:
Ping

nutboltu:
Any updates?

jaredpalmer:
I don't think this is necessary as you can accomplish this via function as described in this comment. 

https://github.com/jaredpalmer/formik/issues/503#issuecomment-467020256

DaddyWarbucks:
While I understand and appreciate @jaredpalmer being cautious of extending the API, I had hoped that because Formik is already so heavily influenced by Yup that we could use the two together without issue OTB.  I think this [comment](https://github.com/jaredpalmer/formik/pull/522#issuecomment-378387180) summarizes the lack of this feature and its workarounds well:
 > "but it is cumbersome and you need to know intricaties of Formik internals to know how to use throw with yupToFormErrors helper"

But I do truly understand not wanting to extend the API, especially on a project so widely used.

With that said, the solution provided by that comment is the best I have seen as well. With a little inspiration from that code there, I think a good solution would be as follows: 

```js
import { Formik, yupToFormErrors } from 'formik';

export default props => {
  return (
    <Formik
      validate={(values, props) => {
        return props.validationSchema
          .validate(values, { abortEarly: false, context: props.context })
          .catch(err => {
            throw yupToFormErrors(err);
          });
      }}
      {...props}
    />
  );
};
```

I rarely use the rest of Formik's components directly from Formik itself. I almost always create my own  `<Input />` component that builds on top of `<Field />` with my classes, logic, etc. So why not do the same for the `Formik` component itself.  Just solve it with simple React.

I do think this deserves its own place in the documentation or examples. This feature seems to be highly requested so I think we should attempt to publish this, or some other, solution for people that want it.

nutboltu:
@DaddyWarbucks totally appreciated. 

I come up with writing `customHandleChange` as below:
```js
export const customHandleChange = async (event, props, schema, context) => {
  const { values, setFieldValue, setFieldError, setFieldTouched } = props;
  const { name, value } = event.target;
  try {
    setFieldValue(name, value, false);
    setFieldTouched(name, true, false);

    const rootValues = set(values, name, value);
    await schema.validateAt(name, rootValues, { context });
    setFieldError(name, null);
  } catch (ex) {
    const [firstError] = ex.errors;
    setFieldError(name, firstError);
  }
};
```

```jsx
import { Formik } from 'formik';
import schema from './schema';

export default props => {
  return (
    <Formik
      validationSchema ={schema}
      render={props => (
        <form onSubmit={props.handleSubmit}>
          <input
            type="text"
            onChange={(event) =>
                customHandleChange(event, props, schema, { test: 'this is context'})}
            name="name"
          />
        </form>
      )}
    />
  );
};

```

@jaredpalmer Can we somehow documented this with example? As this feature seems to be highly requested, it would be nice to have documented possible approaches to tackle validation with context.

kevindkim:
@DaddyWarbucks would you mind walking through your approach a bit more - particularly what's happening when you pass down 'context'?

DaddyWarbucks:
Hey @kevindkim,

`context` is a property that can be passed within Yup's options. It is just an arbitrary object of whatever you want to pass to it. This `context` can then be accessed within a number of Yup's methods to create custom validations. 

See this link for more: https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror

For example, the below is a little custom method I use to make a property required or not based on the CRUD method being used.

```js
  yup.addMethod(yup.mixed, 'requiredWhen', function(_requires = []) {
    const requires = Array.isArray(_requires) ? _requires : [_requires];
    return this.test({
      name: 'requiredWhen',
      exclusive: true,
      message: '${label} is a required field',
      test: function(value) {
        const { context } = this.options;
        if (!context || !context.method) {
          throw new GeneralError(
            `The "requiredWhen" validation method can only be used when "validate" is called with options.context.method`
          );
        }
        if (requires.includes(context.method)) {
          // TODO: Check against more falsey values?
          return value !== undefined && value !== null;
        } else {
          // This field is not required on this method
          return true;
        }
      }
    });
  });
```

Now I can do something like

```js
const mySchema = yup.object({
  name: yup.string().required(), // always required
  age: yup.number().requiredWhen(['update']) // only required if updating, not creating
});

mySchema.validate({ name: 'DaddyWarbucks' }, { context: { method: 'create' } }) // should pass
mySchema.validate({ name: 'DaddyWarbucks' }, { context: { method: 'update' } }) // should fail
```

silasdavis:
> I don't think this is necessary as you can accomplish this via function as described in this comment.
> 
> [#503 (comment)](https://github.com/jaredpalmer/formik/issues/503#issuecomment-467020256)

This means you recreate the schema object on every redraw - not a great solution in general...

StudioSpindle:
Like the approach of @DaddyWarbucks.

However when I implement it the validation gives unexpected results. It seems to be returning the value in the field as the error message strangely. However, when I log the result of the promise and do nothing with it then it behaves correct: 

```
validate={(values) => {
  console.log('values: ', values);
  return schema.validate(values, { abortEarly: true, context: { "foo": "bar" }})
    .then((foo) => {
      // do nothing with this...
      console.log('thing: ', foo);
    })
    .catch((errors) => {
      console.log('errors', errors);
    });
```

It seems that the `runValidationSchema` is the way Formik does this internally: https://github.com/jaredpalmer/formik/blob/master/packages/formik/src/Formik.tsx#L218

Anybody else having this experience?

Issue #1388, opened 2019-03-19
Type-safe implementation of setFieldValue


samstarling:
### Current Behavior

At the moment, to set a field value, you have to call a method with the following signature:

```
setFieldValue(field: string, value: any)
```

When using Formik with TypeScript, this method isn't "safe": you could specify a field that doesn't exist, or specify a value that doesn't confirm to the type signature.

### Desired Behavior

A setFieldValue (or similar) method that takes a "partial" object would allow for writing a TypeScript definition similar to `setState`, for example:

```
setFieldValue({ name: "Sam", zipcode: "12345" })
```

And a definition like (pseudocode):

```
setFieldValue<K extends keyof Values>(value: (Pick<Values, K> | Values | null)): void;
```

### Suggested Solution

I know that TypeScript isn't given first-class support by the maintainers, but I'm interested in knowing your thoughts on such a `setFieldValue` method: I would be happy to raise a PR with this change and write the associated TypeScript definitions if you think this is a good idea. 

### Who does this impact? Who is this for?

TypeScript users.

### Describe alternatives you've considered

I couldn't think of any, but I'd be interested to know if there's an alternative.

justinbhopper:
That new signature could work nicely, but I suggest it just be defined as another callback, aptly named `setFieldValues` (plural).  

The current method of `setFieldValue` needs its signature fixed.  The field parameter should be restricted so its only a key of the `Values` generic.  I'm not sure why its just plainly defined as a `string`, but its concerning to me.

```typescript
setFieldValue(field: keyof Values, value: 
```

samstarling:
@justinbhopper That makes sense: and you're right, `setFieldValues` is a much better name. I might give this a shot, if people feel like it's a sensible idea.

justinbhopper:
I certainly would use it.  I think this would be the signature you would use:

```typescript
export interface FormikActions<Values> {
  setFieldValues(values: Partial<Values>): void;
}
```

You'd roll through each property in `values` one by one, probably using a `for in`.  You should accept the value, even if its `undefined` -- don't make the mistake of skipping/ignoring undefined values, because currently formik allows setting a field's value to undefined in `setFieldValue`.

kevinmook:
We ran into this in a project we're working on as well. I think this signature might work for `setFieldValue` (making the "value" typesafe as well):

```
export interface FormikActions<Values> {
  setFieldValue<Field extends keyof Values>(
    field: Field,
    value: Values[Field],
[...]

PR #1513, opened 2019-05-16
Extend setValue functionality

samstarling:
Formik is great, but there are one or two parts where the methods could be a little more TypeScript friendly, and in this PR I'm introducing a new method which I hope takes a step in this direction.

For context: I have a form where a user can enter an ID into a field to perform a lookup. Data is fetched over HTTP, and I use data from the response to populate other form fields by using multiple `setFieldValue` calls. The signature of `setFieldValue` isn't type-safe – the `value` is `any`, and the `field` may not necessarily exist. (There was some discussion of this in #1388)

This PR introduces a `setFieldValues` method that takes a `Partial<Values>` and merges that with the existing form values. I would find this really useful.

Things that I'm not too sure about, and that I'd welcome feedback on:

* Does this method go against the design of Formik, where fields are referenced using the path-style string (eg. `people.0.first_name`)?
* Is there a better name for this method, than `setFieldValues`? It's basically like `setValues`, but it allows subsets and merges rather than overwriting.


test/Formik.test.tsx
```diff
@@ -44,29 +45,36 @@ function Form({
...
 
-const InitialValues: Values = { name: 'jared' };
-
-function renderFormik<V>(props?: Partial<FormikConfig<V>>) {
+const InitialValues: Values = {
+  name: 'jared',
+  zipcode: 12345,
+};
+
+function renderFormik<V = Values>(
```
samstarling: I had to add an explicit type definition for the return type here, in order for `getProps()` not to be inferred as returning `any`

docs/api/formik.md
```diff
@@ -145,6 +145,10 @@ Number of times user tried to submit the form. Increases when [`handleSubmit`](#
 Set the value of a field imperatively. `field` should match the key of
 `values` you wish to update. Useful for creating custom input change handlers. Calling this will trigger validation to run if `validateOnChange` is set to `true` (which it is by default). You can also explicitly prevent/skip validation by passing a third argument as `false`.
 
+#### `setFieldValues: (values: Partial<Values>, shouldValidate?: boolean) => void`
```
samstarling: Please let me know if there are other sections of documentation I should be updating as well as this!

samstarling:
@jaredpalmer If you had any ideas on why `yarn test` and `yarn build` succeed locally, but not on Circle CI, they would be much appreciated!

johnrom:
@samstarling it looks like the specific version of TypeScript that is used in the build requires an explicit `FormikConfig<V extends object>` in order to use ` { ...values }`. I also had this issue over in #1334, and I fixed it by upgrading typescript to 3.3.x in 3e1d0d9 

> Is there a better name for this method, than setFieldValues

I'm a bit confused as to why we need a new function instead of modifying the behavior of setValues to accept a partial state, and an optional shouldValidate parameter. It seems to do the exact same thing as setValues. `FormikState<Values>['values']` is the same thing as `Values`, but in the future there may be different functionality needed for the formikState values property, so I'd recommend keeping it written the other way. For example, a future version of Formik might not require initialState, and FormikState.values might be `Values | undefined`.

samstarling:
@johnrom Thanks for pointing that out – really useful, and I'll copy across that 3.3.x upgrade. Regarding the difference with `setValues` is that it calls `this.setState({ values })`, which means that `values` is completely replaced with what's passed in (as far as I understand). The method added in this PR just performs a partial update of `values`.

johnrom:
Oh yes, duh. Hope the TS fix helps!

samstarling:
@johnrom It worked: thanks!

Andreyco:
How about this signature?

```ts
setValues(values: Values | (values: Values) => Values)
```

Basicaly, you pass setter funcion and you decide whether to merge or replace values. No breaking change! Cc @jaredpalmer

jaredpalmer:
@andreyco agreed this is a wart.

jaredpalmer:
Oh I’m sorry. Why do we need this? We already have setValues?

samstarling:
@jaredpalmer As far as I understand it, `setValues` replaces `values` entirely. @Andreyco: Do you mean that in the signature `(values: Values) => Values`, the property passed in would be the existing `values`?

johnrom:
I like the setValues(Values | ReplaceValuesCallback) API for back compat. If someone hackishly found a use case for passing a function as Values their code would break, but I'd assume that would have broken Formik itself long before finding a use case. I also think it's reasonable to want to skip validation during this call.

Andreyco:
I am against introducing new API for this.

To explain my previous comment with code...
```js
// Now, Formik's setValues replaces values altogether. Merging is "hard", you need to pass all values
formik.setValues({ anotherField: 'value' }) // -> bad, original values are lost
formik.setValues({ ...allValues, anotherField: 'value' }) // -> good, but... passing values is required

// With setter callback, values could be passed by Formik
formik.setValues(values => ({ ...values, anotherField: 'value' })) // -> good, values are passed. You decide whether to merge or replace
```
No new API, backward compatible.

samstarling:
@johnrom @Andreyco @jaredpalmer  Thanks to you both for your feedback – I've just pushed a commit which implements your suggestion. Could you take a look and let me know what you think?

test/Formik.test.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { render, fireEvent, wait } from 'react-testing-library';
 import * as Yup from 'yup';
 
-import { Formik, FormikProps, FormikConfig } from '../src';
+import { Formik, FormikProps, FormikConfig, ValueUpdate } from '../src';
```
johnrom: doesn't look like ValueUpdate is needed here

test/Formik.test.tsx
```diff
@@ -542,6 +542,17 @@ describe('<Formik>', () => {
         expect(validate).not.toHaveBeenCalled();
       });
 
+      it('setValues takes a function which can patch values', () => {
+        const { getProps } = renderFormik();
+
+        getProps().setValues((values: Values) => ({
```
johnrom: is the Type of `values` inferrable here? if so, avoid manually typing it

test/Formik.test.tsx
samstarling: I just pushed a commit to remove this, thanks

test/Formik.test.tsx
samstarling: Looks like it is: I just pushed a fix.

johnrom:
With the last revisions, I'm liking this API. You should merge `master` into this branch so that there are no conflicts. If TypeScript has been updated in master, that will remove the need to update TypeScript.

Finally, it might be useful to make this callback async friendly in the future, similar to what was proposed in #909. But the process of using Promises for callbacks like this hasn't been established in v1 that I know of.

samstarling:
@johnrom Thanks – I might leave the async-friendliness until later then, if the process hasn't been established yet. For the `.size-snapshot.json`, how should I resolve that conflict? This branch also appears up to date with master, at least to me!

Andreyco:
Rebase, please

samstarling:
@Andreyco I've just rebased

src/Formik.tsx
```diff
@@ -131,12 +131,20 @@ export class Formik<Values = FormikValues> extends React.Component<
...
+          this.runValidations(values);
+        }
+      });
+    } else {
+      this.setState({ values }, () => {
+        if (this.props.validateOnChange) {
+          this.runValidations(values);
+        }
+      });
+    }
```
Andreyco: Since both branched do the same, extract `nextValues`.

```js
const nextValues = isFunction(valuesOrCallback) ? valuesOrCallback(this.state.values) : valuesOrCallback;

this.setState({ values: nextValues }, () => {
	if (this.props.validateOnChange) {
		this.runValidations(nextValues);
	}
})
```

src/Formik.tsx
```diff
@@ -131,12 +131,20 @@ export class Formik<Values = FormikValues> extends React.Component<
...
+    if (isFunction(values)) {
+      this.setState({ values: values(this.state.values) }, () => {
+        if (this.props.validateOnChange) {
+          this.runValidations(values);
+        }
+      });
+    } else {
+      this.setState({ values }, () => {
+        if (this.props.validateOnChange) {
+          this.runValidations(values);
```
Andreyco: Here and above you pass previous values (before update). Not good

src/types.tsx
```diff
@@ -62,6 +62,10 @@ export interface FormikComputedProps<Values> {
   readonly initialValues: Values;
 }
 
+/* Allow either raw values, or a method that takes values and returns values */
+export type ValueFunction<Values> = (values: Values) => Values;
+export type ValueUpdate<Values> = Values | ValueFunction<Values>;
+
```
Andreyco: Not sure about type names

src/Formik.tsx
```diff
@@ -131,12 +131,20 @@ export class Formik<Values = FormikValues> extends React.Component<
     });
   };
 
-  setValues = (values: FormikState<Values>['values']) => {
-    this.setState({ values }, () => {
-      if (this.props.validateOnChange) {
-        this.runValidations(values);
-      }
-    });
+  setValues = (values: ValueUpdate<Values>) => {
```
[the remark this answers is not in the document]
johnrom: `ValuesOrCallback`*

src/types.tsx
johnrom: I recommend: `ValuesCallback`, `ValuesOrCallback` (like you suggested above)

docs/api/formik.md
Andreyco: docs should be updated as well.

src/types.tsx
Andreyco: gosh, naming is hard. How about this?

```ts
interface ValuesSetter<V> {
    (values: Values): void;
    (valuesFactory: (prevValues: Readonly<V>) => V): void;
}

// and then
{
    setValues: ValuesSetter<Values>
}
```

johnrom: I like `ValuesFactory` and use of `Readonly`. I think describing the expected type (`ValuesOrValuesFactory = Values | ValuesFactory`) or describing the Type's usage (`SetValuesParam`) works best. `ValuesSetter` makes me think more of the actual setValues function, if it were to have a type.

Andreyco: Agree

Andreyco:
@samstarling would you be so kind and go thru change requests? TY!

src/types.tsx
```diff
@@ -80,7 +84,7 @@ export interface FormikActions<Values> {
   /** Manually set touched object */
   setTouched(touched: FormikTouched<Values>): void;
   /** Manually set values object  */
-  setValues(values: Values): void;
+  setValues(values: ValueUpdate<Values>): void;
```
Andreyco: I believe the are 2 places where this change is needed. pls, check

test/Formik.test.tsx
```diff
@@ -44,29 +45,36 @@ function Form({
   );
 }
 
-const InitialValues: Values = { name: 'jared' };
-
-function renderFormik<V>(props?: Partial<FormikConfig<V>>) {
+const InitialValues: Values = {
```
Andreyco: I like naming constants in a way they resemble constant :) E.g. `INITIAL_VALUES`

samstarling:
@andreyco Absolutely - thanks for all the reviewing, I’ll address the feedback tomorrow!

src/Formik.tsx
```diff
@@ -26,10 +27,9 @@ import {
   makeCancelable,
 } from './utils';
 
-export class Formik<Values = FormikValues> extends React.Component<
-  FormikConfig<Values>,
-  FormikState<Values>
-> {
+export class Formik<
+  Values extends FormikValues = FormikValues
+> extends React.Component<FormikConfig<Values>, FormikState<Values>> {
```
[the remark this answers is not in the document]
samstarling: No, this is actually how prettier formats this, because without breaks the line would be too long

src/Formik.tsx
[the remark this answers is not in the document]
samstarling: I noticed that the documentation always references the signature of `(fields: { [field: string]: any })`. Would it make sense to add the "true" signature (referencing the generic `Values`) when I update the documentation?

samstarling:
Hey everyone: the only outstanding change now is the documentation, but please let me know if I missed anything. I asked a question above about how the types are documented – once I know how to proceed, I'll update the docs and we should be good to go.

jaredpalmer:
This PR should be against v2.

samstarling:
@jaredpalmer Ah, I hadn't realised that, sorry – how can I fix that?

src/Formik.tsx
samstarling: That should be fixed now :)

src/Formik.tsx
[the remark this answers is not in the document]
samstarling: @Andreyco I just pushed an update that puts proper types into the documentation.

Andreyco:
I changed target to correct branch

src/types.tsx
```diff
@@ -155,13 +158,13 @@ export interface FormikHandlers {
...
  */
-export interface FormikSharedConfig {
+export interface FormikSharedConfig<Props = {}> {
   /** Tells Formik to validate the form on each input's onChange event */
   validateOnChange?: boolean;
   /** Tells Formik to validate the form on each input's onBlur event */
   validateOnBlur?: boolean;
   /** Tell Formik if initial form values are valid or not on first render */
-  isInitialValid?: boolean | ((props: object) => boolean | undefined);
+  isInitialValid?: boolean | ((props: Props) => boolean);
```
Andreyco: 👏

samstarling:
@Andreyco @jaredpalmer Sorry, I think I'm a bit confused about this being a PR against v2 now... It looks like it's so different that perhaps it's better for me to start again from scratch against that branch rather than try to rebase this?

johnrom:
@samstarling the way that state is managed internally is different, but your revised process of calculating the values isn't, right? I know hooks is a totally different paradigm so it may look foreign. (I don't remember all the type names we decided on so I'm just winging the below, but use whatever was approved above.

```
  const setValues = React.useCallback((valuesOrValuesFactory: Values | ValuesFactory<Values>) => {
    const values = isFunction(valuesOrValuesFactory) ? valuesOrValuesFactory(state.values) : valuesOrValuesFactory;
    dispatch({ type: 'SET_VALUES', payload: values });
  }, []);
```

Andreyco:
@samstarling If you dont mind I might pick this up

samstarling:
@Andreyco Please feel free – I'm sorry for dropping the ball on this, it's been a busy couple of weeks!

PatoBeltran:
What is the status of this PR? Any chance to get this in the next minor release instead of waiting for 2.0? Since its extending the current functionality of `setValues` it shouldn't be a breaking change.

FredyC:
@samstarling Can you please fix conflicts so the PR can move forward?

jaredpalmer:
just rebase against master

PatoBeltran:
@samstarling if you give me collab access to your fork I can take this PR and fix conflicts since I'm interested in getting this merged

samstarling:
@PatoBeltran I'm really sorry for the delay, and for neglecting this PR: you now have collaborator access (or at least, an invite)

PatoBeltran:
@jaredpalmer I just rebased from master and fixed the conflicts, anything else needed?

test/Formik.test.tsx
```diff
@@ -9,6 +9,7 @@ jest.spyOn(global.console, 'warn');
 
 interface Values {
   name: string;
+  zipcode?: number;
```
johnrom: formatting: `zipCode`, although since international users use different terminology / formats maybe `age` would be better.

PatoBeltran: Updated to use age instead of zip code

docs/api/formik.md
```diff
@@ -153,13 +153,13 @@ use it to pass API responses back into your component in `handleSubmit`.
 
 Set `isSubmitting` imperatively.
 
-#### `setTouched: (fields: { [field: string]: boolean }) => void`
+#### `setTouched: (fields: FormikTouched<Values>) => void`
 
 Set `touched` imperatively.
 
-#### `setValues: (fields: { [field: string]: any }) => void`
+#### `setValues: (valuesOrValuesFactory: ValuesOrValuesFactory<Values>) => void`
```
jaredpalmer:
```suggestion
#### `setValues: (valuesOrValuesUpdater: Values | (v: Values) => void) => void`
```
I think this is clearer.

jaredpalmer:
I can't push up my changes?

jaredpalmer:
@samstarling can you give me the ability to edit this PR?

https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

docs/api/formik.md
jaredpalmer: Update: We could just use `React.SetStateAction<S>`

samstarling:
Hi @jaredpalmer – sorry for the delay, I was away on holiday. You should be able to push now.

zetavg:
Hi everyone, is there any progress on this PR?

I think the ability to use _updater functions_ as the `setValues` argument is necessary [for some use cases that are setting the value based on the previous value](https://twitter.com/dan_abramov/status/816394376817635329), such as a counter that pressing a button will increase its value.
By using updater functions, possible race conditions can be avoided and that's an important reason why React the such called ["Functional updates"](https://reactjs.org/docs/hooks-reference.html#functional-updates) functionality for `setState`.

I'm interested in working on this PR (or another if preferred) to provide such functionality.

GunnarSturla:
This PR looks like it was really close, any chance of it being picked up again, or would it be better to start a new one?

johnrom:
I had assumed this was closed because the changes were integrated into another branch. If that's not the case, we should reopen.

GunnarSturla:
I can't find these changes on any other branch. It seems to have been closed shortly after being labelled stale.

I assume the PR needs to be rebased against a current branch (master or canary?) and, reading through the comments, I only saw some minor type fixes being suggested by @jaredpalmer; otherwise the PR could be good to go.

Is there anyone with edit permissions willing to take it over the finish line? @samstarling or @PatoBeltran maybe? I could also do it, if needed

Thanks for all your work, I think this PR will be a good addition to formik!

relaxomatic:
I would also really like this, this would be so much better for us. Avoid stale values and prevent the need for passing around the values object when we only need to update a few fields. We could really simplify our code with this change.

What do we need to do? I can help?

samotoo:
Wishing to see this PR get merged, it is neccessary to prevent stale values. At the moment I am using continuous setFieldValue calls to update multiple values, the enhanced API do help make the code more readable.

PatoBeltran:
Hey @jaredpalmer I updated the branch for this PR with latest master and refactored to use `React.SetStateAction` as the type instead of creating a new one. Can we reopen this PR? It would be nice to see this in the next formik's release.

PatoBeltran:
@jaredpalmer should I try to open a new PR to get traction on this functional updates with setValues feature or can we just reopen this one since the comments have already been addressed?

PR #1521, opened 2019-05-17
Fix TS build and remove types for hoist-non-react-statics

jaredpalmer:
I am giving up trying to solve https://github.com/mridgway/hoist-non-react-statics/issues/47 and also dealing with https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33690 . Including it in dependencies did fix the builds. Going to punt on this one. 

Fixes #1502 #1516

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "./dist/formik.umd.production.js": {
-    "bundled": 144255,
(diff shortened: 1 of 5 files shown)


jaredpalmer:
To add more detail, we weren't even really using the typedef anyways since we were casting to React.SFC each time anyways

PR #1522, opened 2019-05-18
[v2]: Distinguish between high/low priority validation

jaredpalmer:
After thinking more deeply about useEffect in v2, I don't think `useEffect()` is suitable for modeling Formik's validation strategies. In the current version of v2, Formik is over-validating and there is no way to avoid validating on change because the "validateOnBlur"-effect depends on `state.values`. This dependency got me thinking that useEffect isn't the correct approach (furthermore, this was going to be a breaking change too).

This PR refactors validation so that it is overridable just like Formik 1.x. Furthermore, validation triggered as a side effect of an update (i.e. `handleChange`, `setFieldValue`, `setValues`, `handleBlur`, `setFieldTouched`, `setTouched`) is now wrapped in `scheduler`'s `unstable_runWithPriority`. This tells React that the resulting update to `state.errors` is low-priority (or non-user-blocking). AFAICT, this is correct model of validation for onChange/onBlur. Conversely, validation triggered imperatively or during pre-submit via `validateForm`, `validateField`, `submitForm`, `handleSubmit` is the same as it was (high-pri/user-blocking). 

In a different PR, I want to play around with removing async validation entirely from Formik. The thinking is the same as rationale as to why you should not use something like Redux for making async requests, but instead just use it a store. 

Anyways, open to ideas. I also now need to fixup the tests to flush out some stuff.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -399,7 +399,7 @@ export function useFormik<Values = object>({
   React.useEffect(() => {
     if (
       enableReinitialize &&
-      isMounted.current != null &&
+      isMounted.current === true &&
       !isEqual(initialValues.current, props.initialValues)
     ) {
       resetForm();
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -259,9 +259,9 @@ export function useFormik<Values = object>({
       props.validationSchema ||
       props.validate ||
       (fields.current &&
-        Object.keys(fields.current).filter(
+        Object.keys(fields.current).some(
           key => !!fields.current[key].validate
-        ).length > 0),
+        )),
     [props.validate, props.validationSchema, fields]
   );
 
@@ -324,7 +324,10 @@ export function useFormik<Values = object>({
       }
       dispatch({ type: 'SET_ISVALIDATING', payload: true });
       return runAllValidations(values).then(combinedErrors => {
-        if (!!isMounted.current && !isEqual(state.errors, combinedErrors)) {
+        if (
+          isMounted.current === true &&
+          !isEqual(state.errors, combinedErrors)
+        ) {
           dispatch({ type: 'SET_ERRORS', payload: combinedErrors });
           dispatch({ type: 'SET_ISVALIDATING', payload: false });
         }
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -20,7 +20,6 @@ import {
   setNestedObjectValues,
   getActiveElement,
   getIn,
-  makeCancelable,
 } from './utils';
 import { FormikProvider } from './FormikContext';
 import invariant from 'tiny-warning';
@@ -103,14 +102,9 @@ function formikReducer<Values>(
   }
 }
 
-/** @private */
-function usePrevious<T>(val: T) {
-  const ref = React.useRef<T>(val);
-  React.useEffect(() => {
-    ref.current = val;
-  }, [val]);
-  return ref.current;
-}
+// Initial empty states
+const emptyErrors: FormikErrors<any> = {};
+const emptyTouched: FormikTouched<any> = {};
 
 export function useFormik<Values = object>({
   validateOnChange = true,
@@ -122,8 +116,8 @@ export function useFormik<Values = object>({
 }: FormikConfig<Values>) {
   const props = { validateOnChange, validateOnBlur, onSubmit, ...rest };
   const initialValues = React.useRef(props.initialValues);
-  const initialErrors = React.useRef(props.initialErrors || {});
-  const initialTouched = React.useRef(props.initialTouched || {});
+  const initialErrors = React.useRef(props.initialErrors || emptyErrors);
+  const initialTouched = React.useRef(props.initialTouched || emptyTouched);
   const initialStatus = React.useRef(props.initialStatus);
   const isMounted = React.useRef<boolean>(false);
   const fields = React.useRef<{
@@ -160,18 +154,17 @@ export function useFormik<Values = object>({
     isValidating: false,
     submitCount: 0,
   });
-  const prevState = usePrevious(state);
 
   const runValidateHandler = React.useCallback(
     (values: Values, field?: string): Promise<FormikErrors<Values>> => {
       return new Promise(resolve => {
         const maybePromisedErrors = (props.validate as any)(values, field);
         if (maybePromisedErrors === undefined) {
-          resolve({});
+          resolve(emptyErrors);
         } else if (isPromise(maybePromisedErrors)) {
           (maybePromisedErrors as Promise<any>).then(
             () => {
-              resolve({});
+              resolve(emptyErrors);
             },
             errors => {
               resolve(errors);
@@ -201,7 +194,7 @@ export function useFormik<Values = object>({
             : validateYupSchema(values, schema);
         promise.then(
           () => {
-            resolve({});
+            resolve(emptyErrors);
           },
           (err: any) => {
             resolve(yupToFormErrors(err));
@@ -223,8 +216,8 @@ export function useFormik<Values = object>({
 
   const runFieldLevelValidations = React.useCallback(
     (values: Values): Promise<FormikErrors<Values>> => {
-      if (fields.current === null) {
-        return Promise.resolve({});
+      if (fields.current == null) {
+        return Promise.resolve(emptyErrors);
       }
       const fieldKeysWithValidation: string[] = Object.keys(
         fields.current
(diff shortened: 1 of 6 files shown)


src/Formik.tsx
```diff
@@ -259,92 +252,94 @@ export function useFormik<Values = object>({
...
+  const hasValidationFns = React.useMemo(
+    () =>
+      props.validationSchema ||
+      props.validate ||
+      (fields.current &&
+        Object.keys(fields.current).filter(
+          key => !!fields.current[key].validate
+        ).length > 0),
+    [props.validate, props.validationSchema, fields]
+  );
```
Andreyco: me likes 👍

src/Formik.tsx
```diff
@@ -259,92 +252,94 @@ export function useFormik<Values = object>({
...
-    isMounted,
-  ]);
+  // Run validations and dispatching the result as low-priority via rAF.
+  //
+  // The thinking is that validation as a result of onChange and onBlur
+  // should never block user input. Note: This method should never be called
+  // during the submission phase because validation prior to submission
+  // is actaully high-priority since we absolutely need to guarantee the
+  // form is valid before executing props.onSubmit.
+  const validateFormWithLowPriority = React.useCallback(
```
Andreyco: this should improve performance and possibly avoid flickering, correct?

src/Formik.tsx
```diff
@@ -259,92 +252,94 @@ export function useFormik<Values = object>({
...
+    (values: Values = state.values) => {
+      if (!hasValidationFns) {
+        // No validation functions, so we bail out immediately
+        return Promise.resolve(emptyErrors);
```
Andreyco: How about putting this resolver into "all validation handlers" and then if/else block here would be redundant.

src/Formik.tsx
```diff
@@ -259,92 +252,94 @@ export function useFormik<Values = object>({
...
+      }
+      dispatch({ type: 'SET_ISVALIDATING', payload: true });
+      return runAllValidations(values).then(combinedErrors => {
+        if (
+          !isEqual(state.errors, combinedErrors) &&
+          isMounted.current != null
```
Andreyco: if you switch order of this conditions, `isEqual` would not need to run if component is not mounted, thus slightly faster in edge cases. Correct?

src/Formik.tsx
```diff
@@ -476,36 +471,55 @@ export function useFormik<Values = object>({
...
-          handlerName: 'handleBlur',
-        });
-      }
+  const setErrors = React.useCallback((errors: FormikErrors<Values>) => {
+    dispatch({ type: 'SET_ERRORS', payload: errors });
+  }, []);
+
+  const setValues = React.useCallback(
+    (values: Values) => {
+      dispatch({ type: 'SET_VALUES', payload: values });
```
Andreyco: in general, I like to avoid using magic string. Constants instead?

Andreyco:
as much as I love hooks, I hope they won't introduce unpredictable inconsistencies/race conditions :D

src/Formik.tsx
jaredpalmer: Agree.

src/Formik.tsx
jaredpalmer: Correct.

src/Formik.tsx
jaredpalmer: We can probs switch to just an enum

src/Formik.tsx
jaredpalmer: Do we even need this isEqual? I have lost track at a this point.

src/Formik.tsx
jaredpalmer: This _should_ tell React to de-prioritize this work.

src/Formik.tsx
Andreyco: Good point, probably not. I think we are good without it

src/Formik.tsx
```diff
@@ -259,92 +253,91 @@ export function useFormik<Values = object>({
...
-    return;
-  }, [
-    prevState.touched,
-    state.isSubmitting,
-    state.touched,
-    validateForm,
-    validateOnBlur,
-    isMounted,
-  ]);
+      return unstable_runWithPriority(LowPriority, () => {
```
jaredpalmer: @aweary is this correctly implemented? My thinking is that form validation in response to user input is less important than the actual input, and thus should be marked as low-pri (it is a decent candidate for debounce). Do I need to also mark the update to value in response to `onChange` as Immediate or User Blocking?

jaredpalmer: I confirmed this use case with Brandon in person.

src/Formik.tsx
```diff
@@ -561,62 +573,61 @@ export function useFormik<Values = object>({
...
-        value: touched,
-      },
-    });
-  }, []);
+      function executeBlur(e: any, path?: string) {
+        if (e.persist) {
+          e.persist();
+        }
+        const { name, id, outerHTML } = e.target;
+        const field = path ? path : name ? name : id;
```
KITSEPAMIgorSergiienko: `path || name || id`?

FredyC: Use the suggestions feature next time so it's easier for maintainers.
```suggestion
        const field = path || name || id;
```

Andreyco: Personally I would limit this to required `name` prop and that's it... @jaredpalmer What do you think? Enforce `name` prop on fields for v2?

src/Formik.tsx
```diff
@@ -259,92 +253,91 @@ export function useFormik<Values = object>({
...
-    isMounted,
-  ]);
+  // Run validations and dispatching the result as low-priority via rAF.
+  //
+  // The thinking is that validation as a result of onChange and onBlur
+  // should never block user input. Note: This method should never be called
+  // during the submission phase because validation prior to submission
+  // is actaully high-priority since we absolutely need to guarantee the
+  // form is valid before executing props.onSubmit.
+  const validateFormWithLowPriority = React.useCallback(
```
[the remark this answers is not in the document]
jaredpalmer: Makes sense

src/Formik.tsx
```diff
@@ -223,8 +217,8 @@ export function useFormik<Values = object>({
 
   const runFieldLevelValidations = React.useCallback(
     (values: Values): Promise<FormikErrors<Values>> => {
-      if (fields.current === null) {
-        return Promise.resolve({});
+      if (fields.current == null) {
```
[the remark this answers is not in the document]
jaredpalmer: good catch.

src/Formik.tsx
```diff
@@ -259,92 +253,91 @@ export function useFormik<Values = object>({
...
-      } else {
-        return Promise.resolve({});
-      }
+  // Are there any validation methods?
+  const hasValidationFns = React.useMemo(
+    () =>
+      props.validationSchema ||
+      props.validate ||
+      (fields.current &&
+        Object.keys(fields.current).filter(
```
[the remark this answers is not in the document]
jaredpalmer: Valid.

src/Formik.tsx
```diff
@@ -259,92 +253,91 @@ export function useFormik<Values = object>({
...
+          key => !!fields.current[key].validate
+        ).length > 0),
+    [props.validate, props.validationSchema, fields]
+  );
+
+  // Run all validations and return the result
+  const runAllValidations = React.useCallback(
```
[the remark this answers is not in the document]
jaredpalmer: Valid

src/Formik.tsx
jaredpalmer: I forgot when it happened exactly, but we used to support `id` only. I think things got weird with FieldArray IIRC. I just want us to be consistent.

src/Formik.tsx
```diff
@@ -476,36 +469,55 @@ export function useFormik<Values = object>({
...
-    | ((e: any) => void) => {
-    if (isString(eventOrString)) {
-      return event => executeBlur(event, eventOrString);
-    } else {
-      executeBlur(eventOrString);
-    }
+  const setTouched = React.useCallback(
+    (touched: FormikTouched<Values>) => {
+      dispatch({ type: 'SET_TOUCHED', payload: touched });
+      return validateOnBlur
```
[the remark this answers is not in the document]
jaredpalmer: This is how it works now in v1. tbh, `validateOnBlur` really means `validateOnTouched` the more i think about it.

src/Formik.tsx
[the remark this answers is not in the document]
jaredpalmer: FWIW the eslint-plugin-react-hooks gets mad when I remove the useCallback here. Thoughts?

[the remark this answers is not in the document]
jaredpalmer: cc @gaearon. I usually just let the hooks plugin do its job. However, does @fredyc have a point about avoiding useCallback here?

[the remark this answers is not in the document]
FredyC: I don't use eslint plugin as it currently gets more in the way that it helps. Perhaps my understanding of `useCallback` is wrong right now, but cannot imagine why it would want to enforce it for such internal function.

src/Formik.tsx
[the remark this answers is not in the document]
FredyC: Well, as I said, it would be nice to give the option to opt-out of this behavior for a `setTouched` call. In case I really do want to set touched for one field and avoid validation for some reason. I don't have a real use case though, so perhaps it's not worth it.

src/Formik.tsx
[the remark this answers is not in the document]
jaredpalmer: I think I am going to go through and pull out a lot of these functions in another PR. Let's keep this discussion focused on high/low pri.

src/Formik.tsx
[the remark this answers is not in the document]
jaredpalmer: The eslint plugin automagically adds deps so this _should_ work as expected.

src/Formik.tsx
jaredpalmer: @Andreyco lets move this to a new issue

src/Formik.tsx
```diff
@@ -209,25 +213,21 @@ export function useFormik<Values = object>({
   const runSingleFieldLevelValidation = React.useCallback(
     (field: string, value: void | string): Promise<string> => {
       return new Promise(resolve =>
-        resolve(fields.current[field].validate(value))
-      ).then(x => x, e => e);
+        resolve(fieldRegistry.current[field].validate(value))
+      );
     },
-    [fields]
+    [fieldRegistry]
```
FredyC: It's a general recommendation not to use refs in dependencies because their reference never changes. Not that it would cause any harm in this situation, just pointing out... Unless eslint plugin gets crazy about it (👎).

src/Formik.tsx
```diff
@@ -411,13 +389,13 @@ export function useFormik<Values = object>({
       // what is called when using validateForm.
 
       if (
-        fields.current !== null &&
-        fields.current[name] &&
-        fields.current[name].validate &&
-        isFunction(fields.current[name].validate)
+        fieldRegistry.current !== null &&
+        fieldRegistry.current[name] &&
```
FredyC: I think it's safe to omit this check, the non-truthy value wouldn't pass through `typeof === 'function'` anyway.

jaredpalmer: Yep

src/Formik.tsx
jaredpalmer: See #1558

src/Formik.tsx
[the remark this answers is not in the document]
jaredpalmer: I agree. If we didn't need to keep v2 backwards compat, i would not have this side-effect stuff. This is something we will fix for v3.

jaredpalmer:
Going to merge.

PR #1536, opened 2019-05-20
Fixes #1489: Allow error value to be any

### LINKED ISSUE #1489: Why limit error to just a string?
### Current Behavior

Form errors are currently of type `string`. In most other form validation libraries I've used this is an `any` type so a user could do things like the following:

```js
setErrors({
  password: ['Must be more than 8 characters', 'Must contain a special character']
})
```
or for localization:

```js
setErrors({
  password: [{type: 'min', min: 3, value: 1}]
})
```

```js
setErrors({
  password: [{localizationKey: 'blah', value: 1}]
})
```

Is there any reason why it needs to strictly be of type `string`? It actually works setting it to whatever I want so long as I cast it but makes working with TypeScript funky and limits the flexibility of users.

Similar to #1244, #1293

jefk:
See https://github.com/jaredpalmer/formik/issues/1489

The current typing of errors is incorrect, Formik allows setting error state to more than a string. In my case, I need to set the value to a `JSX.Element`.

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -11,9 +11,7 @@ export interface FormikValues {
  * Should be always be and object of strings, but any is allowed to support i18n libraries.
  */
 export type FormikErrors<Values> = {
-  [K in keyof Values]?: Values[K] extends object
-    ? FormikErrors<Values[K]>
-    : string
+  [K in keyof Values]?: any;
 };
 
 /**


jefk:
Actually, the fix will be a bit more complicated. I will reopen after looking through the consequences.

Andreyco:
Instead of opting for `any`, how about using generic type? I'd like to ask you to not submit PR with `any` type, it won't be accepted.

Issue #1525, opened 2019-05-18
[v2]: Add parse and format props to useField/Field


jaredpalmer:

Andreyco:
Yes please. I am curious about useField API. Do you have anything on your mind already?

Ciaran0:
This would be game changing

klis87:
For now I did such stuff in my custom fields, adjusting `onChange` and `value`, would be cool to have it built-in!

Fauntleroy:
This is exactly what I need right now. I've got lots of glue code in unreliable places converting ids to numbers all over the place as it stands.

johnrom:
@jaredpalmer this is what I meant by suggesting a valueConverter function, but these names make so much more sense 👍 

I'd like to suggest basing this off of #1334 so parse and format have access to real types in TypeScript. Also to recap something I was thinking about as part of this, it'd be nice to not require the parse / format props when a field has a string value, but to _require_ it when a field has a number value. I couldn't figure out how to make TS+React play nicely there.

joepuzzo:
https://joepuzzo.github.io/informed/?path=/story/formatting--format-and-parse If you want to take a look at how I've implemented format + parse, and mask

szimek:
Just one remark. 

I'm currently implementing a credit card payment form and was looking at how Stripe forms behave and they are doing pretty complex stuff with setting cursor position etc. especially with the expiration date input. E.g. when you type `10`, it will display `10 / ` and move the cursor at the end, so that you can enter the year. If you then press backspace, it will display `10`, removing ` / ` at the end. In both cases, the parsed value for month would be `10` (unless they don't parse incomplete fields at all), but the formatted (i.e. displayed) value is different, once it's `10 / ` and once it's `10`. So it looks like for such use case, `format` function would not only need access to the current value, but to the last event as well. Another issue is that one input provides 2 values - expiration month and year.

However, maybe `parse` and `format` props don't have to support such complex use cases.

joepuzzo:
Something like this https://joepuzzo.github.io/informed/?path=/story/formatting--mask-with-cursor-offset ?

szimek:
@joepuzzo Yeah, though I'm not sure if you can replicate Stripe behavior exactly, if the `mask` function doesn't additionally has access to an event, because as I described above, for the same input value (e.g. `10`), it will display either `10 / ` or `10`. 

You can play with Stripe forms here: https://stripe.dev/elements-examples/

johnrom:
[...]

Issue #360, opened 2018-01-21
How to populate array with checkboxes?


Woile:
## How to populate array with checkboxes?

Hi, I'm wondering if it's possible to do this. 
I have 2 check-boxes, and every time one is checked, an array should be populated. I've been playing around, even tried the latest `FieldArray`, but I'm not sure I understand well how to use it, or even if it's for that purpose.

## Info
 
I am using `Formik` component with `render` method.
Something like:
```js
import React from 'react';
import { render } from 'react-dom';
import { Formik } from 'formik';


const App = () => (
  <div>
    <h1>My Form</h1>
    <Formik
      initialValues={{
        groups: [],
      }}
      onSubmit={values =>
        console.log(values)
        // Expected output would be something like:
        // {groups: ['admin', 'normal']}
      }
      render={props => (
        <form onSubmit={props.handleSubmit}>
          <input
            type="checkbox"
            name="groups.0"
            value="admin"
            onChange={props.handleChange}
            onBlur={props.handleBlur}
            checked={props.values.groups === 'admin'}
          />
          Admin
          <br/>
          <input
            type="checkbox"
            name="groups.1"
            value="normal"
            onChange={props.handleChange}
            onBlur={props.handleBlur}
            checked={props.values.groups === 'normal'}
          />
          Normal
          <br />

          <button type="submit" >
            Submit
          </button>
        </form>
      )}
    />
  </div>
);

render(<App />, document.getElementById('root'));
```

Here is an example: https://codesandbox.io/s/03pmmoo0kp

But what I really want is to have this as a result:
```js
{
  groups: ['admin', 'normal']
}
```
Is it possible doing something like that? What am I missing?

- Formik Version: `0.11-beta`
- OS: Debian unstable
- Node Version: ` node@v8.6.0`
- Package Manager and version: `yarn@1.3.2`

Woile:
Okay, found a solution. I'm still wondering if it's the way to go:

```js
import React from 'react';
import { render } from 'react-dom';
import { Formik } from 'formik';


const App = () => (
  <div>
    <h1>My Form</h1>
    <Formik
      initialValues={{
        groups: [],
      }}
      onSubmit={values =>
        console.log(values)
        // Expected output would be something like:
        // {groups: ['admin', 'normal']}
      }
      render={props => (
        <form onSubmit={props.handleSubmit}>
          <input
            type="checkbox"
            onChange={(event) => {
              const value = event.target.checked ? 'admin' : ''
[...]

PR #1555, opened 2019-05-28
[v2] Add support for checkboxes and multiple select

jaredpalmer:
This brings `useField` / `<Field>` / `getFieldProps` inline with similar checkbox and select behavior found in [Vue](https://vuejs.org/v2/guide/forms.html), [Svelte](https://svelte.dev/docs#Element_bindings), and even [Angular](https://netbasal.com/handling-multiple-checkboxes-in-angular-forms-57eb8e846d21)

### Checkboxes
```jsx
// Single checkbox, boolean value.
<Field type="checkbox" name="newsletter" /> // => true/false

// Multiple checkboxes, bound to the same Array.
<Field type="checkbox" name="colors" value="red" /> 
<Field type="checkbox" name="colors" value="green" /> 
<Field type="checkbox" name="colors" value="blue" />  
// => so if they are all checked, will be stored as 
// { colors: ["red", "green", "blue"] } 
// checking/unchecking will add/remove item from the array
```

### Select

```jsx
// Single select 
<Field name="car" as="select">
  <option>Please select one</option>
  <option value="saab">Saab</option>
  <option value="opel">Opel</option>
  <option value="audi">Audi</option>
</Field>

// Multiple select (bound to an Array)
<Field name="car" as="select" multiple>  
  <option value="saab">Saab</option>
  <option value="opel">Opel</option>
  <option value="audi">Audi</option>
</Field> // => { car: ['saab', 'opel', 'audi'] }
```

## Implementation

- Refactor `getFieldProps` to accept an object.
- Allow `useField` to take an object (of your props)
- Do magical stuff when `type="checkbox"`
- Do magical stuff when there is a prop called `multiple`

## Discussion

This is backwards compat, but we may or may not want to call out in the docs that `getFieldProps` includes some magic (if you want it to). 

Closes #527 #211 #360 #1115 

Related to #1525


src/Formik.tsx
```diff
@@ -921,3 +942,46 @@ function arrayMerge(target: any[], source: any[], options: any): any[] {
...
+
+/** Return the next value for a checkbox */
+function getValueForCheckbox(
+  currentValue: string | any[],
+  checked: boolean,
+  valueProp: any
+) {
+  if (valueProp == 'true' || valueProp == 'false') {
+    return !!checked;
+  } else {
```
igor0ser: `else` is redundant if you have `return`

src/Formik.tsx
```diff
@@ -921,3 +942,46 @@ function arrayMerge(target: any[], source: any[], options: any): any[] {
...
+    if (checked) {
+      return Array.isArray(currentValue)
+        ? currentValue.concat(valueProp)
+        : [valueProp];
+    } else {
+      if (!Array.isArray(currentValue)) {
+        return !!currentValue;
+      }
+      const index = currentValue.indexOf(valueProp);
```
igor0ser: you can use `includes` to instead of `indexOf`

src/Formik.tsx
```diff
@@ -921,3 +942,46 @@ function arrayMerge(target: any[], source: any[], options: any): any[] {
...
+  if (options) {
+    for (let index = 0; index < options.length; index++) {
+      const option = options[index];
+      if (option.selected) {
+        result.push(option.value);
+      }
+    }
+  }
+  return result;
+}
```
igor0ser:
```js
function getSelectedValues(options: any[]) {
   return options.filter(option => option.selected);
 }
```

Andreyco:
@jaredpalmer planning to do a review on this

src/Formik.tsx
jaredpalmer: Yep.

src/Formik.tsx
jaredpalmer: I need to reuse the index.

jaredpalmer:
I think this all makes a lot of sense, my only one concern is that this could get implemented in userland as `useCheckbox` and `useCheckboxGroup` etc. idk.

src/Field.tsx
```diff
@@ -64,17 +64,33 @@ export interface FieldConfig {
...
   if (process.env.NODE_ENV !== 'production') {
     invariant(
       formik,
       'useField() / <Field /> must be used underneath a <Formik> component or withFormik() higher order component'
     );
   }
-  return formik.getFieldProps<Val>(name, type);
+
+  if (isObject(propsOrFieldName)) {
+    if (process.env.NODE_ENV !== 'production') {
```
Andreyco: minor, but I like more __DEV__ constant provided by tsdx (and I know you know)

src/Formik.tsx
```diff
@@ -522,8 +525,10 @@ export function useFormik<Values = object>({
           }
           val = /number|range/.test(type)
             ? ((parsed = parseFloat(value)), isNaN(parsed) ? '' : parsed)
-            : /checkbox/.test(type)
-            ? checked
+            : /checkbox/.test(type) // checkboxes
+            ? getValueForCheckbox(getIn(state.values, field!), checked, value)
+            : !!multiple // <select multiple>
+            ? getSelectedValues(options)
```
Andreyco: too many ternary operators here

Andreyco:
While this is very often requested change, I think it should not be part of formik core.

This change introduces unnecessary noise and possible error surface area into core, while it can be implemented thru user land code, or provided as special set of official extensions (custom fields, hooks, etc).

Exactly as within your implementation, checkboxes represent collections/arrays of selected options, in my opinion. Then, there are special use cases when checkbox serves as a boolean switch. This is exact reason why so many if/else conditions landed in this PR and this could not be generalised in a "proper" way.

How about providing official components that make use of already lean Formik's core?

src/Field.tsx
jaredpalmer: Let's make a different issue for that.

Andreyco: Sure, I can do a find and replace on whole repo

jaredpalmer:
You raise good points and I agree this adds a lot of complexity. However, it has some precedent in the sense that I quite literally implemented the Vue's API.

@eonwhite What do you think? `<Field>` has been magical since day 1.

jaredpalmer:
also ping @nathanforce @crosscompile.

src/Field.tsx
```diff
@@ -64,17 +64,32 @@ export interface FieldConfig {
...
+      invariant(
+        (propsOrFieldName as FieldAttributes<Val>).name,
```
johnrom: `isObject` should be defined to narrow the type automatically, like:

```
type NonObject = string | number | boolean | symbol | null | undefined;
const isObject = <T extends object>(maybeObject: NonObject | T): maybeObject is T => 
    typeof value === 'object'; // or whatever isObject's test really is, maybe check for array?
// then
if (isObject(propsOrFieldName)) { // propsOrFieldName is inferred as FieldAttributes<Val> }
```

jaredpalmer: Can you open a new issue for this?

jaredpalmer:
I would like to get more eyes on this before going forward. cc @elado @slightlytyler @prichodko

src/Field.tsx
johnrom: Done, I'll resolve this conversation

#1599

src/Formik.tsx
```diff
@@ -718,20 +718,36 @@ export function useFormik<Values extends FormikValues = FormikValues>({
...
-    ): [FieldInputProps<any>, FieldMetaProps<any>] => {
-      const field = {
+    ({
+      name,
+      type,
+      value: _value,
+      as: is,
+      multiple,
+    }): [FieldInputProps<any>, FieldMetaProps<any>] => {
+      const value = getIn(state.values, name);
```
crosscompile: For this `value` and `_value`, is there a name that more explicitly describes one?  Reading through it's hard to tell what the difference between each is.

jaredpalmer: value vs. valueProp

crosscompile: Ah got it, I think something like `valueState` and `valueProp` would help me differentiate.

jaredpalmer: Yep

adrianwix:
Hi, just a question. I am using a Field type="checkbox" component with a CustomComponent. In Formik I am passing the initialValue of that component like `{ open: false }` but inside the custom component I am getting a PropTypes validation error saying that field.value is undefined. Which would be the expected behavior?
``````

**Comment by jaredpalmer** [comment]
@adrianwix can you create a codesandbox demo so i can take a look?

**Comment by adrianwix** [comment]
@jaredpalmer I just leave my laptop with the project repairing. I can upload it to GitHub and send you the link to the project next Tuesday. The project is not so big I have only implement one page in the client site a not completely and the server but can be completely ignored.

**Comment by adrianwix** [comment]
> @adrianwix can you create a codesandbox demo so i can take a look?

https://codesandbox.io/s/create-react-app-6f7wu That's the example

**Comment by kevinmiguet** [comment]
Somehow I can't make it work with useFormik method `const Home: NextPage<FormikProps<HomeProps>> = (props) => {
  const { values, setFieldValue } = props
  useEffect(() => {
      console.log(values)
    }, [values])
    return (
      <Form>
        Colors
        <Field type="checkbox" name="colors" value="red" /> 
        <Field type="checkbox" name="colors" value="green" /> 
        <Field type="checkbox" name="colors" value="blue" />  
      </Form>
    )
};

const HomeForm = withFormik<Operation[], any>({

  handleSubmit: (values) => {
  },

  displayName: 'HomeFiltersForm',

})(Home);

export default HomeForm;`

and when tickling these checkboxes it just show true or false... am I doing something wrong ?

**Comment by cyonder** [comment]
Why type property isn't passed inside field? Both name and value is passed though. Why do I have to write the type again?

```
<Field name="schedule" type="radio" value="Wow">
    {({ field }) => <Form.Check { ...field } custom type="radio" id="anId" label="Seriously?"/>}
</Field>
```

johnrom:
@cyonder might be better to open a new issue with your question, as this PR is resolved.

PR #1559, opened 2019-05-29
[v2] Upgrade to TS 3.5.x

jaredpalmer:
See https://devblogs.microsoft.com/typescript/announcing-typescript-3-5


jaredpalmer:
Can someone with more TypeScript fu than myself, review this one?

johnrom:
I'll take a look when I have a chance.

dgreene1:
looks good to me. Nice use of `unknown`

src/Formik.tsx
```diff
@@ -257,7 +257,7 @@ export function useFormik<Values = object>({
         props.validate ? runValidateHandler(values) : {},
       ]).then(([fieldErrors, schemaErrors, validateErrors]) => {
         const combinedErrors = deepmerge.all<FormikErrors<Values>>(
-          [fieldErrors, schemaErrors, validateErrors],
+          [fieldErrors, schemaErrors, validateErrors] as FormikErrors<Values>[],
```
johnrom: `validateSchemaErrors` should return a `Promise<FormikErrors<Values>>`, then this cast can be removed.

jaredpalmer:
> ```
> const maybePromisedErrors = (props.validate as any)(values, field);
> ```
> 
> `props.validate` may be undefined. Is there a reason we are still casting this to any? I assume it's because validate used to accept a second parameter, but does not anymore. Should we get rid of this param as part of v2, and remove the cast?

Yes

jaredpalmer:
@johnrom @weswigham Is there a way to programmatically test Formik's types against versions of TypeScript and @types/react. I feel like it would be nice to show a matrix or chart of what TS version we actually support. With adding this `unknown` to the types I also don't fully understand what happens to users on earlier versions of TS.

src/Formik.tsx
jaredpalmer: I thought I tried that and TS was yelling. Let me do it again

PR #1566, opened 2019-05-30
Mimic class properties with useEventCallback

jaredpalmer:
In v1, all helper methods we class properties. This meant that they were bound to the instance of the class. However, with hooks, there is no concept of `this` and each helper method is wrapped in `useCallback`. However, certain behaviors (e.g. auto-saving) are not possible because of circular `useCallback` dependencies. For example, `submitForm` depends on `values`, so if you change `values`, then `submitForm` will be different. This prevents you from debouncing `submitForm` in response to `values` because you get a brand new function on each change. Sadness.

[The suggested solution in the React docs](https://reactjs.org/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback) is to use this hook:

```jsx
function Form() {
  const [text, updateText] = useState('');
  // Will be memoized even if `text` changes:
  const handleSubmit = useEventCallback(() => {
    alert(text);
  }, [text]);

  return (
    <>
      <input value={text} onChange={e => updateText(e.target.value)} />
      <ExpensiveTree onSubmit={handleSubmit} />
    </>
  );
}

function useEventCallback(fn, dependencies) {
  const ref = useRef(() => {
    throw new Error('Cannot call an event handler while rendering.');
  });

  useEffect(() => {
    ref.current = fn;
  }, [fn, ...dependencies]);

  return useCallback(() => {
    const fn = ref.current;
    return fn();
  }, [ref]);
}
```

This PR adds this hook internally and utilizes it where necessary (on any callback that relies on `state`). It also adds a new `DebouncedAutoSave` example to the examples directory.

### DIFF
--- a/.storybook/example.stories.js
+++ b/.storybook/example.stories.js
@@ -13,11 +13,14 @@ import SchemaValidation from '../examples/SchemaValidation';
 import SyncValidation from '../examples/SyncValidation';
 import FieldLevelValidation from '../examples/FieldLevelValidation';
 import CombinedValidations from '../examples/CombinedValidations';
+import DebouncedAutoSave from '../examples/DebouncedAutoSave';
 
 const AsyncValidationCode = require('!raw-loader!../examples/AsyncValidation')
   .default;
 const ArraysCode = require('!raw-loader!../examples/Arrays').default;
 const BasicCode = require('!raw-loader!../examples/Basic.js').default;
+const DebouncedAutoSaveCode = require('!raw-loader!../examples/DebouncedAutoSave.js')
+  .default;
 const CustomInputsCode = require('!raw-loader!../examples/CustomInputs')
   .default;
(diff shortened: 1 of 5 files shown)


VanTanev:
`useEventCallback()` can be written as

```typescript
function useEventCallback<T extends (...args: any[]) => any>(
  fn: T
): T {
  const ref: any = React.useRef(() => {
    throw new Error('Cannot call an event handler while rendering.');
  });

  // we copy a ref to the callback scoped to the current state/props on each render
  React.useLayoutEffect(() => {
    ref.current = fn;
  });

  return React.useCallback((...args) => (0, ref.current)(...args), []) as T;
}
```

as per https://github.com/facebook/react/issues/14099#issuecomment-440172008

To me this would make it more clear that `useEventCallback()` is essentially a hack to have a callback which has the entire current state/props, and not only the deps, but the identity of the callback itself doesn't change, making it safe to send downstream and not cause rerenders.

Issue #1565, opened 2019-05-30
V2 Formik Provider not being updated with new values on reinitialize


outerlook:
### Current Behavior

Hi! Passing different initialValues won`t update the provider even with enableReinitialize=true.
Am I missing something?

### Expected behavior
When I change initialValues, I expect a "dirty = false" with the new values, for example when I update a form then ask my server to refetch data.

### Reproducible example

[Here's the sandbox](https://codesandbox.io/s/formik-codesandbox-template-u5r5w)

Using variable name as state of my component.

Formik initiate with current state

I put a button that will directly update the state to a new name "John".

When the button triggers, the component updates with new state but formik props won't change.

Am I missing something simple? Sorry if I am. Thanks in advance

### Your environment

| Software         | Version(s) |
Reproduced on Sandbox with
![image](https://user-images.githubusercontent.com/12937160/58647781-a9861c80-82de-11e9-8831-f1302537dd19.png)

falldowngoboone:
Possibly same as #1528

outerlook:
Yes, it is. Missed this open issue, sorry 😅

Issue #1528, opened 2019-05-19
[v2] enableReinitialize broken after 2.0.1-alpha.2


jaripekkala:
### Current Behavior

`enableReinitialize` is not updating the `initialValues` anymore on change

### Expected behavior

2.0.1-alpha.2 https://codesandbox.io/s/formik-codesandbox-template-z5974

### Reproducible example

2.0.1-alpha.3 https://codesandbox.io/s/formik-codesandbox-template-scl9q
2.0.1-rc.0 https://codesandbox.io/s/formik-codesandbox-template-lgcrx

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.1-rc.0
| React            | 16.8.6

Unrealman1:
can confirm I am experiencing the same issue

jaripekkala:
I tried to figure this out after creating the issue and couldn't find anything obvious causing this issue. Could somebody else look into this who's more familiar with the codebase and recent changes?

However there is prior open issue that will block usage even if this gets fixed https://github.com/jaredpalmer/formik/issues/1523

falldowngoboone:
#1523 is closed now.

falldowngoboone:
Looks like the initial values ref (`initialValues.current`) is set as the default value for resetting the form on reinitialize. I'm thinking a new Formik state with the values set to `props.initialValues` should be passed to `resetForm` when the form is reinitialized.

falldowngoboone:
Actually, `initialValues.current` should be set to `props.initialValues` before `resetForm` is set. Working on a fix right now.

PR #1570, opened 2019-05-31
Fix initialValues ref when `enableReinitialize` is true

kimochg:
### What

enableReinitialize should work now, ~~and an enableReinitialize example is added.~~

### Issues

https://github.com/jaredpalmer/formik/issues/1565
https://github.com/jaredpalmer/formik/issues/1528


falldowngoboone:
Tests probably need to be updated as well.

Andreyco:
wasn't fix for this merged already by @jaredpalmer ?

jaredpalmer:
I lost track

jaredpalmer:
It’s being deprecated in v2

kimochg:
@jaredpalmer Example removed.  Nicely check it again. btw, is there a new API for reinitialvalue or formik is no longer support reinitialvalue?

jaripekkala:
https://github.com/jaredpalmer/formik/pull/1583#issuecomment-499374210

> In v2, reinitialize has be deprecated with a warning in favor of initialXYZ props. If defined, it keeps the v1 behavior, otherwise the behavior is disabled in favor of initialValues, initialErrors, and initialStatus. If you change those, then the form will reset. This also fixes initial dirty/pristine confusion.

I'm confused.

jaredpalmer:
If you pass true/false to `enableReinitalize`, it will work but will show a warning in v2.

ArnaudBarre:
> If you change those, then the form will reset. This also fixes initial dirty/pristine confusion.

So it sounds like enableReinitalize is true by default in v2 ?

jaripekkala:
It clearly isn't working on the latest release if thats so

https://codesandbox.io/embed/formik-codesandbox-template-fgk8x

jaredpalmer:
Actually I think I misspoke. The form will _not_ reset in v2.

ArnaudBarre:
So is there another way to achieved this behaviour without using a deprecated parameter ?

jaredpalmer:
I'M SORRY. I got confused when reading this too. Total Brain fart. enableReinitialize is not deprecated and works like v1. `isInitialValid` is being deprecated. My bad.

PR #1573, opened 2019-05-31
Optionally show multiple validation errors per field

mbrowne:
This PR adds a new prop called `validationSchemaOptions` to add the feature described here:
https://github.com/jaredpalmer/formik/issues/243

Example usage:

```js
  <Formik
    validationSchema={validationSchema}
    validationSchemaOptions={{ showMultipleFieldErrors: true }}
    ...
```

`showMultipleFieldErrors` is false by default, to maintain consistency with the current behavior of returning just a single string per field.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -10,6 +10,7 @@ import {
   FormikProps,
   FieldMetaProps,
   FieldInputProps,
+  ValidationSchemaOptions,
 } from './types';
 import {
   isFunction,
@@ -122,6 +123,7 @@ function useFormikInternal<Values = object>({
   isInitialValid,
   enableReinitialize = false,
   onSubmit,
+  validationSchemaOptions = { showMultipleFieldErrors: false },
   ...rest
 }: FormikConfig<Values>) {
   const props = { validateOnChange, validateOnBlur, onSubmit, ...rest };
@@ -202,12 +204,12 @@ function useFormikInternal<Values = object>({
             resolve(emptyErrors);
           },
           (err: any) => {
-            resolve(yupToFormErrors(err));
+            resolve(yupToFormErrors(err, validationSchemaOptions));
           }
         );
       });
     },
-    [props.validationSchema]
+    [props.validationSchema, validationSchemaOptions]
   );
 
   const runSingleFieldLevelValidation = React.useCallback(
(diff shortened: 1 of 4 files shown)


Andreyco:
Do we really need an additional configuration prop for this? I believe you can set (or resolve errors while validating) like this.

```jsx
<Formik
    validate={values =>
        return {
            email: ['min length error', 'format error'],
        }
    }
/>
```

mbrowne:
@andreyco Yes, that works fine for `validate` callbacks. I should have mentioned that this PR is specifically about `validationSchema` using Yup. The internal logic only returns one error at a time (per field) when using Yup.

malerba423:
@mbrowne were you ever able to merge this in? 

thanks,
another user hoping for the same functionality

loganknecht:
Hello! @jaredpalmer I absolutely LOVE this library!

Is this, or something similar to this functionality arriving in the library soon? I would love to have the support and ability to provide context for multiple different errors at a time without having to re-declare everything in the `validate` function!

Thank you for all the hard work! Please keep it up!

T04435:
:+1:

slorber:
in the meantime, here's a userland solution to make this work today, based on the code of this PR: https://codesandbox.io/s/formik-example-forked-wxzl1?file=/index.js

kpebron:
@slorber 's code works but here is my updated version based on my specifications:

```
import { getIn, setIn } from "formik"; // remove validateYupSchema
import { AnySchema } from "yup";

/**
 * Transform Yup ValidationError to a more usable object
 */
function yupToFormErrors(yupError: any, validationSchemaOptions: any) {
  let errors: any = {};
  if (yupError.inner.length === 0) {
    return setIn(errors, yupError.path, yupError.message);
  }
  // if showMultipleFieldErrors is enabled, set the error value
  // to an array of all errors for that field
  if (validationSchemaOptions.showMultipleFieldErrors) {
    for (let err of yupError.inner) {
      let fieldErrors = getIn(errors, err.path);
      if (!fieldErrors) {
        fieldErrors = [];
      }
      // Push to array if not yet added
      if (!fieldErrors.includes(err.message)) {
        fieldErrors.push(err.message);
      }
      errors = setIn(errors, err.path, fieldErrors);
    }
  } else {
    for (let err of yupError.inner) {
      if (!errors[err.path]) {
        errors = setIn(errors, err.path, err.message);
      }
    }
  }
  return errors;
}

export const ValidateYupSchemaArrErrors = async (
  values: any,
  schema: AnySchema
) => {
  try {
    // change to .validate method so that we can add abortEarly
    await schema.validate(values, {
      abortEarly: false,
    });
    return {};
  } catch (e) {
    return yupToFormErrors(e, { showMultipleFieldErrors: true });
  }
};
```

This will solve the following:
1. in the same field different rule but the same message, it will merge as 1
    rule below will validate email + custom email validation
```
    .email("Please enter valid email")
    .matches(emailRule, { message: "Please enter valid email" })
```

2. on the original code, when you submit the form as blank, only required rule will be validated, meaning if field has value (which passes required rule) but fails on other rules, it will still pass

PR #1578, opened 2019-06-02
Run low-pri validation at a given field / path in schema

jaredpalmer:
This PR tweaks validation so that Formik runs low-priority validation at the field level. That means that only the slice of the Yup schema and/or a single `<Field validate>` will be executed for a given change/blur. During pre-submit (i.e. high-priority validation), the entire schema is validated against values and all field-level validate fns are run (this is the current behavior). (I think) Yup is smart enough now that to deal with dependent schema paths (`when` and `refs`) so this should be non-breaking. 

Continues v1 work by @jquense in #780

## Details
- Use Yup's `validateAt` to validate the schema at a given path (instead of the entire schema) during run low-priority validation. 
- Only execute a field's registered `<Field validate>`fn during low-priority validation (instead of running all registered field-level validation fns) 

## Next Steps

I rushed through this initially to get an MVP going. We may want to refactor some stuff to make things easier.


jaredpalmer:
@jquense does validate at properly degrade with .when?

jquense:
If you are using `validateAt` it will do the correct thing with `when`s e.g. it resolve the dependent values and then only run validation for the specific field. 

The only edge case is when field a change to field `foo` would invalidate field `bar`, but there isn't really anything `yup` can do in those cases, same as with plain `validate` functions for fields.  Theoretically `yup` could do _better_ than individual `validate()`s since it knows which fields may be invalidated, but I'm not sure what that api would look like. Happy to brainstorm tho, if you want that added benefit, otherwise the yup case here is the same as the non-yup case for that gotcha

jaredpalmer:
@jquense my solution to that is to use the function form of validationSchema anyways.

jquense:
nice yeah, I'd say that covers workarounds then? It's not a super common thing to deal with anyway, idk that there is a lot of value in yup trying to be smarter to handle it?

jaredpalmer:
Probably not. We should document this situation and make an example of the solution.

Smau5:
i would love to have some examples for this.

gauravkhuranacopart:
Can we have some example of it ?

estshy:
@jaredpalmer Could you share an example of how to use it? For me it seems both input and form level validations are executed together.

earksiinni:
For anyone coming in from Google, it appears that this PR was reverted by https://github.com/jaredpalmer/formik/commit/377c36e3d1fecc0f0556d0218a18725ad04a7505.

@jaredpalmer, what was the reason for the revert?

thexpand:
@jaredpalmer Any info on this?

rbong:
@earksiinni looks like the related MR references https://github.com/formium/formik/issues/1654 as the reason it was reverted

PR #1583, opened 2019-06-05
documentation - enableReinitialize - note suggestion

jwallet:
At work we had performance issue, we have a pretty big form that was having a hard time to load, around 10 seconds, and one of the main reason was the use of this prop, causing ~10 re-renders on mount, now only 1. so I recommand using this prop to debug your form and remove it after. btw, the other thing we did was adding more memoize functions and selectors (store redux). now the form loads in 2-3 seconds, which is great considering how many values it has.

edit: also, #181 this is an issue to have a prop that will keep the form `dirty` when `enableReinitialize` is triggered instead of reapplying `pristine`, could make it worse. It's useful to know for a big form if the user can leave the page, so by looking at the form status `pristine`. I think the solution is not using the `enableReinitialize`, and keeping at the same time the pristine state of the form.

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -275,8 +275,8 @@ Default is `false`. Control whether Formik should reset the form if
 `initialValues` changes (using deep equality).
 
 Note: `enableReinitialize` will break the form `initialValues`, 
-meaning the form won't be pristine anymore, but dirty. You shouldn't 
-render form until all props are fetched by using a Loader, for instance. 
+meaning the form won't stay pristine. You shouldn't 
+render the form until all props are fetched (by using a Loader, for instance). 
 Also, remember that `enableReinitialize` will re-render your form and all 
 chidren components when the props change, so try not using it in production 
 if you are not sure how to use it.
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -274,6 +274,13 @@ don’t use both in the same `<Formik>`.
 Default is `false`. Control whether Formik should reset the form if
 `initialValues` changes (using deep equality).
 
+Note: `enableReinitialize` will break the form `initialValues`, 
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Would love to help you optimize this form. 2-3 seconds is crazy slow. How big is it?!?!

jaredpalmer:
Ahh yes. This is true. I guess I never thought folks would do otherwise. As you have pointed out, you should to render Formik only when all your props are defined and ready/loaded.

jaredpalmer:
Formik was actively developed against both CSR and SSR apps internally, so we probably overlooked describing some of these fundamental assumptions. In v2, reinitialize has be deprecated with a warning in favor of initialXYZ props. If defined, it keeps the v1 behavior, otherwise the behavior is disabled in favor of initialValues, initialErrors, and initialStatus. If you change those, then the form will reset. This also fixes initial dirty/pristine confusion.

jwallet:
> Would love to help you optimize this form. 2-3 seconds is crazy slow. How big is it?!?!

what I meant by 2-3 secs, it's now, it takes the right amount of time to load, which is the time needed to perform the api calls before displaying the form, the form loads quite nicely. And Yeah, the form is big but it's not open-source, we use formik at work. thanks again btw for that, way better than redux-form. We switched completely from it to yours in the last year. 

So the re-rendering was causing the form to act and render strangely, but it may be also react, so it was having a hard time to get a proper stable render.
 

> Ahh yes. This is true. I guess I never thought folks would do otherwise. As you have pointed out, you should to render Formik only when all your props are defined and ready/loaded.

Yes, so that's why I wanted to add a note to the doc. Often, a dev first look at the doc, and if the info is not there, he asks around.

jaredpalmer:
I think an async data example (or three) would be useful

Issue #1577, opened 2019-06-02
Use __DEV__ constant


Andreyco:
## 🚀 Minor codebase update

Use `__DEV__` constant instead of `process.env.NODE_ENV === 'production'`

Do this change agains `next` branch

jaredpalmer:
This broke React Storybook. I think we need to alter the webpack config?

gunn4r:
I can do a PR to fix this. I got it working locally right now.

gunn4r:
Opened and issue #1607 and a PR #1608.

👍 :)

PR #1594, opened 2019-06-09
Update to use __DEV__ constant

Rahmon:
Change from `process.env.NODE_ENV === 'production'` to `__DEV__`.

Closes: #1577

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict'
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./formik.cjs.production.js');
-} else {
+if (__DEV__) {
   module.exports = require('./formik.cjs.development.js');
+} else {
+  module.exports = require('./formik.cjs.production.js');
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -112,6 +112,9 @@
     "semi": true
   },
   "jest": {
+    "globals": {
+      "__DEV__": true
```
Andreyco: `boolean` here, not `true`

Rahmon: I'll fix this.

Rahmon: I've done a new PR: #1596

index.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./formik.cjs.production.js');
-} else {
+if (__DEV__) {
```
TrySound: Will this file appear in node_modules as is?

TrySound: /cc @jaredpalmer

Andreyco: it is replaced by `process.env.NODE_ENV === 'production'`

But as you can see, it is too verbose to type in every time

TrySound: Ah, ok.

PR #1597, opened 2019-06-10
Add ability to debounce low-priority validation

jaredpalmer:
- Debounce low-priority validation by 400ms by default
- Add a new prop called `debounceValidationMs` to `useFormik` / `formik`.
- Update docs and new sections to validation guide
- Add example with 200 controlled fields to storybook

With this change, Formik gets a huge performance boost on large forms / underpowered devices. Woo! 

**To Do**

We should figure out the default debounce delay. 400ms was my guess, but it would be good to have some evidence around this number. I did a quick google search:

| Time | Source |
|------|--------------------------------------------------------------------------------------|
| 250 | https://stackoverflow.com/questions/42361485/how-long-should-you-debounce-text-input |
| 500 | https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be |
| 400 | https://lawsofux.com/doherty-threshold |


jaredpalmer:
cc @FredyC

FredyC:
All considered I think a lower number (250) makes more sense so people are not too suprised by an out of box behavior. Once they get hang of Formik, they will surely find out about this option and possibly increase if their app requirements allows it.

examples/TwoHundredFields.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { Formik, FastField, Form } from 'formik';
+import { unstable_trace as trace } from 'scheduler/tracing';
+
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+const onSubmit = async values => {
+  await sleep(300);
```
FredyC: Never seen this as a good idea to deliberately slow down tests, even if it's for a bunch of ms. Jest timer mocks are probably a better option. Besides, it feels like some arbitrary number that can brake tests.

examples/TwoHundredFields.js
```diff
@@ -0,0 +1,48 @@
...
+  const initialValues = keys.reduce((prev, curr) => {
+    prev['field' + curr] = '';
+    return prev;
+  }, {});
+  return (
+    <Formik onSubmit={onSubmit} initialValues={initialValues}>
+      <Form>
+        <h1>Formik v2.x - {numFields} controlled fields</h1>
+        {keys.map(key => (
+          <FastField name={`field${key}`} key={key}>
```
FredyC: Um, what's the point of using FastField? It's just renamed export of Field, right? :) It's probably misleading that regular Field would be bad for this use case.

src/Formik.tsx
```diff
@@ -316,6 +323,13 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     [runAllValidations, state.values]
   );
 
+  // Create a debounced version of low-priority validation.
+  // @see https://lodash.com/docs/4.17.11#debounce
+  const debouncedValidateFormWithLowPriority = React.useCallback(
+    debounce(validateFormWithLowPriority, props.debounceValidationMs),
+    [validateFormWithLowPriority]
```
FredyC:
```suggestion
    [validateFormWithLowPriority, props.debounceValidationMs]
```

src/Formik.tsx
```diff
@@ -460,7 +474,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     (values: Values) => {
       dispatch({ type: 'SET_VALUES', payload: values });
       return validateOnChange
-        ? validateFormWithLowPriority(state.values)
+        ? debouncedValidateFormWithLowPriority(state.values)
```
FredyC: This is really an obvious change, dispatch will never update `state` object synchronously.
```suggestion
        ? debouncedValidateFormWithLowPriority(values)
```

examples/TwoHundredFields.js
jaredpalmer: It's coming back. I deleted it but just didn't change the example

examples/TwoHundredFields.js
jaredpalmer: It's not a test. It's an example.

namchaubonbe:
Formik

jaredpalmer:
You mean validation in setTouched?

FredyC:
Ouch, sorry for a confusing comment, that has confused me as well :) Yes, I think the `setTouched` validation should be debounced same way as `setValues` is.

jaredpalmer:
Agreezy

jonnyasmar:
@jaredpalmer Just curious if there is a roadmap for wrapping this up & getting it merged in and if we can expect the change to bubble up to v2's current RC?

fdiogo:
I'm sorry to be commenting on a closed MR, but I don't think there's a better place to discuss this but here.

We (where I work) have been using a modified version of this MR in production for several months now and it has been working wonderfully for us. I can tell you that the performance benefits of this change completely saved our butts on some complex forms we have.

We had some issues using this MR **as is** mainly due to how the `setValues` and `setTouched` are chained together in some cases. More specifically, when `setTouched` ran it was validating the form with stale values.

The way **we** fixed it was by essentially debouncing `setTouched` the same way as in `setValues`, just as @FredyC suggests. We also added a queue of promises awaiting for the debounced validations and a `useRef` to the state so that the validations would always run with the latest values.

The point of this comment is to understand if you would, in any way, be interested in what we have done. I don't know if this feature was simply dropped for some other reasons we're not aware of.

Please let me know if I could be of any use here @jaredpalmer

Cheers

MaximSagan:
@jaredpalmer Was there any reason this didn't end up getting merged? This would be of tremendous value for those of us with complex forms.

onionhammer:
Just stumbled across this - would love to be able to use it!

kurt-o-sys:
would like to have this as well, also wondering it wasn't merged and deleted, @jaredpalmer ?

konrad-garus:
+1, async validation without debounce is Bad.

conioX:
why this not merged? this feature really great for fieldarray

mwanago:
@jaredpalmer Do you have some plans on merging this one?
Cheers :)

PR #1617, opened 2019-06-19
Update type of resetForm to accept Partial<S>

jaredpalmer:
**Before**
Calling `resetForm` without passing a full state object yielded this TS error:

```Type '{ values: { email: string; }; }' is missing the following properties from type 'FormikState<{ email: string; }>': errors, touched, isSubmitting, isValidating, submitCount  TS2345```

**After**
No error when passing partial state object to `resetForm`


jaredpalmer:
Cutting a new rc

PR #1635, opened 2019-06-24
Update arrays.md

jsonz1993:
An example where the initial value is an array case has an error, it miss a `Form` component
`React.Children.only expected to receive a single React element child.`

-----
[View rendered docs/guides/arrays.md](https://github.com/jsonz1993/formik/blob/jsonz1993-update-docs/docs/guides/arrays.md)


billfienberg:
Confirmed:
https://codesandbox.io/embed/broken-basic-array-formik-example-36e2w

PR #1637, opened 2019-06-26
submitForm should not break promise chain

### LINKED ISSUE #1636: Internal `submitForm` implementation does not throw errors
## 🚀 Feature request
Catch errors thrown  from `onSubmit` inside of form/fields components

### Current Behavior
Formik Internal implementation of `submitForm` does not throw any errors.
It is not possible to get in `catch` hook with the following example:
```javascript
const onSaveButtonClick = async () => {
        // submitForm - formik internal method. It invokes `onSubmit` method
        //which is passed with props to formik
        const { submitForm } = props.formikActions;
        return submitForm()
          .then(promise => {
           // i'm always getting here
          })
          .catch((e: any) => {
           //cant get here
          });
      };
```
[Here is a sandbox exaple](https://codesandbox.io/s/formik-201-rc7-async-submitform-issue-k7fl9?fontsize=14)
### Desired Behavior
Have the ability to catch errors which are thrown from `onSubmit` method passed to formik as a prop

### Suggested Solution
```javascript
    // ...some code here
    return validateFormWithHighPriority().then(
        // ...some code here
          return Promise.resolve(executeSubmit())
            .then(() => {
              if (!!isMounted.current) {
              }
            })
            .catch(_errors => {
              if (!!isMounted.current) {
              }
            // THROW ERROR HERE
            });
        // ...some code here
        return;
      }
    );
  }, [executeSubmit, validateFormWithHighPriority]);
```

This is a cool additional feature for all users. In this way, internal `submitForm` method (which could be used inside <Field /> component ) will completely implement Promise interface.

andrewMuntyan:
`submitForm` helper should not break promise chain if handleSubmit has returned rejected Promise

Close #1636

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -658,13 +658,11 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         if (isActuallyValid) {
           return Promise.resolve(executeSubmit())
             .then(() => {
-              console.log('than inside formik');
               if (!!isMounted.current) {
                 dispatch({ type: 'SUBMIT_SUCCESS' });
               }
             })
             .catch(_errors => {
-              console.log('catch inside formik');
               if (!!isMounted.current) {
                 dispatch({ type: 'SUBMIT_FAILURE' });
                 throw _errors instanceof Error ? _errors : new Error(_errors);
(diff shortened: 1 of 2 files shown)


src/Formik.tsx
```diff
@@ -665,6 +665,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
             .catch(_errors => {
               if (!!isMounted.current) {
                 dispatch({ type: 'SUBMIT_FAILURE' });
+                throw _errors instanceof Error ? _errors : new Error(_errors);
```
slorber: Can't we just rethrow directly? If user throw weird things, he should receive weird things in catch imo

Issue #1640, opened 2019-06-27
Add `shouldValidate` parameter to `FormikHelper.setFieldValue()`


jt-helsinki:
## 🐛 Bug report

`FormikHelper.setFieldValue()` is missing the `shouldValidate: boolean` property. 

### Current Behavior

**Formik.tsx**
```
...
  const setFieldValue = useEventCallback(
    (field: string, value: any, shouldValidate: boolean = true) => {
      dispatch({
        type: 'SET_FIELD_VALUE',
        payload: {
          field,
          value,
        },
      });
      return validateOnChange && shouldValidate
        ? validateFormWithLowPriority(setIn(state.values, field, value), field)
        : Promise.resolve();
    },
    [validateFormWithLowPriority, state.values, validateOnChange]
  );
...
```

**types.tsx.FormikHelper**
```
...
export interface FormikHelpers<Values> {
...
    setFieldValue(field: keyof Values & string, value: any): void;
...
}
...
```

### Suggested solution(s)

Add the parameter to the function. 

**types.tsx.FormikHelper**
```
...
export interface FormikHelpers<Values> {
...
    setFieldValue(field: keyof Values & string, value: any, shouldValidate: boolean): void;
...
}
...
```

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.1 RC7
| React            | 16.8.6
| TypeScript       | 3.5.2
| Browser          | Chrome
| npm/Yarn         | Yarn & NPM
| Operating System | OSX

PR #1643, opened 2019-06-28
Missing shouldValidate parameter from typings

dagstuan:
The `FormikHelpers`-type is currently missing the parameter `shouldValidate?: boolean` from `setFieldValue` and `setFieldTouched`.


dagstuan:
Fixes #1640

PR #1644, opened 2019-06-28
Docs: Remove duplicate conditions

billfienberg:
Before: 
```
{errors.email && touched.email && errors.email}
{errors.password && touched.password && errors.password}
```

After
```
{errors.email && touched.email}
{errors.password && touched.password}
```

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -149,7 +149,7 @@ const Basic = () => (
             onBlur={handleBlur}
             value={values.email}
           />
-          {errors.email && touched.email && errors.email}
+          {errors.email && touched.email}
           <input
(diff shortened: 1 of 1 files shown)


jaredpalmer:
This will just return undefined. We want the duplicate condition here because we want to return the value of errors or undefined when the first conditions are true

billfienberg:
> This will just return undefined. We want the duplicate condition here because we want to return the value of errors or undefined when the first conditions are true

https://media3.giphy.com/media/3orifivhfTnPglD8as/giphy.gif

Thanks for the explanation, and I apologize for the unsolicited PR.

Issue #1650, opened 2019-07-04
Field-level validation not working when using `useField` hook


bummey:
### Current Behavior

When using the `useField` hook, field-level validation using the `validate` property doesn't work.
```ts
const [field] = useField({ name, validate: () => 'not good' });
```

### Expected behavior

The validate prop should work just like its `Field` counterpart.

### Reproducible example

https://codesandbox.io/embed/formik-codesandbox-template-p5zc1

### Suggested solution(s)

Manually registering the validate function does the trick for me:
```ts
useEffect(() => {
  registerField(name, {
    validate: () => 'not good'
  });

  return () => unregisterField(name);
}, [name, registerField, unregisterField]);
```

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.1-rc.9
| React            | 16.8.6
| TypeScript       | 3.5.2
| Browser          | Chrome 75
| npm/Yarn         | npm
| Operating System | Ubuntu 19.04

jaredpalmer:
Ahh yes, good catch

FredyC:
On that point, I think it makes sense to register field only if `validate` function is specified. Or is there some other plan for a field registry besides validation?

gaelduplessix:
This is fixed by https://github.com/jaredpalmer/formik/pull/1699, released in `v2.0.1-rc.13`

avetisk:
> This is fixed by #1699, released in `v2.0.1-rc.13`

Doesn't seem to be fixed still in `2.1.4` :'(

johnrom:
@avetisk this works fine for me. Are you able to reproduce in CodeSandbox?

avetisk:
> @avetisk this works fine for me in v2.1.4. Are you able to reproduce in CodeSandbox?

I moved back to RFF, but I still tried to recreate the issue on CodeSandbox and couldn't, so I guess I missed something there in my code.

So no worries, it works fine! :)

johnrom:
Awesome, @bummey please have a look and see if this fixes your issue. Thanks!

bummey:
Works fine now, thanks!

mordechaim:
Why is this behavior not documented [here](https://jaredpalmer.com/formik/docs/api/useField)?

I ran into this page by searching how to do field-level validation using hooks.

johnrom:
@mordechaim you're free to open a PR to update the documentation

paillave:
[...]

PR #1655, opened 2019-07-08
Consolidate types for field validate fn

FredyC:
Tiny TypeScript change as the validation function had two different signatures. Run into because of #1650 and had to call `registerField` manually. Besides, it's useful to export that signature for purposes of passing a validation function along the user code.


jaredpalmer:
Rename to `FieldValidator`

jaredpalmer:
Nice

PR #1656, opened 2019-07-08
Change getIn behavior so it returns `def` for empty `path` argument

### LINKED ISSUE #1038: getIn returns given value instead of fallback value when path is undefined
### Current Behavior

If you use the following code snippet `getIn(this.props.formik.values, this.props.name, undefined)` and `this.props.name` happens to be undefined it returns the value of `this.props.formik.values` (here `{}`) instead of `undefined`

### Expected behavior

I would expect that `getIn` method returns the fallback value `undefined` when the path is `undefined`

### Reproducible example

Reproducible by using: 
`getIn(this.props.formik.values, undefined, undefined)`

### Suggested solution(s)

I would expect an exit early check in `getIn` would solve this problem e.g. `if (!path) { return def }`

### Additional context

If you want to use `getIn` to check if a value at the given path in the object is available it will fail not as expected. 

### Your environment

| Formik           |1.3.1
| React            |16.5.2
| TypeScript       |N/A
| Browser          | 70.0.3538.67
| npm/Yarn         |1.10.1
| Operating System | MacOS 10.14 (18A391)

RusinovAnton:
closes #1038 

AR: 
```js
getIn({ errors: { name: 'John' } }, '') // -> { errors: { name: 'John' } }
```

ER: 
```js
getIn({ errors: { name } }, '') // -> undefined
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -68,7 +68,10 @@ export function getIn(
   def?: any,
   p: number = 0
 ) {
+  if (!key || !key.length) return def;
   const path = toPath(key);
+  if (!path || !path.length) return def;
+
   while (obj && p < path.length) {
     obj = obj[path[p++]];
   }
--- a/test/utils.test.tsx
+++ b/test/utils.test.tsx
@@ -4,6 +4,7 @@ import {
   isPromise,
   getActiveElement,
   isNaN,
+  getIn,
 } from '../src/utils';
 
 describe('utils', () => {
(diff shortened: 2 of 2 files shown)


jaredpalmer:
Should we just throw an error?

RusinovAnton:
@jaredpalmer 
I would say **no**, unless we want to bring breaking changes in. 

The lodash's "get" behavior is to return the default value or `undefined`:
```
_.get({ errors: { name } }, '') // -> undefined
_.get({ errors: { name } }, '', null) // -> null
```

I'm not saying that Formik's `getIn` should mimic lodash's `get` behavior but its something that people are used to.

For example, I'm using this package
https://github.com/stackworx/formik-material-ui

and the reason, I'm trying to fix `getIn`'s behavior, is because Im getting a weird error of object being passed as React's component child if I'm not to setting TextField's "name" attribute because this is how `getIn` being used there:

```jsx
 const fieldError = getIn(errors, name);
```

https://github.com/stackworx/formik-material-ui/blob/master/src/TextField.tsx#L20

PR #1662, opened 2019-07-09
WIP: Opinionated throttled + debounced validation

jaredpalmer:
This adds the following opinionated, yet (probably more optimal) technique for validation:

**If an update for a field value is shorter than 5 characters or ends in a space, then run a throttled low-pri validation. Otherwise, run a debounce low-pri validation.**

Then end result is more immediate feedback on short inputs and more regimented feedback on longer inputs. 

Obviously, this is super-opinionated, but I just wanted to start a conversation. In a perfect world, this would be handled in user-land. However, the current API makes implementing something like this at the form-level very hacky. This raises another key question, which is whether this should be at the form-level or handled at the per-field level.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -25,6 +25,7 @@ import { FormikProvider } from './FormikContext';
 import invariant from 'tiny-warning';
 import { LowPriority, unstable_runWithPriority } from 'scheduler';
 import debounce from 'lodash/debounce';
+import throttle from 'lodash/throttle';
 // We already used FormikActions. So we'll go all Elm-y, and use Message.
 type FormikMessage<Values> =
   | { type: 'SUBMIT_ATTEMPT' }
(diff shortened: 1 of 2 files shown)


src/Formik.tsx
```diff
@@ -499,11 +509,18 @@ export function useFormik<Values extends FormikValues = FormikValues>({
           value,
         },
       });
-      return validateOnChange && shouldValidate
-        ? debouncedValidateFormWithLowPriority(
+      if (validateOnChange && shouldValidate) {
+        if (value.length < 5 || value.endsWith(' ')) {
```
ivansky: Why 5? What a magic number? Let's make a constant to make this part a bit more clear.

FredyC: Or better yet, it should be configurable and not just number, but that space behavior. Ideally, some optional function that contains that logic by default, but can be overridden.

FredyC:
I think that we should focus on core problems first that are blocking release. Tuning the optimization can be always done later.

Nonetheless, what are you going to do with fields that work with an array, objects, boolean or number values? Such a condition will fail horribly. Not all forms are about collecting strings.

So if anything, it should be configurable per field. Form condition doesn't make much sense really.

jaredpalmer:
This was just an idea.

PR #1669, opened 2019-07-11
Allow to pass any type of value to fields' onChange/onBlur

### LINKED ISSUE #1667: [v2] Improve Formik.handleChange to accept not only string or event
## 🚀 Feature request
I'm trying to use `useField` hook to connect formik custom component:

My custom component works with object value, e.g.:
```javascript
const CustomComponent = ({ value, onChange }) => (
  <button onClick={() => onChange({ some: 'value' })}>{value.some}</button>
)
```

### Current Behavior
I'm getting `Cannot read property 'type' of undefined` error when I connect my custom component to formik like this:
```javascript
const FormikCustomComponent = (props) => {
  const [field] = useField(props.name);
  return (
    <CustomComponent {...props} {...field} />
  );
};
```

Error thrown because my `CustomComponent` sends to `onChange` object instead of string or event. `Formik.handleSubmit` checks if passed value is string (it is not) and tries to treat it as an event (and of course fails).

### Desired Behavior
Add ability to pass to `onChange` callback any desired value and use it "as is". `handleChange` should try to parse value ONLY when EVENT is passed as a value. 

### Suggested Solution
Currently we check if `eventOrTextValue` is string, and if it is not a string, we treat it as event. 
(https://github.com/jaredpalmer/formik/blob/2.0/src/Formik.tsx#L290)
My suggestion is to check if value is EVENT instead of string. If event is passed, then, you need to run code to extract value from the event. In other cases just use passed value "as is"
Example solution: 
```typescript
const executeChange = (
  eventOrTextValue: string | React.ChangeEvent<any>,
  maybePath?: string
) => {
  let field = maybePath;
  let val = eventOrTextValue;
  let parsed;
  // If the first argument is a synthetic React Event or a real Dom Event,
  // we handle like we would a normal HTML change event.
  if (eventOrTextValue && (eventOrTextValue instanceof Event || eventOrTextValue.nativeEvent instanceof Event)) {
    // no changes here ...
  }

  // no changes here...      
};
```

### Possible workarounds
Of course, I'm able to `JSON.stringify` value before passing it to `onChange` and `JSON.parse` it back in component to be able to work with any type of value, but I think it is not a good idea...

It will make much easier to connect formik to custom component that works with objects / arrays as a value.

afitiskin:
Fixes #1667

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -81,7 +81,7 @@ export function useField<Val = any>(
   }
 
   if (isObject(propsOrFieldName)) {
-    if (process.env.NODE_ENV !== 'production') {
+    if (__DEV__) {
       invariant(
         (propsOrFieldName as FieldAttributes<Val>).name,
         'Invalid field name. Either pass `useField` a string or an object containing a `name` key.'
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -172,7 +172,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
               resolve(errors || emptyErrors);
             },
             actualException => {
-              if (process.env.NODE_ENV !== 'production') {
+              if (__DEV__) {
                 console.warn(
                   `Warning: An unhandled error was caught during validation in <Formik validate />`,
                   actualException
@@ -217,7 +217,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
               resolve(yupToFormErrors(err));
             } else {
               // We throw any other errors
-              if (process.env.NODE_ENV !== 'production') {
+              if (__DEV__) {
                 console.warn(
                   `Warning: An unhandled error was caught during validation in <Formik validationSchema />`,
                   err
@@ -493,20 +493,24 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   const executeChange = React.useCallback(
-    (eventOrTextValue: string | React.ChangeEvent<any>, maybePath?: string) => {
+    (eventOrValue: any | React.ChangeEvent<any>, maybePath?: string) => {
       // By default, assume that the first argument is a string. This allows us to use
       // handleChange with React Native and React Native Web's onChangeText prop which
       // provides just the value of the input.
       let field = maybePath;
-      let val = eventOrTextValue;
+      let val = eventOrValue;
       let parsed;
       // If the first argument is not a string though, it has to be a synthetic React Event (or a fake one),
       // so we handle like we would a normal HTML change event.
-      if (!isString(eventOrTextValue)) {
+      if (
+        eventOrValue &&
+        (eventOrValue instanceof Event ||
+          eventOrValue.nativeEvent instanceof Event)
+      ) {
         // If we can, persist the event
         // @see https://reactjs.org/docs/events.html#event-pooling
-        if ((eventOrTextValue as React.ChangeEvent<any>).persist) {
-          (eventOrTextValue as React.ChangeEvent<any>).persist();
+        if ((eventOrValue as React.ChangeEvent<any>).persist) {
+          (eventOrValue as React.ChangeEvent<any>).persist();
         }
         const {
           type,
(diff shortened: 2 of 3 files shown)


src/Field.tsx
```diff
@@ -81,7 +81,7 @@ export function useField<Val = any>(
   }
 
   if (isObject(propsOrFieldName)) {
-    if (process.env.NODE_ENV !== 'production') {
+    if (__DEV__) {
```
afitiskin: I'm not sure which way is better: to use `process.env.NODE_ENV !== 'production'` or `__DEV__`, but I think we need to keep only one of them. I prefer `__DEV__`

src/Formik.tsx
```diff
@@ -578,32 +582,47 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   const executeBlur = React.useCallback(
-    (e: any, path?: string) => {
-      if (e.persist) {
-        e.persist();
+    (eventOrValue: any | React.FocusEvent<any>, maybePath?: string) => {
```
afitiskin: I changed `handleBlur` / `executeBlur` as well, since custom components may not to pass real blur event, but may pass value instead. Also I made pairs of functions `handleBlur`/`handleChange` and `executeBlur`/`executeChange` more similar (same type, same name of variables, etc)

test/Field.test.tsx
```diff
@@ -121,12 +121,11 @@ describe('Field / FastField', () => {
...
       );
 
-      const { handleBlur, handleChange } = getFormProps();
       injected.forEach((props, idx) => {
         expect(props.field.name).toBe('name');
         expect(props.field.value).toBe('jared');
-        expect(props.field.onChange).toBe(handleChange);
-        expect(props.field.onBlur).toBe(handleBlur);
+        expect(typeof props.field.onChange).toBe('function');
+        expect(typeof props.field.onBlur).toBe('function');
```
afitiskin: Not the best test ever :)
IMO here we have to test, that calling `field.onChange`/`field.onBlur` actually updates form state.

src/Field.tsx
jaredpalmer: Use __DEV__

src/Formik.tsx
```diff
@@ -765,8 +784,8 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       const field: FieldInputProps<any> = {
         name,
         value: valueState,
-        onChange: handleChange,
-        onBlur: handleBlur,
+        onChange: (e: any) => executeChange(e, name),
```
jaredpalmer: just curry handleChange(name)

test/Field.test.tsx
jaredpalmer: don't do this here, do it in useField or

GoPro16:
Lmk if you need any help with this @afitiskin. Working on a `DateRange` component with `react-dates` and need this for it to work: [Code Link](https://github.com/Availity/availity-react/pull/167)

src/Formik.tsx
afitiskin: Yep, that was the first idea, but I'm not good at TS generic types, and I'm getting error type error when I'm using `onChange: handleChange(name)`...

GoPro16: Having a hard time understanding what generic type needs to change [here](https://github.com/jaredpalmer/formik/blob/master/src/types.tsx#L132-L134)? I am wanting this feature and was going to finish PR but also not a typescript 10X Engineer.

afitiskin: It's simply impossible with generics in current TS version :) (link to the TS issue is correct, we are experiencing exactly the same problem).
IMO we do not need to use `handleChange` here since it's just do the same thing as my code above

IMO 2 possible ways of usage `handleChange` is a kind of "magic" behaviour and it may leads to bugs. I prefer never try to get the name of the field from an `event` and always bind the name to handler.

jaredpalmer:
This is survived by #2255

PR #1695, opened 2019-07-22
Initial context should be undefined

FredyC:
With object being initial value, the invariant in the `useFormikContext` never fails and causes weird errors later when the hook is actually misused.


jaredpalmer:
This has been fixed.

Issue #1666, opened 2019-07-10
[v2] Rewrite part/all <Formik /> docs with useFormik and FromikContext or analogs


PetroPavlenko:
## 📖 Documentation
I would like to start replacing `withFormik` HOC with `useFormik` Hook. But I need some more docs with `useFormik`. So I won't use internal methods.

sw-yx:
```tsx
import React from 'react'
import { useFormik } from 'formik'

type Props = {
  label: string
  name: string
  type: string
  formik: {
    getFieldProps: ReturnType<typeof useFormik>['getFieldProps']
  }
}
const MyTextField = ({ label, formik, ...props }: Props) => {
  const [field, meta] = formik.getFieldProps(props)
  return (
    <>
      <label>
        {label}
        <input {...field} {...props} />
      </label>
      {meta.touched && meta.error ? <div className="error">{meta.error}</div> : null}
    </>
  )
}

const Example = () => {
  const formik = useFormik({
    initialValues: { email: '', firstName: 'red', lastName: '' },
    onSubmit: (values, actions) => {
      setTimeout(() => {
        alert(JSON.stringify(values, null, 2))
        actions.setSubmitting(false)
      }, 1000)
    },
  })
  return (
    <div>
      <h1>My Form</h1>
      <form onSubmit={formik.handleSubmit}>
        <MyTextField formik={formik} name="firstName" type="text" label="First Name" />
        <MyTextField formik={formik} name="lastName" type="text" label="Last Name" />
        <MyTextField formik={formik} name="email" type="email" label="Email" />
        <button type="submit">Submit</button>
      </form>
    </div>
  )
}
export default Example
```

pmonty:
I read Jared said that useFormik would not be exported. I am dying for official documentation!

maddhruv:
Closing due to long inactivity

johnrom:
[...]

PR #1717, opened 2019-07-29
document useFormik

sw-yx:
first stab at documenting `useformik`

addresses https://github.com/jaredpalmer/formik/issues/1666

also i want to show `getFieldProps` since we expect people to use that alot... also wanted to show how to type the things since it can trip up people

```tsx
import React from 'react'
import { useFormik } from 'formik'

type Props = {
  label: string
  name: string
  type: string
  formik: {
    getFieldProps: ReturnType<typeof useFormik>['getFieldProps']
  }
}
const MyTextField = ({ label, formik, ...props }: Props) => {
  const [field, meta] = formik.getFieldProps(props)
  return (
    <>
      <label>
        {label}
        <input {...field} {...props} />
      </label>
      {meta.touched && meta.error ? <div className="error">{meta.error}</div> : null}
    </>
  )
}

const Example = () => {
  const formik = useFormik({
    initialValues: { email: '', firstName: 'red', lastName: '' },
    onSubmit: (values, actions) => {
      setTimeout(() => {
        alert(JSON.stringify(values, null, 2))
        actions.setSubmitting(false)
      }, 1000)
    },
  })
  return (
    <div>
      <h1>My Form</h1>
      <form onSubmit={formik.handleSubmit}>
        <MyTextField formik={formik} name="firstName" type="text" label="First Name" />
        <MyTextField formik={formik} name="lastName" type="text" label="Last Name" />
        <MyTextField formik={formik} name="email" type="email" label="Email" />
      </form>
    </div>
  )
}
export default Example

```

-----
[View rendered docs/api/useFormik.md](https://github.com/sw-yx/formik/blob/patch-1/docs/api/useFormik.md)


PetroPavlenko:
@sw-yx Looks like it's better to use `useField` hook in MyTextField and use FormikContext.Provider

pmonty:
So is useFormik not going to be exposed? What is going to be exposed is my next question :)

Looking forward to a Version 2 release!

johnrom:
@sw-yx it doesn't look like this the purpose use of `useFormik`. The `Example` component would require a FormikProvider (React.Context Provider, provided by `<Formik />` component) above it in the hierarchy in order for this to work.

I believe what you're looking for is `useFormikContext`, which returns something similar to the `render` function from v1. As @PetroPavlenko wrote above, you can also use `useField` when you just want access to the field properties. If you really want to use `useFormik()`, you will need to manually build a context for Formik (untested, possibly unsupported?):

(note, I haven't been keeping up too much with v2 so I may give some wrong function names here)

```
// NOTE: You probably want `<Formik />` in the hierarchy instead!! 
// This does basically the same thing.
const FormikWrapper = (props) => {
  const formikBag = useFormik(props);
  return (
    <FormikProvider value={formikBag}>
        <MyForm />
    </FormikProvider>
  )
}

const MyForm = () => {
    const { handleSubmit, handleReset } = useFormikContext(); // grabs FormikProvider.Consumer, basically
    
    return (
            <MyInput name="email" />
            <button type="button" onClick={handleReset}>Reset Form</button>
        </form>
    );
}

const MyInput = ({ name }) => {
    const field = useField(name);

    <input name={field.name} value={field.value} onChange={field.onChange} />
}
```

sw-yx:
hmm ok. i'm happy to close this and document whatever ends up being actually recommended.

PR #1740, opened 2019-08-09
Recursively validateYupSchema. Fix #1520

### LINKED ISSUE #1520: Inconsistent error handling with nested data
### Current Behavior

Nested properties are showing up in the errors object for non-required fields after being touched, which is incorrect behavior in my opinion and inconsistent with how top level properties are handled. Let me explain with a screenshot and some code.

In the following screenshot, every field has been touched exactly once. The top two inputs correspond to top level object properties, while the last input corresponds to a property that is nested inside of an object.

Screen Shot 2019-05-17 at 11 25 29 AM

Here is the code that goes along with the screenshot.

```
const SampleSchema = object().shape({
  foo: string()
    .min(5, 'Too Short')
    .max(20, 'Too Long')
    .required('Required'),
  bar: string()
    .min(5, 'Too Short')
    .max(20, 'Too Long'),
  bif: object().shape({
    baz: string()
      .min(5, 'Too Short')
      .max(20, 'Too Long')
  })
})
```
```
<Formik
  initialValues={{
    foo: '',
    bar: '',
    bif: {
      baz: ''
    }
  }}
  validationSchema={SampleSchema}
>
```

As you can see, neither `bar` nor `baz` are required fields. However, after each field has been touched the top level property correctly skips running validations because the field is still empty, while the nested property incorrectly runs validations even though the field is empty and not required.

### Expected behavior

Validations should be skipped on a non-required nested property that has been touched but is still empty.

Top level properties are already handled correctly.

### Reproducible example

https://codesandbox.io/s/formik-codesandbox-template-ctu7h?fontsize=14

### Suggested solution(s)

I'm not entirely certain whether this is a bug with Formik, or if this is a bug with yup...

### Your environment

| Formik           | ^1.5.4
| Yup              | ^0.27.0
| React            | ^16.8.6
| Browser          | Chrome 74.0
| Yarn             | 1.15.2
| Operating System | macOS 10.14.4

Xstoudi:
I ran into #1520 while developing with Formik and fixed the issue by making the validateYupSchema preparing datas recursively with support of objects and arrays.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -934,7 +934,7 @@ function prepareDataForValidation<T extends FormikValues>(
             return value !== '' ? value : undefined;
           }
         });
-      } else if (typeof values[key] === 'object') {
+      } else if (typeof values[key] === 'object' && values[key] !== null) {
         data[key] = prepareDataForValidation(values[key]);
       } else {
         data[key] = values[key] !== '' ? values[key] : undefined;
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -899,7 +899,6 @@ export function yupToFormErrors<Values>(yupError: any): FormikErrors<Values> {
   }
   return errors;
 }
-
 /**
  * Validate a yup schema.
  */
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Dope.

src/Formik.tsx
```diff
@@ -909,17 +909,39 @@ export function validateYupSchema<T extends FormikValues>(
...
-      validateData[key] = values[key] !== '' ? values[key] : undefined;
+      if (Array.isArray(values[key]) === true) {
+        data[key] = values[key].map((value: any) => {
+          if (Array.isArray(value) === true || typeof value === 'object') {
+            return prepareDataForValidation(value);
+          } else {
+            return value !== '' ? value : undefined;
+          }
+        });
+      } else if (typeof values[key] === 'object') {
```
targos: should this also check for `null` ?

jaredpalmer: yes

Xstoudi: done in 246212fec48de4573ef57025662a60f5d669ab1d

Xstoudi:
I saw your request a little bit late, it's done in 246212fec48de4573ef57025662a60f5d669ab1d

alexkolson:
It would be so cool if this could make it in to the next version of formik :)

alexkolson:
This is merged and that makes me so happy! Thank you for making Formik and for making it even better!!!

sambrodriguez:
Not sure if anyone is still watching this, but why was the decision made to change any empty string values into undefined? We had a number of Yup tests that worked with the empty string values that break now. And fixing them is not a big deal, but I'd like to know why this change was made. Is there a problem with passing empty strings to Yup?

@Xstoudi

PR #1746, opened 2019-08-13
Feature/yup validation context

ldicocco:
This is for https://github.com/jaredpalmer/formik/issues/503 and takes again from https://github.com/jaredpalmer/formik/pull/522
It adds a validationSchemaContext property tha can be an object or a function taking the current values and returning an object. The object will be used as a Yup context in validation, allowing yup functions like when to refer to external values or field valus outside from siblings and descendants.

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -394,3 +394,9 @@ events and `change`-related methods. More specifically, when either
 [A Yup schema](https://github.com/jquense/yup) or a function that returns a Yup
 schema. This is used for validation. Errors are mapped by key to the inner
 component's `errors`. Its keys should match those of `values`.
+
+### `validationSchemaContext?: Record<string, any> | ((values. Values) => Record<string, any>)`
+
+An object or a function that gets the current field values and returns an object that is passed as a context to the validation Schema.
+It can be used to reference an external value in Yup `when` (must be prefixed with `$`) or, returning the passed values in the function form,
+to reach in Yup `when` outside of siblings and descendants.
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -200,10 +200,14 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         const schema = isFunction(validationSchema)
           ? validationSchema(field)
           : validationSchema;
+        const validationSchemaContext = props.validationSchemaContext;
+        const schemaContext = isFunction(validationSchemaContext)
+          ? validationSchemaContext(values)
+          : validationSchemaContext;
         let promise =
           field && schema.validateAt
-            ? schema.validateAt(field, values)
-            : validateYupSchema(values, schema);
+            ? schema.validateAt(field, values, { context: schemaContext })
+            : validateYupSchema(values, schema, undefined, schemaContext);
         promise.then(
           () => {
             resolve(emptyErrors);
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -206,6 +206,12 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
    * A Yup Schema or a function that returns a Yup schema
    */
   validationSchema?: any | (() => any);
+  /**
+   * An object or a function that returns an object that will be used as Yup context
+   */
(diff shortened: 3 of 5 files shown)


ldicocco:
As the addition of a schema.validateAt for field level validation makes the previous workaround for passing the validation context no longer equivalent to full validation context, I think that an explicit validation context solution is now an improvement.

jaredpalmer:
This is solid.

jaredpalmer:
@eonwhite do you have any issues with this?

silasdavis:
I have seen a previous PR addressing this that was closed, so just to add a voice in favour and a use case example:

Currently I am recreated my entire schema object on each component update in order to make use of a calculated password strength which is shared state amongst the form (for a separate password meter).

If I could just pass the context `{passwordStrength: strength}` then I can write base my password strength validation on yup's `when('$passwordStrength`, (strength, schema) => ...)`.

I realise I could wire up `validateYupSchema` directly which takes this argument. But it seems a shame to have the integration with yup not fully supporting context which allows for nice stateful validation.

shurygindv:
> to add a voice in favour and a use case example:

My case: there is react element: `<FormComponent isFieldRequired={isSomethingLogic} /> `depending on prop `isFieldRequired` I need to make N field in FormComponent required (dynamic validation), validationContext is great decision for this! Now, I use a "solution/workaround" from topic: https://github.com/jaredpalmer/formik/issues/503

Thank you.

jaredpalmer:
I believe the prior PR just passed in all props as context.

silasdavis:
I should have linked this in my original comment, but I think the PR in question was similar to this with a separate property for the context.

from: https://github.com/jaredpalmer/formik/pull/1508#issuecomment-497817074 



> I don't think this is necessary as you can accomplish this via function as described in this comment.
There you reference the following comment:
- @jaredpalmer 

The comment you referred to:

>I will add, that for now I mitigated this issue by passing schema creator to validationSchema, sth like
validationSchema={() => createSchema(context)}, but of course recreating schema especially a big one for every render is not good for performance, and thats why we have yup context

So my +1 was really about avoiding the schema recreation that this seems to do.

Is this PR awaiting a response from @eonwhite ?

Thanks for your efforts

ldicocco:
I've noticed in the comments that the new validationSchemaContext is referred to as validationContext, which probably makes more sense. In case this PR is approved, should I change the name? It takes just a few minutes to do so.

mikew:
Was just gonna start a PR with the same thing :)

`validationSchemaContext` makes more sense to me, because it exclusively works with `validationSchema`.

Anything holding this PR up? There are workarounds using `validate` + some helpers from Formik, but it's only form-level submission validation, not field level, which can introduce inconsistencies.

Nnoerregaard:
Is there any news as to when this PR will be released?

jrobeson:
it'd need to be rebased due to conflicts

mathieulj:
For my use case, I'd love to see this land with 
```md
### `validationSchemaContext?: Record<string, any> | ((values: Values, status?: any) => Record<string, any>)`
```

Or even better with
```md
### `validationSchemaContext?: Record<string, any> | ((state: FormikState<Values>) => Record<string, any>)`
```

ldicocco:
I've fixed the merge conflicts.

timojokinen:
what is the status here? will this be merged anytime soon?
my team ran into an issue that is going to be fixed with this pr

KarelVerschraegen:
Any updates on when this feature will be merged?

sandrooco:
@ldicocco Thanks for providing this PR. Please resolve the conflicts because we really need this feature. :)

AlashaHonore:
@ldicocco Waiting for this feature too :)

tastafur:
@ldicocco @jaredpalmer Waiting for this feature me too :)

gat-bryszard:
We're also waiting for the feature :).

bryszard:
We're also waiting for the feature. Thanks in advance for merging it :).

Crocsx:
so ? Any Update ?

PR #1747, opened 2019-08-14
[v2] Add useForm hook as alternative to <Formik /> render prop

qw-in:
tl;dr

```jsx
function MyForm() {
  const [Form, { handleChange, values }] = useForm({
    onSubmit: console.log,
    initialValues: { name: 'hi' }
  });

  return (
    <Form>
      <input name="name" value={values.name} onChange={handleChange('name')} />
      <button type="submit">submit</button>
    </Form>
  );
}
```

_Sorry if I missed a discussion on this sort of hook somewhere along the way...had a quick search around but couldn't find anything_

### Overview
Added a `useForm` hook that provides an alternative (public) api to the `<Formik />` render prop we have all been using forever. My main motivation for this is that I have found myself writing the below pattern in a lot of projects when I have other hooks I want to use within the form (loading indicators, user feedback, etc). For larger forms it's no problem but for simple forms with just a couple of fields it can feel a bit boilerplate heavy

```js
function InternalMyForm(formik) {
  useMyHookThatDependsOnFormikState(formik);

  const { handleChange, values } = formik;

  return (
    <Form>
      <input name="name" value={values.name} onChange={handleChange('name')} />
      <button type="submit">submit</button>
    </Form>
  );
}

function MyForm() {
  return (
    <Formik>
      {(formik) => <InternalMyForm {...formik} />}
    </Formik>
  );
}
```

Now I know you are all thinking - "just use `withFormik`!" - but come on, HOCs in 2019?

### Considerations
1. I initially hijacked the internal `useFormik` hook used by `<Formik />` and just modified the API to what `useForm` is now. However that felt a bit weird and I liked `useForm` better in the end. It also means that this won't break code using the internal hook
1. I started by returning a fake `<Formik />` component from the hook that just wrapped the context provider. Returning the wrapped `<Form />` just felt cleaner in the end though and for some reason seems potentially less confusing

### Test
I copied over the `<Formik />` tests and modified the render function to suit the hook. I did remove ~3 cases but that was because they were specifically dealing with `<form>`less cases

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -8,3 +8,4 @@ export * from './types';
 export * from './connect';
 export * from './ErrorMessage';
 export * from './FormikContext';
+export * from './useForm';
--- a/src/useForm.tsx
+++ b/src/useForm.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { useFormik } from './Formik';
+import { FormikFormProps } from './Form';
+import { FormikValues, FormikConfig, FormikContext } from './types';
+import { FormikProvider } from './FormikContext';
+
+export function useForm<Values extends FormikValues = FormikValues>(
+  config: FormikConfig<Values>
+) {
+  const formikProps = useFormik<Values>(config);
+
+  const Form = (p: FormikFormProps) => (
+    <FormikProvider value={formikProps}>
+      <form
+        onSubmit={formikProps.handleSubmit}
+        onReset={formikProps.handleReset}
+        {...p}
+      />
+    </FormikProvider>
+  );
+
+  const ret: [typeof Form, FormikContext<Values>] = [Form, formikProps];
+
+  return ret;
+}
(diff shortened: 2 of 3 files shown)


Superpencil:
> Now I know you are all thinking - "just use withFormik!" - but come on, HOCs in 2019?
This!

jaredpalmer:
In v2 you can use useFormik().

```js
import React from 'react'
import { useFormik } from 'formik'

function MyForm() {
  // setup formik the OG way without using React context provider
  const formik = useFormik({
     initialValues: { ... },
     validationSchema: {},
     onSubmit: ()=> {},
  });

 useMyHookThatDependsOnFormikState(formik);

 return (
    <form onSubmit={formik.handleSubmit}>
      <input {...formik.getFieldProps('name')} />
      <button type="submit">submit</button>
    </form>
  );
}
```

PR #1763, opened 2019-08-21
Refactor useEventCallback() with React.useLayoutEffect()

### LINKED ISSUE #1762: Rewrite `useEventCallback()`
`useEventCallback()` can be written as

```typescript
function useEventCallback<T extends (...args: any[]) => any>(
  fn: T
): T {
  const ref: any = React.useRef();

  // we copy a ref to the callback scoped to the current state/props on each render
  React.useLayoutEffect(() => {
    ref.current = fn;
  });

  return React.useCallback((...args) => ref.current.apply(void 0, args), []) as T;
}
```

as per https://github.com/facebook/react/issues/14099#issuecomment-440172008

To me this would make it more clear that `useEventCallback()` is essentially a hack to have a callback which has the entire current state/props, and not only the deps, but the identity of the callback itself doesn't change, making it safe to send downstream and not cause rerenders.

That is to say, even the current version does essentially this, but `useLayoutEffect()` makes it more explicit than `useEffect()` + deps + `throw on incomplete render`.

*Edit* Another thing that falls off from this change is, now whenever `useEventCallback()` is used, it looks weird, because it has no deps. And, I think it's good to look weird, because it /is/ weird. It makes it more clear that this isn't a normal hook.

_Originally posted by @VanTanev in https://github.com/jaredpalmer/formik/pull/1566#issuecomment-497933241_

jaredpalmer:
Close #1762. Rewrite `useEventCallback` with `React.useLayoutEffect`.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -312,8 +312,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
           return combinedErrors;
         });
       });
-    },
-    [runAllValidations, state.values]
+    }
   );
 
   // Run all validations methods and update state accordingly
@@ -329,8 +328,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         }
         return combinedErrors;
       });
-    },
-    [state.values, state.errors, runAllValidations]
+    }
   );
 
   const resetForm = React.useCallback(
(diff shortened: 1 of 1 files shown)


FredyC:
Well, I would say that doing it with `useLayoutEffect` is just an ugly hack. Are you aware it will be throwing warnings in SSR which cannot use this hook?

FredyC:
You can get inspired in neighborhood ... https://github.com/joepuzzo/informed/pull/221

jaredpalmer:
Good looka

--
Jared Palmer
jared@skiptu.com
+1 917-952-2960

PR #1782, opened 2019-08-29
Fix FieldArray add/remove operations and validateOnChange behavior

inv8der:
**Formik version:** 2.0.1-rc.13
**Example:** https://codesandbox.io/s/formik-fieldarray-example-v2-6q9v8?fontsize=14

Using FieldArray, whenever a value is added (insert/unshift) or removed (remove/pop) and the form is currently valid, I discovered it would become invalid after the operation completes. It turns out these operations add to the errors/touched state unnecessarily in some cases. By default, the `validateOnChange` prop should cause form validations to run again, which I think would resolve the issue. However, this doesn't happen since `setFormikState` doesn't actually accept a callback function.

This pull request should hopefully fix these issues. I'm having trouble getting `yarn link formik` to work properly in my project, so I haven't been able to verify.


src/FieldArray.tsx
```diff
@@ -140,9 +140,9 @@ class FieldArrayInner<Values = {}> extends React.Component<
             : prevState.touched,
         };
       },
-      () => {
+      (state: FormikState<any>) => {
         if (validateOnChange) {
-          validateForm();
+          validateForm(state.values);
```
inv8der: Since `validateForm` is just an alias for `validateFormWithHighPriority`, there's a chance that `SET_ERRORS` won't be dispatched if validation results are equal to the current state. Any changes made using `setFormikState` will not have updated state yet.

jaredpalmer:
See #2115

PR #1786, opened 2019-08-31
Improve Field typings

Carl-Foster:
I was initially just going to fix up the use of `any` as that what the props of `Field` was, but I've also added in the inference of the props any component that you pass in through the `component` or `as` prop. This has worked quite well in a little test where a custom `Select` component requires `Array<Option>` and didn't want to have to use the children props to enforce it. It also now strongly types the HTMLAttributes when a string is passed in for the `component` or `as`.

In addition, this will restrict using both `render` and `children`.

Happy for improvements where wanted.

-----
[View rendered .github/CONTRIBUTING.md](https://github.com/Carl-Foster/formik/blob/fix/field-types/.github/CONTRIBUTING.md)

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -221,12 +221,8 @@ export const Field = <A extends PickOneConfig>({
         children
       );
     }
-    const Component = component;
-    return (
-      <Component {...legacyBag} {...props}>
-        {children}
-      </Component>
-    );
+
+    return React.createElement(component, { ...legacyBag, ...props }, children);
   }
 
   // default to input here so we can check for both `as` and `children` above
@@ -241,11 +237,6 @@ export const Field = <A extends PickOneConfig>({
     );
   }
 
-  const Component = asElement;
-  return (
-    <Component {...field} {...props}>
-      {children}
-    </Component>
-  );
+  return React.createElement(asElement, { ...field, ...props }, children);
 };
 export const FastField = Field;
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -7,7 +7,7 @@ import {
   FieldValidator,
 } from './types';
 import { useFormikContext } from './FormikContext';
-import { isFunction, isEmptyChildren, isObject } from './utils';
+import { isFunction, isEmptyChildren } from './utils';
 import invariant from 'tiny-warning';
 
 export interface FieldProps<V = any> {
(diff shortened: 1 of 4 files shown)


johnrom:
For reference, here is my version over v1 in case there are some extra edge cases. #1336

williamboman:
This is great, especially like the component props infererence!

src/Field.tsx
```diff
@@ -7,7 +7,7 @@ import {
   FieldValidator,
 } from './types';
 import { useFormikContext } from './FormikContext';
-import { isFunction, isEmptyChildren, isObject } from './utils';
+import { isFunction, isEmptyChildren } from './utils';
 import invariant from 'tiny-warning';
 
 export interface FieldProps<V = any> {
```
johnrom: This isn't your change, but here `FieldProps<V>` appears to be used as both `typeof FieldInputProps<Value>['value']` and `typeof FormikProps<Values>['values']`.

src/Field.tsx
```diff
@@ -65,21 +41,76 @@ export interface FieldConfig {
...
+  : T;
+type AddedProps<T extends FieldConfig> = T extends StringComponentConfig<
+  infer U
+>
+  ? GenericFieldHTMLAttributes<U>
+  : T extends { as: React.ComponentType<infer P> }
+  ? Omit<P, keyof FieldInputProps<any>>
+  : {};
+
+export type FieldAttributes<T extends PickOneConfig> = AddedProps<T> &
```
johnrom: FieldAttributes doesn't appear to result in `as={MyComponent}` having the proper types.

```jsx
interface ExtraProps {
  minimumAgeOfPurchase: number;
}

const MyComponent = (props: FieldInputProps<any> & ExtraProps) => null;

// below has Error! FieldInputProps<any> is not assignable to FieldInputProps<any> & ExtraProps
<Field name="hello" as={MyComponent} minimumAgeOfPurchase={21} />
```

src/Field.tsx
```diff
@@ -65,21 +41,76 @@ export interface FieldConfig {
...
+   */
+  component?: StringElements | React.ComponentType<LegacyFieldProps>;
+  // | React.ComponentType;
 
-export function useField<Val = any>(
-  propsOrFieldName: string | FieldAttributes<Val>
+  /**
+   * Component to render. Can either be a string e.g. 'select', 'input', or 'textarea', or a component.
+   */
+  as?: React.ComponentType<FieldProps['field']> | StringElements;
```
johnrom: `AsElement` receives more than just `FieldInputProps<>`, it receives `{...field, ...props}`. Not sure the best way to type that, and honestly I think some of `...props` should not be passed to the component (it is passed itself as `as`. I thought the definitive type was FieldProps in v1, but now that appears to no longer be what is passed. In that event, we need another definition that will include all of the props that a given component is passed similar to `FieldProps['field'] & AddedProps`

src/Field.tsx
```diff
@@ -65,21 +41,76 @@ export interface FieldConfig {
...
+  children?: React.ReactNode | ((props: FieldProps) => React.ReactElement);
+}
+
+type StringComponentConfig<U extends StringElements> =
+  | { component: U }
+  | { as: U };
+
+// type RenderableConfigKeys = keyof FieldConfig;
+
+type PickOne<T, K extends keyof T> = Pick<T, K> &
```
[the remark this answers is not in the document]
Carl-Foster: I think that's what it does but only if you add both. Intellisense seems to strip out properties which are set to `never`

Carl-Foster:
Is it worth updating this with the types. I don't use the library so I'm not sure what's missing from a type point of view.

jaredpalmer:
Can you summarize the changes with **Before** and **After** and also in which cases they will be breaking? Are we forcing people to add generics to fields? Or is this opt-in?

johnrom:
I'd recommend someone build on top of #1336 if they want to strengthen the typings of Field / Formik. There's a _lot_ to add, and I think I actually did most of it. I plan to circle back one day and build it over v2 but I don't have much flexibility right now. It also had no breaking changes that I could tell since users had to opt-in to strong types.

PR #1799, opened 2019-09-04
Allow re-initialize of errors and touched

ikesler-tenengroup:
Motivation of this PR - to be able to pass errors to formik store from the outside via props, without using callbacks. `errors` and related `touched` objects passed by formik to render function are supposed to be created by `mapPropsToErrors` and `mapPropsToTouched` functions in case when `enableReinitialize` flag is set to true. For example, this config should work:
```
withFormik({
  mapPropsToErrors: (props) => {
    return props.newErrors;
  },
  mapPropsToTouched: (props) => {
        return props.newTouched;
  },
  enableReinitialize: true,
});
```
Currently, when using `mapPropsToErrors` and `mapPropsToTouched` from `withFormik` hoc, values returned by that functions are actually ignored - `resetForm` does not use them. Similar issue - status.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -386,17 +386,17 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       enableReinitialize &&
       isMounted.current === true
     ) {
-      var needReset = false;
+      const needReset = false;
       if (!isEqual(initialValues.current, props.initialValues)) {
         initialValues.current = props.initialValues;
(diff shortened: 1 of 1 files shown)


elham9066:
۱۳۹۸-۰۶-۱۳ ۸:۳۱ ‎−۰۷:۰۰ گرینویچ, Igor Kesler <notifications@github.com>:

PR #1829, opened 2019-09-19
Fix race condition between components using the same field

andyrichardson:
Fixes race condition where two mutually exclusive components use the same field.

If the teardown effect of Component A doesn't complete synchronously, Component B could register a field before having it torn down by Component A which should no longer be considered mounted.

### DIFF
--- a/src/Field.tsx
+++ b/src/Field.tsx
@@ -81,7 +81,7 @@ export function useField<Val = any>(
   const validateFn = isAnObject
     ? (propsOrFieldName as FieldAttributes<Val>).validate
     : undefined;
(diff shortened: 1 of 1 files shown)


src/Field.tsx
```diff
@@ -81,7 +81,7 @@ export function useField<Val = any>(
   const validateFn = isAnObject
     ? (propsOrFieldName as FieldAttributes<Val>).validate
     : undefined;
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
```
jaredpalmer: This will break SSR?

Issue #1722, opened 2019-08-01
Rename useFormik (and maybe useFormikContext)


johnrom:
### Current Behavior

Currently, `useFormik` is an internal hook used to initialize Formik, and `useFormikContext` is the hook used to actually use Formik functionality in user code. The problem is, when a user wants to "use Formik", they reasonably think, "duh, there's a `useFormik` hook! how simple!"

What I'm hearing around the issues here, is that the intended solution is simply to not export `useFormik`, and nobody will need to worry about it. However, I think `useFormik` can be useful if someone wants to take the core formik functionality and use it in a different context. I propose instead renaming these hooks to be more semantic and developer-friendly.

### Desired Behavior

We want a developer who is about to try and "use formik" to be able to see the `useFormik` hook and know they can depend on it, or at least _not_ see a `useFormik` due to the internal hook being semantically named.

### Suggested Solution

Rename the following hooks. These are suggestions which may not ultimately be what we settle on, but I think they move in the right direction.

`useFormik`: `useFormikInitializer`, `useFormikCore`, `useFormikInternals`, etc
`useFormikContext`: `useFormik`

Alias: `useFormikContext` -> `useFormik` in case users have already started using this. Add a development time deprecation warning.

#### Alternate Suggestion

Let there not be a `useFormik` hook at all, using `useFormikInternals` (or whatever) and `useFormikContext`.

### Who does this impact? Who is this for?

Developers who want to "use Formik".

### Describe alternatives you've considered

Answering dozens of questions about why `useFormik` doesn't work in Github Issues, Discord.

### Additional context

I think this will solve a lot of headaches from misunderstandings in the community, while allowing the flexibility to initialize Formik and use it in a different React.Context.

#### Examples of the misconception in action

> Note: Some of these may be my misreading since I glanced quickly through them.

https://github.com/jaredpalmer/formik/issues/1719#issuecomment-516048064
https://github.com/jaredpalmer/formik/pull/1717
https://github.com/jaredpalmer/formik/issues/1667
https://github.com/jaredpalmer/formik/issues/1666
https://github.com/jaredpalmer/formik/issues/1582
https://github.com/jaredpalmer/formik/issues/1571
https://github.com/jaredpalmer/formik/issues/1548
https://github.com/jaredpalmer/formik/pull/1063#discussion_r286970337

jaredpalmer:
Yep

tgroshon:
[...]

PR #1831, opened 2019-09-19
v2 tutorial

jaredpalmer:
Took a first pass at v2 tutorial. 

Note: This tutorial assumes that `getFieldProps()` doesn't return an array, but just field. I think this is more ergonomic for usage outside of `useField`. Will open a separate PR just for that.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -3,217 +3,880 @@ id: tutorial
 title: Tutorial
 ---
 
-## Learn Formik by Building it.
+## Before we start
 
-My talk at React Alicante is really the best way to get started. It introduces the library (by watching me build a mini version of it) and demos how to build a non-trivial form (with arrays, custom inputs, etc.) using the real thing. It's around 54 minutes long.
+Welcome to the Formik tutorial. This will teach you everything you need to know to build simple and complex forms in React.
 
-<iframe width="800" height="315" src="https://www.youtube.com/embed/oiNtnehlaTo" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen title="Taming Forms in React - Jared Palmer"></iframe>
+If you're impatient and just want to start hacking on your machine locally, checkout [the 60-second quickstart](@todo).
 
-## Basics
+### What are we building?
 
-_If you just want to jump into some code, here's a quick walkthrough of the basics..._
+In this tutorial, we'll show how to build a complex newsletter signup form with React and Formik.
 
-Imagine you want to build a form that lets you edit user data. However, your
-user API has nested objects like so.
+You can see what we'll be building here: [Final Result](https://codesandbox.io/s/formik-v2-tutorial-final-t5fe0). If the code doesn't make sense to you, don't worry! The goal of this tutorial is to help you understand Formik.
 
-```js
-{
-   id: string,
-   email: string,
-   social: {
-     facebook: string,
-     twitter: string,
-     // ...
-   }
+### Prerequisites
+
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -3,217 +3,904 @@ id: tutorial
...
 ---
 
-## Learn Formik by Building it.
+## Before we start
 
-My talk at React Alicante is really the best way to get started. It introduces the library (by watching me build a mini version of it) and demos how to build a non-trivial form (with arrays, custom inputs, etc.) using the real thing. It's around 54 minutes long.
+Welcome to the Formik tutorial. This will teach you everything you need to know to build simple and complex forms in React.
 
-<iframe width="800" height="315" src="https://www.youtube.com/embed/oiNtnehlaTo" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen title="Taming Forms in React - Jared Palmer"></iframe>
+If you're impatient and just want to start hacking on your machine locally, checkout [the 60-second quickstart](@todo).
```
johnrom: Should the 60-second quickstart be hooked up? If it should be left as a todo, we should add "(todo)" after it so people aren't confused.

docs/tutorial.md
```diff
@@ -3,217 +3,904 @@ id: tutorial
...
 
-## Basics
+### What are we building?
 
-_If you just want to jump into some code, here's a quick walkthrough of the basics..._
+In this tutorial, we'll show how to build a complex newsletter signup form with React and Formik.
 
-Imagine you want to build a form that lets you edit user data. However, your
-user API has nested objects like so.
+You can see what we'll be building here: [Final Result](https://codesandbox.io/s/formik-v2-tutorial-final-t5fe0). If the code doesn't make sense to you, don't worry! The goal of this tutorial is to help you understand Formik.
```
johnrom: This line should have a bit more of a caveat that _the way the tutorial begins isn't really how we'd recommend someone starting out builds with Formik_. I think we get a lot of questions in the chat and with issues concerning how a format like the introductory code doesn't work with `useField`.

I think #1722 can help with that, but I also think this line can be more descriptive that the goal is to get to the end which is totally different.

"We're going to start by building a Formik form manually, which will eventually evolve into using the extremely simple and recommended API provided by Formik. The manual method can still be used for more complex scenarios etc etc, but is here mainly to demonstrate how the API works internally."

docs/tutorial.md
jaredpalmer: After reading this today, I don't think this is really a tutorial, but more of a "thinking" in formik. It's a start, but lots more to do.

docs/tutorial.md
jaredpalmer: Want to jump on discord or a zoom and collab on this?

johnrom: I had limited time this morning and spent it reading through this 😆 but I think calling this "Thinking in Formik" and creating a separate "Quickstart Tutorial" where this Thinking in Formik document basically builds the Quickstart example manually and demonstrates how the Quickstart example works underneath, then the Quickstart starts with something like "How is it this easy?! Learn how with "Thinking in Formik ->"" or something.

johnrom: I may have some async import tasks later where I can squeeze a quick chat in while I'm waiting, but I don't want to promise. I'll ping you in Discord if I do, and if it works out, it works out!

jaredpalmer: Sounds good. I agree 100% with the split between quick start and thinking in formik

docs/tutorial.md
```diff
@@ -3,217 +3,904 @@ id: tutorial
...
+      onSubmit={(values, { setSubmitting }) => {
+        setTimeout(() => {
+          alert(JSON.stringify(values, null, 2));
+          setSubmitting(false);
+        }, 400);
+      }}
+    >
+      <Form>
+        <label htmlFor="firstName">First Name</label>
+        <Field id="firstName" type="text" />
```
fredrik-sogaard: All of these `id`s should be `name`s, no?

johnrom: They should actually be both, because `label for={}` applies only to the id for accessibility purposes.

johnrom:
FYI I created an issue for documentation here, and split the tutorial item into Quickstart and Thinking in Formik: #1836

docs/tutorial.md
```diff
@@ -3,217 +3,904 @@ id: tutorial
...
+        .required('Required'),
+    }),
+    onSubmit: values => {
+      alert(JSON.stringify(values, null, 2));
+    },
+  });
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      <label htmlFor="firstName">First Name</label>
+      <input id="firstName" {...getFieldProps('firstName')} />
```
denisborovikov: It should be `...formik. getFieldProps` everywhere in the code.

lednhatkhanh: It should be `...formik. getFieldProps({ name: 'firstName' })`, same for the other fields.

Issue #321, opened 2017-12-23
<Field /> and undefined initial value fires error


Zaitsev:
if value in `<Formic initialVales>` is undefined React throws error

> Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
>  in input (created by Field)

   
```JavaScript
const InnerForm2=()=>(
    <Form className={styles.Form}>
            <Field type={"text"} name={"clusterName"}/>
    </Form>
)
...
             <Formik
                    validate={this.validate as any}
                    validateOnBlur
                    onSubmit={values => console.log(values)}
                    initialValues={{clusterName: undefined,type:1}}
                    component={InnerForm2 as any}
                />
...
```
with `undefined`  value `<input />` rendered _without_ `value` attribute and first update causes `value` rendering and this throw error

jaredpalmer:
Correct. You must initialize values just as you would with `state`

mellis481:
@jaredpalmer This is fine for fields of type `string` (I'm using Typescript), you can initialize as an empty string.  But for inputs of type `number`, you have to use an initial value of `null` or `undefined` if you don't want a value to appear in the input by default.  And this continues to show the "uncontrolled input" warning.

Here is my code:


```
export interface InvoiceLevelTotal {
  amount?: number;
  currency?: string;
}

export enum FormKeys {
  Amount = 'amount',
  Currency = 'currency'
}

const formSchema: Yup.ObjectSchema<InvoiceLevelTotal> = Yup.object().shape({
  amount: Yup.number().required('Amount is required'),
  currency: Yup.string().required('Currency is required')
});

const initialFormValues: InvoiceLevelTotal = {
  amount: undefined,
  currency: ''
};

<Formik
  initialValues={initialFormValues}
  onSubmit={(values, { resetForm }) => {
    this.addInvoiceLevelTotal(values);
    resetForm();
  }}
  validationSchema={formSchema}>
    {(formProps: FormikProps<InvoiceLevelTotal>) => (
      <Form>
        <Field
          name={FormKeys.Amount}
          render={({ field, form }: FieldProps<InvoiceLevelTotal>) => (
            <input
              className="input-block-level"
              maxLength="20"
              {...field}
            />
          )}
        />
        <Field
          name={FormKeys.Currency}
[...]

PR #1850, opened 2019-09-30
Allow initialValues to be an optional prop

ksweetie:
`initialValues` should be an optional prop. All fields need to be initialized with a value, but that can be done at the `Field` level, as shown here:

https://github.com/jaredpalmer/formik/issues/321#issuecomment-506963902

```js-jsx
<Field name={name}
  render={({field}) => (
    <Input
        {...field} 
        value={field.value || ''}  
     />
    )}
/>
```

Since `initialValues` is currently required, that forces us to pass in an empty object for no reason:

```js-jsx
<Formik initialValues={{}} onSubmit={handleSubmit}>
  <Form>
    <CustomInput name="firstName" />
  </Form>
</Formik>
```

This reduces the boilerplate code required to set up a form. Additionally, this DRYs up the form more, as you don't have to specify `firstName` on the input and inside the `initialValues`

### DIFF
--- a/src/types.tsx
+++ b/src/types.tsx
@@ -180,7 +180,7 @@ export interface FormikConfig<Values> extends FormikSharedConfig {
   /**
    * Initial values of the form
    */
-  initialValues: Values;
+  initialValues?: Values;
 
   /**
    * Initial status


johnrom:
@ksweetie I worry about making a change like this because, as far as I know, Formik uses the shape of the initialValues to determine the shape of things like `touched` and `errors`.

jaredpalmer:
Correct. If the user hits enter immediately formik will not be able to understand what should be touched or not.

PR #1916, opened 2019-10-25
Bring back FastField using contextType

jaredpalmer:
- Add back FastField from v1.
- Get rid of `connect` and just use `contextType`


sibelius:
What about a memoField helper? That’s wrapper a React.memo with the same check of shouldComponentUpdate

jaredpalmer:
@sibelius if you can get it done in the next 1.5hr then we can get it into v2 release.

PR #1933, opened 2019-10-27
RFC: Declarative Field Constraints

kuzmycz:
Add field constrains as a simpler method for validation fields. This is mean to cover 80% of user requirements and should lead to simpler code and less boilerplate. The constraints added are:

- required
- minLength
- maxLength
- numeric
- min
- max
- equal
- isEmail
- match

Sample usage

```
import React from 'react';
import { Formik, Form, Field } from 'formik';

export const FieldLevelValidationExample = () => (
  <div>
    <h1>Signup</h1>
    <Formik
      initialValues={{
        username: '',
        email: '',
      }}
      onSubmit={values => {
        // same shape as initial values
        console.log(values);
      }}
    >
      {({ errors, touched, isValidating }) => (
        <Form>
          <Field name="email" required isEmail />
          {errors.email && touched.email && <div>{errors.email}</div>}

          <Field name="username" required minLength={6} />
          {errors.username && touched.username && <div>{errors.username}</div>}

        </Form>
      )}
    </Formik>
  </div>
);

```
Note, field level validation can be used in conjunction with field constraints

-----
[View rendered docs/guides/validation.md](https://github.com/kuzmycz/formik/blob/declarative-validation/docs/guides/validation.md)


jaredpalmer:
This is an interesting idea. It is definitely expanding Formik's scope into validation, but that has been something I've thought about over the years. I think this definitely solves for some simple use-cases where even Yup is overkill. 

After skimming the PR, my first immediate API suggestion is that you should each  validator props should optionally take an array where the first element is the value of the constraint and the second is a string error message. This gives more Yup-like control, better DX, and allows for i18n.

```tsx
<Field name="username" minLength={3} /> // => Some default error message
<Field name="username" minLength={[3, 'Too short'])}  /> // => 'Too short'
```

@eonwhite @nathanforce any thoughts?

nathanforce:
At a glance, I worry that a bottom-up approach like this will lead to more code complexity and thus potential edge cases, and, maybe more notably, an extra render pass before validating is possible. Again, I didn't spend a ton of time thinking on it, but that's my initial take.

kuzmycz:
Hello @nathanforce, @jaredpalmer, @eonwhite,

I have added support to allow a user to specify a different error message. As requested, you can optionally provide a touple where the first value is the value of the constraint and the second is the error message.

For example,

```
<Field name="username" minLength={3} /> // => Some default error message
<Field name="username" minLength={[3, 'Too short'])}  /> // => 'Too short'

```

Also, wrt performance, I have tested the validator rebuilding and this only happens if the constraint values (or error messages) change. If the constraint is dynamic then the validator is recomputed only if any dependencies are changed.

PR #1940, opened 2019-10-28
Fix prepareDataForValidation when the object is an instance.

krailler:
The prepareDataForValidation function, call recursively when the value is an object but it is possible that the value is an instance, like a Moment. With this change (Inline 1007) we're checking that the value is one javascript object, if not... don't touch.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -1001,7 +1001,11 @@ function prepareDataForValidation<T extends FormikValues>(
             return value !== '' ? value : undefined;
           }
         });
-      } else if (typeof values[key] === 'object' && values[key] !== null) {
+      } else if (
+        typeof values[key] === 'object' &&
+        values[key] !== null &&
+        Object.getPrototypeOf(values[key]).constructor.name === Object.name
+      ) {
         data[key] = prepareDataForValidation(values[key]);
       } else {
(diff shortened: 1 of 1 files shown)


src/Formik.tsx
```diff
@@ -1001,7 +1001,11 @@ function prepareDataForValidation<T extends FormikValues>(
             return value !== '' ? value : undefined;
           }
         });
-      } else if (typeof values[key] === 'object' && values[key] !== null) {
+      } else if (
+        typeof values[key] === 'object' &&
```
jaredpalmer: Can you add a comment about this? also can we use `isObject()`?

krailler: Mmm, I'm not sure if possible to use isObject. Using 

> values[key] instanceof Object

 also returns true...

I didn't find any other possible solution to avoid this.

krailler: The problem is...
`{ date: Moment() /* moment instance */ }`

How we can check that the date property is an object?

Andrey-Bazhanov: We can use [is-plain-object](https://github.com/jonschlinkert/is-plain-object) for this

krailler: Yes, changed.

PR #1949, opened 2019-10-29
Fix prepareDataForValidation work with instances

Andrey-Bazhanov:
This pull request fix same problem as [1940](https://github.com/jaredpalmer/formik/pull/1940), but I used lodash `isPlainObject` methods and writed tests for `prepareDataForValidation`. Unfortunately I exported this function, because for this case it was hard to test whole Formik component

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import isEqual from 'react-fast-compare';
 import deepmerge from 'deepmerge';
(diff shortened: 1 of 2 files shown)


krailler:
Yayyy, perfect! You also include the tests 👏

krailler:
Hi! This fix is not on the latest release (2.0.3), when will this come out?

austincrft:
@krailler - v2.0.4 was just released 20 hours ago. Updating to that version fixed it for me

PR #1959, opened 2019-10-29
Changed from .min() to max() method 

martincomito:
Changed from `.min()` method to `max()` to match the use case:
This part of the Tutorial uses validationSchema to limit the input to 15 or less characters, and uses `min()` instead of `max()` method to make the Yup schema validation.

-----
[View rendered docs/tutorial.md](https://github.com/martincomito/formik/blob/patch-2/docs/tutorial.md)

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -501,10 +501,10 @@ const SignupForm = () => {
     },
     validationSchema: Yup.object({
       firstName: Yup.string()
-        .min(15, 'Must be 15 characters or less')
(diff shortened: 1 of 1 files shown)


bahoffman:
I think you missed the other code examples on this page with the same issue - https://jaredpalmer.com/formik/docs/tutorial

martincomito:
> I think you missed the other code examples on this page with the same issue - https://jaredpalmer.com/formik/docs/tutorial

You´re right, I missed it! Thanks!

PR #1964, opened 2019-10-30
Fix unnecessary let usage

karl-run:
When only properties are updated on the error object it should be `const` not `let`.

-----
[View rendered docs/guides/validation.md](https://github.com/karl-run/formik/blob/patch-2/docs/guides/validation.md)


jaredpalmer:
Can you go through the /website/versioned_docs folder and update this everywhere on older versions?

karl-run:
A find replace should do the trick, the few places where `let` is correct doesn't assign an empty object as the initial value.

PR #1966, opened 2019-10-30
Update tutorial with missing example / typos

meeoh:
The very first example in the tutorial was missing the actual call to `useFormik`. Not sure if this was intended but seems like it wasnt.


jaredpalmer:
Thanks!bitmoji

PR #1972, opened 2019-10-30
Expose formikbag as imperative methods

### LINKED ISSUE #1603: [v2]  <Formik /> `ref` prop
### Current Behavior

Since switching `<Formik />` to a function component if you try to attach a ref to it as such:
```js
<Formik ref={myRef} /> 
```
you will get the following error:
`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

### Desired Behavior

The desired behaviour is for there to be no error and the ref to provide a set of APIs to programmatically interact with `<Formik />`.

### Suggested Solution

To enable the functionality it seems that the hook `useImperativeHandle` will be needed to wrap `<Formik/>` and expose the available methods.

```
export const Formik = forwardRef((props, ref) => {
  const formikRef = useRef()
  useImperativeHandle(ref, () => ({
    submitForm: () => handleSumbit()
  })
  
  return (
    <FormikProvider ref={formikRef} value={formikbag}>
      {...}
    </FormikProvider>
  )
})
```

An example per the [docs](https://reactjs.org/docs/hooks-reference.html#useimperativehandle)

This is for anyone who needs to implement any programmatic functionality on <Formik />.

Use the prior version where <Formik /> is a class component.

### Additional context

I was trying to implement the debounced autosave that was handled by the parent component, without the need for a specific component that is a child of <Formik />.

brunohkbx:
### Motivation
Since Formik 2 is built on top of React Hooks it was not possible to attach a ref and consequently, some people can't upgrade to v2. It resolves #1603.

My first approach was to use `React.forwardRef` like that:
```
export const Formik = React.forwardRef(<
  Values extends FormikValues = FormikValues,
  ExtraProps = {}
>(props: FormikConfig<Values> & ExtraProps, ref?: any) => {
```

But it breaks the interface [FormikConfig](https://github.com/jaredpalmer/formik/blob/master/src/types.tsx#L163) giving the following errors:
```
Types of property 'render' are incompatible.
Types of property 'onReset' are incompatible.
Types of property 'valiadte' are incompatible.

'FormikValues' is assignable to the constraint of type 'V', but 'V' could be instantiated with a different subtype of constraint '{}'.
```

I've managed to make it work when I changed `FormikProps<Values>` to `FormikProps<any>` on `render?`, `onReset`, `validate`, but I don't know how to properly fix these types :disappointed:

Do you guys have any thoughts on this? I would appreciate any feedback :grimacing:


jaredpalmer:
I think this should work?

jaredpalmer:
@johnrom thoughts?

johnrom:
I'm very much in dotnet-land right now, but if you want feedback on the TypeScript I think you know my answer :). There should be a proper type (typeof formikBag ?) instead of `any`.

```ts
export interface FormikConfig<Values> extends FormikSharedConfig {
  innerRef?: React.MutableRefObject<FormikBag<Values>>;
}
```

Additionally, I don't know that a reference to the FormikBag is equivalent to a "reference" to the Formik component, and I think there are better ways to access the Formik API than prop drilling -- there is discussion around this on the issue linked above. Here's an example using Context. It's verbose, but with the benefit of being completely customizable.

```tsx
type FormikSubmitContext = {
    registerOnSubmit: (handler: FormikHandlers['onSubmit']) => void;
    onSubmit?: FormikHandlers['onSubmit'];
}
const formikSubmitContext = React.createContext<FormikSubmitContext | undefined>();
const MyParentComponent = () => {
    const [onSubmit, registerOnSubmit] = React.useState<FormikHandlers['onSubmit']>();
    return <formikSubmitContext.Provider value={{
        onSubmit,
        registerOnSubmit
    }}>
        <Formik>
            {formikProps => {
                const myContext = React.useContext(formikSubmitContext);
                
                if (!myContext) throw new Error("no context");

                React.useEffect(() => {
                    myContext.registerOnSubmit(formikProps.onSubmit);

                    return () => myContext.registerOnSubmit(undefined);
                })
            }}
            {() => {
                const myContext = React.useContext(formikSubmitContext);

                return <button onClick={myContext.onSubmit}>Submit</button>
            }}
        </Formik>
    </formikSubmitContext.Provider>
}
```

belgamo:
Any updates on this?

jaredpalmer:
Why can't we just do useImperativeHandle?

brunohkbx:
I'm ok with this

ksweetie:
@jaredpalmer Is the path forward here to do `useImperativeHandle` and properly type it? I can take a stab at it if that's what we want to do.

jaredpalmer:
@ksweetie yes

jaredpalmer:
And update the PR to deal with Lerna

jaredpalmer:
@ksweetie want to take a stab at this?

drivasperez:
Is there any progress on this? It would make my life easier and I don't mind taking it on, but I don't want to swoop in if it's already being worked on.

brunohkbx:
I'll fix the Conflicting files

rally25rs:
Using 
```
formik 2.1.2
react 16.12.0
```

if I do:
```
const formikRef = React.useRef();

<Formik ref={formikRef}>
```

i still get:
```
  console.error node_modules/react-dom/cjs/react-dom.development.js:530
    Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

    Check the render method of `ForwardRef(FormImpl)`.
        in Formik (created by ForwardRef(FormImpl))
```

and `formikRef.current` is `undefined`

Is this not the right way to be using this now? It used to work in a past version of react and formik, just going through a round of upgrading all my deps...

PR #1982, opened 2019-11-01
Add generic type to FieldArray types.

### LINKED ISSUE #911: Add generics to ArrayHelpers type
## Current Behavior
The current `ArrayHelpers` type does not use generics in its type.

## Desired Behavior
It would be nice if the ArrayHelpers added generics so that we could type check the inner methods as well.

## Suggested Solution
```js
export type FieldArrayRenderProps<T = any, Values = any> = ArrayHelpers<T> & {
  form: FormikProps<Values>;
  name: string;
};

interface ArrayHelpers<T = any> {
  push: (obj: T) => void;
  handlePush: (obj: T) => () => void;
  swap: (indexA: number, indexB: number) => void;
  handleSwap: (indexA: number, indexB: number) => () => void;
  move: (from: number, to: number) => void;
  handleMove: (from: number, to: number) => () => void;
  insert: (index: number, value: T) => void;
  handleInsert: (index: number, value: T) => () => void;
  replace: (index: number, value: T) => void;
  handleReplace: (index: number, value: T) => () => void;
  unshift: (value: T) => number;
  handleUnshift: (value: T) => () => void;
  handleRemove: (index: number) => () => void;
  handlePop: () => () => void;
  remove(index: number): T | undefined;
  pop(): T | undefined;
}
```

## Who does this impact? Who is this for?
* No impact if we use defaults for the generics
* Allows users to strongly type the arrayHelpers prop if desired.

I'm willing to submit a PR for this if this sounds ok.

IronSean:
I added a generic type to the FieldArrayRenderProps type and the ArrayHelpers type.

This will allow usage of FieldArray like this:

```ts
<FieldArray name={"myArrayFieldName"}>
  {(fieldArrayProps: FieldArrayRenderProps<MyFormType, MyFieldArrayType>)  => {
    fieldArrayProps.form; //typed as FormikProps<MyFormType> instead of FormikProps<any>
    fieldArrayProps.push; //takes a typed input object of MyFieldArrayType
    fieldArrayProps.pop; //returns an object of type MyFieldArrayType
    return (
      //My componenets    
  )}}
</FieldArray>
```

However, because of the existing generic property on `pop` and `remove`, to allow adding types just on the `pop`/`remove` calls, this may be a **breaking change**. as I had to change it from 
`pop<T>(): T | undefined;` 
to 
`pop: () => T | undefined`

Fixes #911, as well as making the FieldArray component able to be type checked in TypeScript.

### DIFF
--- a/src/FieldArray.tsx
+++ b/src/FieldArray.tsx
@@ -9,8 +9,11 @@ import {
 } from './types';
 import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
 
-export type FieldArrayRenderProps = ArrayHelpers & {
-  form: FormikProps<any>;
+export type FieldArrayRenderProps<
+  FormValue = any,
+  FieldArrayValue = any
+> = ArrayHelpers<FieldArrayValue> & {
+  form: FormikProps<FormValue>;
   name: string;
 };
 
@@ -20,11 +23,11 @@ export type FieldArrayConfig = {
   /** Should field array validate the form AFTER array updates/changes? */
   validateOnChange?: boolean;
 } & SharedRenderProps<FieldArrayRenderProps>;
-export interface ArrayHelpers {
+export interface ArrayHelpers<FieldArrayValue = any> {
   /** Imperatively add a value to the end of an array */
-  push: (obj: any) => void;
+  push: (obj: FieldArrayValue) => void;
   /** Curried fn to add a value to the end of an array */
-  handlePush: (obj: any) => () => void;
+  handlePush: (obj: FieldArrayValue) => () => void;
   /** Imperatively swap two values in an array */
   swap: (indexA: number, indexB: number) => void;
   /** Curried fn to swap two values in an array */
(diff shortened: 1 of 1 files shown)


intgr:
Hi, thanks for working on this. We recently found out by surprise that `ArrayHelpers` is not type-safe in TypeScript.

Really looking forward to when this is merged. But it's almost a year later and nobody has shown interest. :(

src/FieldArray.tsx
```diff
@@ -9,8 +9,11 @@ import {
 } from './types';
 import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
 
-export type FieldArrayRenderProps = ArrayHelpers & {
-  form: FormikProps<any>;
+export type FieldArrayRenderProps<
+  FormValue = any,
+  FieldArrayValue = any
```
intgr, line 14: Note that I'm just a lurker here, not a Formik developer.

But it seems the convention is that type variables are either called `T` or short single-letter words.

So for consistency, maybe `FormValue` -> `Values` and `FieldArrayValue` -> `T`

maddhruv:
Hey @IronSean sorry for the late response, can you please rebase with the latest master and also resolve the conflict 

would be pleasured to review

src/FieldArray.tsx
jaredpalmer: Correct.

jaredpalmer (changes requested):

Philipp91:
> can you please rebase with the latest master and also resolve the conflict

https://github.com/formium/formik/pull/3401

mateuszpigula:
What's the status?

IronSean:
I am closing this because we stopped using Formik a couple years ago and I have no idea how relevant it is or isn't anymore

Issue #1976, opened 2019-10-31
[v2] calling resetForm({values}) with nested objects won't reset dirty and isValid state to false


simmco:
### Current Behavior

If you are calling ```resetForm({values})``` with values of nested objects the ```dirty``` and ```isValid``` state is still true.

### Expected behavior

As in v1 ```dirty``` and ```isValid``` state should be false after calling ```resetForm({values})``` with the new values.

### Reproducible example

v1.5.8: https://codesandbox.io/s/formik-codesandbox-template-tkqcc
v2.0.3: https://codesandbox.io/s/formik-codesandbox-template-076j5

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.3
| React            | 16.11.0

Tigge:
Does the sandbox show the `isValid` issue as well? I wasn't sure how to reproduce it. I've create the merge request #1986 to solve the `dirty` part at least.

simmco:
I just noticed that the ```isValid``` state is also true. And in v1 it is false. The sandbox should show it. Don't know if it is intended so?
Thanks for your work! 👏

jsilvax:
I'm also experiencing this issue. I tested the PR @Tigge has open, and can verify it fixes the issue.

Tigge:
@simmco for `isValid` I think this is to be expected in version 2. The `isInitialValid` (which defaulted to `false`) is now deprecated, and replaced with `initialErrors` (which defaults to `{}`, meaning no errors).

This is a good change, and while `isInitialValid` is deprecated but still usable in version 2, so you can get the same behaviour by setting it to `false`. But if you want to start with an invalid form it is better to set `initialErrors` instead.

sahlouls:
Any updates on this one ?

PR #1986, opened 2019-11-01
Fix dirty flag when calling resetForm with current values

Tigge:
When you are resetting the form using `resetForm` and specify current
formik values as the next state the dirty hook won't be run since the
current values didn't change. This adds `intialValues.current` as a
dependency so the hook will run again if either of these changes.

This fixes the dirty part of issue #1976.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -847,7 +847,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
 
   const dirty = React.useMemo(
     () => !isEqual(initialValues.current, state.values),
-    [state.values]
+    [initialValues.current, state.values]
   );
 
   const isValid = React.useMemo(
--- a/test/Formik.test.tsx
+++ b/test/Formik.test.tsx
@@ -881,6 +881,19 @@ describe('<Formik>', () => {
     });
   });
 
+  describe('resetForm', () => {
+    it('should reset dirty when reseting to same values', () => {
+      const { getProps } = renderFormik();
+      expect(getProps().dirty).toBe(false);
+
+      getProps().setFieldValue('name', 'jared-next');
+      expect(getProps().dirty).toBe(true);
+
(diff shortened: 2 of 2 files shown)


ianks:
@jaredpalmer any chance of getting this merged?

src/Formik.tsx
```diff
@@ -847,7 +847,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
 
   const dirty = React.useMemo(
     () => !isEqual(initialValues.current, state.values),
-    [state.values]
+    [initialValues.current, state.values]
```
jaredpalmer: I don't think you need .current here

Tigge: It's needed. The `initialValues` reference will never change, even if some of its properties does.

There are some potential pitfalls with having refs in the dependency list. See <https://github.com/facebook/react/issues/14387> for example. This is a slightly different case since I think `resetForm` will always trigger a rerender, and this ref is not a ref to a dom node, which would have been updated after the effect hook was run.

Another solution would be to move these initial values/errors to the state instead, but I haven't looked into what that would mean for functionality and performance.

babalugats76:
Thanks to you guys for all the awesome work.  I hate to be **that** guy, but is there an ETA on this?  For my projects, this is core functionality.  Currently, I have no way to set `dirty` back to false unless I push the data down and rely upon form re-initialization via new props.  This may be a "small" feature, but it is critically important.  Many thanks!

Tigge:
@babalugats76 This has already been merged and was released in v2.0.7.

babalugats76:
Thanks @Tigge!  Looks like I need to do some more testing.  I am using `2.1.4`.  I also have nested objects which I know have been problematic in the past--don't know if deep comparisons are supported since hooks are know prominently featured in Formik.  I think I will will roll back to `2.0.7`, recreate my use case, see what happens, and then roll forward from there.

Issue #1957, opened 2019-10-29
setSubmitting not working


KrzysztofCzaplejewicz:
### Current Behavior

setSubmitting method not changing FormikProps isSubmitting

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.3
| React            | 16.10.2
| TypeScript       | 3.6.4
| Browser          | Opera
| npm/Yarn         | 1.19.1
| Operating System | Windows10

warnellw:
Experiencing a similar issue. `isSubmitting` is automatically getting set back to `false` in V2. According to the [docs](https://jaredpalmer.com/formik/docs/guides/form-submission#submission) and the behavior in V1, `isSubmitting` must manually be set to `false` in the handler.

artemkrynkin:
same problem

vkasraj:
I first thought that my form has some issues. After debugging I can say that `setSubmitting` is working but the issue is that `isSubmitting` set back to `false` after `handleSubmit` get executed.

KrzysztofCzaplejewicz:
> I first thought that my form has some issues. After debugging I can say that `setSubmitting` is working but the issue is that `isSubmitting` set back to `false` after `handleSubmit` get executed.

true, need to create for example promise or async and await for response submit,

paulkre:
Can someone fix this or change the documentation, please? This is basic functionality for most forms and I went almost insane trying to figure out what was wrong 😅

Tigge:
It's a bit unclear which behaviour is the intended one for version 2. If we can get some clarification I'd be happy to create a pull request.

kbi-daniel:
> It's a bit unclear which behaviour is the intended one for version 2. If we can get some clarification I'd be happy to create a pull request.

@Tigge For what it's worth, my vote would be to duplicate as closely as possible the behavior of Version 1, since I believe this would alleviate the need to refactor code to address this issue and in turn make migration to Version 2 as seamless as possible.

Just as an example, I have been using isSubmitting in my fields and buttons to disable functionality. 
However, after migrating to Version 2, I now get a weird effect where my fields and buttons are made active for a split second before I have an opportunity to transition away.  From a strictly selfish perspective, I'd hate to have to refactor all these forms in order to stay current with the latest version of Formik.

[...]

PR #1987, opened 2019-11-02
Change `isSubmitting` behaviour to mimic v1

Tigge:
In version 2 the behaviour of `isSubmitting` was changed from having the
user handle its state after submitting to always resetting it to false
after the `onSubmit` handler had returned (or it's promise). This reverts
it back to the version 1 behaviour of not touching the `isSubmitting`
state after the `onSubmit` handler has been called.

Solves: #1957

### DIFF
--- a/src/Form.tsx
+++ b/src/Form.tsx
@@ -8,7 +8,7 @@ export type FormikFormProps = Pick<
     'onReset' | 'onSubmit'
   >
 >;
-    
+
 type FormProps = React.ComponentPropsWithoutRef<'form'>;
 
 // @todo tests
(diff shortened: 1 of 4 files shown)


AllenAnthes:
I agree that the lack of docs around this change was definitely unfortunate but it does appear to fix a longstanding issue (#486) that prevented chaining functions onto `submitForm` (for example showing loading/submission state on a button).

That being said, the above only applies to returning a promise from `submitForm`, not the automatic setting of `isSubmitting`

Tigge:
@AllenAnthes yeah, the promise behaviour could certainly be kept as well in this PR. I've updated it accordingly.

rofazayn:
@Tigge, Who can we contact to merge this? Is there a way I could fix this myself by editing my local Formik files?

Sorry for the dumb question, I am a noob x)

Tigge:
@rofazayn that's @jaredpalmer - I'm sure he'll get around to addressing this eventually. For now, if you want to test, you might be able to do it using <https://docs.npmjs.com/files/package.json#git-urls-as-dependencies> by point to my branch.

vkasraj:
Why it is not yet merged? I think everyone's form is working fine.

PR #1993, opened 2019-11-04
feat: remove lodash-es and lodash, using individual modules instead

jonathan-fielding:
This pull request swaps out lodash-es which is currently in use for individual parts that are being used, the aim is to make the build more performant


jonathan-fielding:
Before:
Screenshot 2019-11-04 at 19 19 56
After:
Screenshot 2019-11-04 at 20 17 49

jaredpalmer:
Cc @trysound

TrySound:
1. `lodash.*` packages will not be supported in lodash 5. So consider them as deprecated.
2. These packages does not support esm which increases bundle size especially with webpack.
3. Common code cannot be reused between `lodash.*` packages.
4. Webpack bundle analyzer may lie. Check please bundle size on disk also make sure you use production build. And provide repro please.

So just leave as is. Build will not be more performant. Probably worse as we loose esm which is easier to process. The only tradeoff is node_modules size.

TrySound:
I get this
with `lodash.*` packages - 51865
with `lodash-es/*` - 45728

with the following config
```
module.exports = {
  mode: 'production',
  entry: './index.js',
  externals: ['react', 'scheduler']
}
```
and this entry point
```
import * as formik from 'formik';
console.log(formik)
```

astei:
I'm going to necro this PR, because I've run across this very problem. I am currently building a React application that uses `formik` with `yup` validation (as recommended by the documentation). The application also uses `lodash` itself. When I finished integrating yup support, I checked the bundle size and found my app had grown by 21kB. This seemed quite large, so I analyzed the bundle and found this:
image

After digging further, I found that the source of the problem is, in fact, `formik` (by way of TSDX) and `yup` having very different ideas of how to include `lodash`. As a part of post-build processing, TSDX rewrites all lodash imports to use `lodash-es` instead. `yup` doesn't do this, instead (presumably) relying on tree-shaking to do the job for us (we use `create-react-app`, so this is automagically taken care of using webpack) - I haven't had time to fully investigate as it's pretty late for me.

The obvious solution seems to be to follow what `yup` does and not rewrite `lodash` imports, but I'll keep an open mind for a better solution.

cc @jaredpalmer

TrySound:
@astei The difference is that formik uses lodash-es and yup uses commonjs version. The problem is missing es modules support in yup. Which is actually added 8 days ago but not released yet. Though there is no support for lodash rewrites. You can submit PR to add them similar to formik.

Omitting esm is not a solution. Commonjs often adds a lot of overhead.

astei:
@TrySound I can try and look into doing that but I'm not very familiar with `rollup` and it's pretty late for me. I'll get some rest and I'll see how I can proceed.

TrySound:
There is nothing to do with rollup. Fomrik uses https://github.com/laat/babel-plugin-transform-rename-import to do the work.

astei:
Blame it on the lack of sleep :man_facepalming:

astei:
@TrySound It's been kind of frustrating, but I finally got it working and opened jquense/yup#756.

willynlabs:
I will try to add a specific case and supporting documentation in case others encounter something similar. 

Like others (I believe) I am trying to include Formik in an intermediate library of sorts, and include it with some custom form controls, layouts, etc.  I use rollup to build that library, which is then published as an npm module, and included in my app, which is built with webpack.

My intermediate library grew from about 82k to about 237k when I added Formik using the published npm module.  Obviously, part of this was Formik, but it had to be more than just the (small) Formik components.

With the individual imports like `import clone from 'lodash/clone';` and
`import toPath from 'lodash/toPath` the rollup build was incorporating much of lodash-es into the resulting bundle.  Here is a part of the resulting map file.

"sources": [
    "../src/layouts/SimpleClosableLeftPanelLayout.js",
    "../src/layouts/SimpleClosableRightPanelLayout.js",
    "../src/layouts/SimpleClosableBottomPanelLayout.js",
    "../src/layouts/SimpleClosableTopPanelLayout.js",
    "../src/layouts/PageNavView.js",
    "../src/layouts/OneOverlayView.js",
    "../src/views/PopoverMenuView.js",
    "../node_modules/@babel/runtime/helpers/esm/extends.js",
    "../node_modules/@babel/runtime/helpers/esm/inheritsLoose.js",
    "../node_modules/@babel/runtime/helpers/esm/assertThisInitialized.js",
    "../node_modules/memoize-one/dist/memoize-one.esm.js",
    "../node_modules/react-window/dist/index.esm.js",
    "../node_modules/react-virtualized-auto-sizer/dist/index.esm.js",
    "../src/views/GridView.js",
    "../src/views/ListView.js",
    "../node_modules/react-fast-compare/index.js",
    "../node_modules/deepmerge/dist/es.js",
    "../node_modules/lodash-es/_freeGlobal.js",
    "../node_modules/lodash-es/_root.js",
    "../node_modules/lodash-es/_Symbol.js",
    "../node_modules/lodash-es/_getRawTag.js",
    "../node_modules/lodash-es/_objectToString.js",
    "../node_modules/lodash-es/_baseGetTag.js",
    "../node_modules/lodash-es/_overArg.js",
    "../node_modules/lodash-es/_getPrototype.js",
    "../node_modules/lodash-es/isObjectLike.js",
    "../node_modules/lodash-es/isPlainObject.js",
    "../node_modules/lodash-es/_listCacheClear.js",
    "../node_modules/lodash-es/eq.js",
    "../node_modules/lodash-es/_assocIndexOf.js",

I tried a bunch of different rollup configuration options, but stopped short of a transpiling plugin or doing a "replace" behind the scenes.

Finally, I forked the current repository, and changed the imports to use `import { clone, toPath } from 'lodash';` instead, and then added lodash-es to the rollup external option:

external: [
    'react-router-dom',
    'classnames',
    'lodash',
    'lodash-es',
    ...

which allowed rollup to skip these modules, but still resulted in the (desired) lodash-es imports.  Here is the source map from this build of my library:


"sources": [
    "../src/layouts/SimpleClosableLeftPanelLayout.js",
    "../src/layouts/SimpleClosableRightPanelLayout.js",
    "../src/layouts/SimpleClosableBottomPanelLayout.js",
    "../src/layouts/SimpleClosableTopPanelLayout.js",
    "../src/layouts/PageNavView.js",
    "../src/layouts/OneOverlayView.js",
    "../src/views/PopoverMenuView.js",
    "../node_modules/@babel/runtime/helpers/esm/extends.js",
    "../node_modules/@babel/runtime/helpers/esm/inheritsLoose.js",
    "../node_modules/@babel/runtime/helpers/esm/assertThisInitialized.js",
    "../node_modules/memoize-one/dist/memoize-one.esm.js",
    "../node_modules/react-window/dist/index.esm.js",
    "../node_modules/react-virtualized-auto-sizer/dist/index.esm.js",
    "../src/views/GridView.js",
    "../src/views/ListView.js",
    "../node_modules/react-fast-compare/index.js",
    "../node_modules/deepmerge/dist/es.js",
    "../node_modules/tiny-warning/dist/tiny-warning.esm.js",
    "../node_modules/react-is/cjs/react-is.production.min.js",
    "../node_modules/react-is/cjs/react-is.development.js",
    "../node_modules/react-is/index.js",
    "../node_modules/hoist-non-react-statics/dist/hoist-non-react-statics.cjs.js",
    "../node_modules/@willynlabs/formik/dist/formik.esm.js",
    "../src/components/Fields.js",
    "../src/views/FormView.js",
    "../src/components/ListItem.js",
    "../src/components/OverlayContentContextProvider.js"
  ],

That is the complete source list for this build.

I am not 100% certain, but I don't believe that webpack is able to "merge" the copy of lodash-es that is included in my library with the published Formik module incorporated, and it would not be able to eliminate it even with tree shaking.

I am not submitting a PR for this, and I am happy to simply use my fork of the code for now to workaround it.  As @TrySound mentions above, these individual modules are deprecated, so this will eventually resolve itself.

Separately, I did verify that using the `import {clone, toPath} from "lodash";` is tree shakable, which I confirmed with some of my other modules/libraries.

Hope it helps (someone at some point).

TrySound:
This will load large lodash bundle.
```
import {clone, toPath} from "lodash";
```
https://unpkg.com/browse/lodash@4.17.15/lodash.js

Also I never say individual exports are deprecated they will never be because of side effects in some modules. Only individual packages like `lodash.clone` are deprecated, not modules like `lodash/clone`.

> I am not 100% certain, but I don't believe that webpack is able to "merge" the copy of lodash-es that is included in my library with the published Formik module incorporated, and it would not be able to eliminate it even with tree shaking.

This actually should be solved by yarn/npm. Single version is hoisted to the root. Otherwise you probably installed lodash-es and formik in different time. Remove both and install them again to get the latest hoisted version.

marcosparagna-eb:
any update on this?

willynlabs:
I stand by my assessment, **which is specific to the use of Formik in a module using Rollup**.  In fact, after the exchange above, I spent a day confirming that using the individual imports in Formik prevents Rollup from recognizing lodash (or lodash-es) entries as "external", and it winds up embedding the individual module imported in the manner described above in my intermediate library.

The ultimate issue I faced was that these lodash-es imports are then not shared across my other modules, and some of those utilize the same common lodash-es methods of clone, toPath, etc.

If I use ```import {clone, toPath} from "lodash";``` and add lodash-es as an external module in my rollup.config, then all of my modules can share the same instance of lodash-es, and ultimately, the  webpack tree shaking takes this down to just those modules referenced, which is just a handful across all of my modules (including Formik).

As I mentioned in my post, I am not submitting any pull request for changes to Formik, and have opted to put this on a local branch for my purposes instead.

BTW, I don't believe this is limited to lodash and Formik, but applies to any module that does individual imports of this sort, and then uses Rollup to create another module that incorporates them.  I guess you could consider this to be a Rollup issue, but I don't know if it is worth posting as such (at this point).

TrySound:
lodash and lodash-es are different packages. This is all problem. If you use lodash-es code is shared. If you use lodash packages you get two packages in your bundle. There is nothing to fix. But yeah, you would not see any problem if lodash package provided es modules support out of the box.

Issue #1543, opened 2019-05-23
testing form with validationSchema


mfolnovic:
### Current Behavior

I have a test that basically screenshots a form after removing focus from field that is required (while not filling in any value). For validations, I'm using Yup.

During `fireEvent.blur(textField);`, I get warning:
```
Warning: An update to Formik inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */
```

I'm also trying to screenshot page after blur, expecting to see validation message. From my experiments, I need to do: `await waitForDomChange({ input })` to get correct screenshot.

I've also noticed there's no warning if I don't have `validationSchema`.

I assume this is connected to https://github.com/jaredpalmer/formik/issues/1524 ?

### Expected behavior

No warning.
No need for `waitForDomChange`.

### Reproducible example

https://codesandbox.io/s/agitated-lalande-5nhh0

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.1-rc.0 |
| React            | 16.8.6 |
| TypeScript       | 3.4.5 |
| Browser          | Chrome |
| npm/Yarn         | Yarn |
| Operating System | Linux |

mfolnovic:
I'd like to add this also happens with `validate` (instead of `validationSchema`), e.g.:

```
const validate = (values) => {
  const errors = {};

  if (!values.test) {
    errors.test = "required";
  }

  return errors;
};
```

mfolnovic:
And I forgot to mention that, even surrounding `fireEvent.blur` with `act` doesn't help, e.g.:

```
act(() => {
  fireEvent.blue(input);
});
```

cmelion:
I'm seeing the same issue when using a validation schema.
https://github.com/cmelion/react-cra-demo/tree/cfulnecky/formik

mayteio:
I am having the same issue when trying to change an input value:
`fireEvent.change(getByLabelText(/email/i), {target: {value: fakeUsername}});`

Obviously, wrapping this in `act(() => {})` doesn't do anything to mitigate the problem.

2Steaks:
Have you tried this?

```
await act(async() => {
    fireEvent.change(getByLabelText(/email/i), {target: {value: fakeUsername}});
});
```

duncanleung:
@2Steaks Thanks for that snippet. Works for me.

[...]

PR #2006, opened 2019-11-08
Batch async state updates

maxdavidson:
Async validations trigger `act` warnings during testing:

```
an update to Formik inside a test was not wrapped in act
```

When validating, the next error state is updated at the earliest at the end of the current tick. These will cause act warnings unless they take place during the window of an async `act` call. It's hard to know the timing of these, so it should be Formik's responsiblity to prevent these warnings.

What `act` does is basically forcing updates to be batched. React automatically batches updates in event handlers and component lifecycles/effects, but for other updates we need to use `unstable_batchedUpdates`. Unfortunately it's not exported by React itself so we need to use some requires with fallbacks to support different renderers.


src/utils.ts
```diff
@@ -169,3 +169,33 @@ export function setNestedObjectValues<T>(
 
   return response;
 }
+
+export type Batch = (cb: Function) => void;
```
jaredpalmer: Add a comment about why we need this inside of source code for newcomers.

wachunga:
These warnings are causing a lot of clutter in our test output. Looks like this issue is related: https://github.com/jaredpalmer/formik/issues/1543

dreyks:
any updates on this?

BravenxX:
any updates?

maxdavidson:
I don't think the `require` in a try-catch will be portable enough, since it won't work in ES module-only contexts. Dynamic imports could work, but that would make it async. 

React-redux does it like this to support React Native:

https://github.com/reduxjs/react-redux/blob/v7.2.3/src/utils/reactBatchedUpdates.js
```js
export { unstable_batchedUpdates } from 'react-dom';
```

https://github.com/reduxjs/react-redux/blob/v7.2.3/src/utils/reactBatchedUpdates.native.js
```js
export { unstable_batchedUpdates } from 'react-native';
```

The file structure would need to be preserved for this trick to work though, or you would have to output a separate bundle for RN.

Alternatively, using some sort of dependency injection based on the entry point could also do it (formik-dom/formik-native?)

PR #2012, opened 2019-11-11
Improve type checking around `setFieldValue`

carlmw:
This strengthens checks around the value passed to `setFieldValue` — using a type argument to ensure that value matches the type of `field` in `Values`.

**before type checker does not complain**

Screen Shot 2019-11-11 at 17 05 27

**after type checker complains that name must be a string to match the inferred type of `initialValues`**

Screen Shot 2019-11-11 at 17 04 13

Screen Shot 2019-11-11 at 17 04 22


stunaz:
Does this fix this issue ? Seems like the typings does not work with nested fields
image

carlmw:
@stunaz I'm afraid not. I might have a look to see if that is even possible with TypeScript.

jaredpalmer:
It's not possible with nested fields, this is why we haven't done this in the past.

Issue #1948, opened 2019-10-29
setFieldTouched, setFieldValue, etc. not working with arrays in Typescript anymore


mikecousins:
### Current Behavior
![image](https://user-images.githubusercontent.com/725614/67737163-cb8be400-f9cf-11e9-9b39-f9afb5bd9cc8.png)

My initial values is a property called targets which contains an array of UITargetPositions.

![image](https://user-images.githubusercontent.com/725614/67737194-e2323b00-f9cf-11e9-8abb-8e2533920d67.png)

Calling setFieldValue or setFieldTouched results in a Typescript error that the string isn't one of the possible values.

### Expected behavior

No error should be thrown.

### Reproducible example

I can't get it to work with Typescript. For some reason when I changed it to .tsx it just gives me a white screen now, but here you go:
https://codesandbox.io/s/formik-codesandbox-template-iru3w

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.3
| React            | 16.11.0
| TypeScript       | 3.6.4
| Browser          | all
| npm/Yarn         | yarn 1.19.1
| Operating System | Win 10 Pro

sa9sha9:
i'm facing same issue :(

![スクリーンショット 2019-10-31 11 29 29](https://user-images.githubusercontent.com/14252323/67916850-753ec100-fbda-11e9-91f2-d754db05456b.png)

fiendfiend:
it appears that setFieldValue is now attempting to be a bit too smart in terms of inferring the possible field values based on the generic form values parameter.

In the case of nested fields, this no longer works, since it doesn't attempt to traverse the possible object graph in order to derive all possible string permutations for fields - nor do I think it can or should achieve such a thing.

fenech:
The issue is  https://github.com/jaredpalmer/formik/blob/master/src/types.tsx#L89
```typescript
setFieldValue(field: keyof Values & string, value: any, shouldValidate?: boolean): void;
```
`keyof Values` only includes top-level properties, so cannot be used for any nested structures.

The only way I can think of to make this work (without reverting to just `field: string`) would be to add a typed element further down the tree, which exposed a "scoped" version of `setFieldValue`:
```typescript
<Formik<Values>>
  ...
  <FieldValueSetter<NestedValues> name="foo.bar[0].baz">    
    {({ setFieldValue }: FieldValueSetterProps<NestedValues>) => ...
  </FieldValueSetter>
</Formik>
```
[...]

PR #2039, opened 2019-11-15
Fix setFieldValue/Error/Touched types for nested objects and arrays

stevenm-alintaenergy:
Fixes: https://github.com/jaredpalmer/formik/issues/1948


fenech:
@stevenm-alintaenergy why did you end up closing this? Was there something wrong with it?

The only thing I'd say is that `keyof Values | string` is redundant and can be replaced with just `string`, but other than that, it looks fine to me.

PR #2040, opened 2019-11-15
package.json: remove unused lodash-es dependency

hlian:
I was using `yarn why` on my project, and I noticed that lodash-es seems to be a spurious dependency. (Formik only ever imports from `'lodash'`.)


jaredpalmer:
`lodash-es` _is_ used. It is optimized within tsdx to allow for treeshaking. 

See https://github.com/jaredpalmer/tsdx#using-lodash

PR #2045, opened 2019-11-18
Note about lack of file support

codeinaire:
It was my assumption (like many others as per the issue I referenced above) that this library would handle file uploads. Based on this assumption I thought I was using the library wrong when the file wouldn't upload correctly. It would've saved me a lot of time and effort if there was a note somewhere in the docs clearly stating that it cannot handle file uploads. So, this is that note.

This seemed the most obvious place to put it. Let me know if y'all would rather it elsewhere.

### DIFF
--- a/docs/api/field.md
+++ b/docs/api/field.md
@@ -8,6 +8,8 @@ custom_edit_url: https://github.com/jaredpalmer/formik/edit/master/docs/api/fiel
 attribute to match up with Formik state. `<Field />` will default to an HTML
 `<input />` element.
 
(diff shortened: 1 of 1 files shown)


jaredpalmer:
How _should_ files work out of the box, in your opinion? 

Files are highly specialized. Even in your example, you don’t deal with multi file uploads, or image preview(s), and so creating a specific default handler seemed presumptive.  

Not opposed with this PR, but am curious what the behavior _should_ be? Perhaps we provide some upload helper component(s)?

Issue #1974, opened 2019-10-30
Performance issues in v2


mvestergaard:
### Current Behavior
Changes to any field (both Field and FastField) causes all fields in the form to re-render twice.

### Expected behavior

Changes to fields should only cause fields to re-render once, and more over, FastField should only re-render when changes are made to that field specifically.

### Reproducible example
Here's the same scenario demonstrated in both v2 and v1
v2.0.3 https://codesandbox.io/s/vigorous-wood-esy70
v1.5.8 https://codesandbox.io/s/keen-bouman-sx5r0

### Additional context
According to the react dev tools, the extra renders happen because "Context changed".

Possibly same problem as reported on #1759

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.3
| React            | 16.8.4
| TypeScript       | 3.6.2
| Browser          | Chrome
| npm/Yarn         | yarn
| Operating System | windows

danpantry:
Yes, looking into this I'm seeing that `onChange` from `useField()` is changing twice every single time a value in the form changes. This can be seen in [this][sandbox] sandbox - Note that the counter displayed increases twice every time you enter a character in the text box. That counter is only incremented when `onChange` changes.

This prevents using some simple optimisations using `React.memo()`.

[sandbox]: https://codesandbox.io/s/beautiful-river-qqzvb

PR #2069, opened 2019-11-26
Avoid unnecessary re-renders by comparing errors to previous errors

mvestergaard:
This addresses part of #1974

Each time any field is touched (onChange, onBlur) the errors on the context changed, causing all fields to re-render once more than needed (2 renders per change)

This fix will compare the new errors object being assigned to the existing one, and only change the state if they are different.

FastField is still fundamentally broken, because it will re-render on each change to the context. Should be fixed in #2070


mvestergaard:
I tried my best to write a test that reproduces the problem, but it's really difficult, if not impossible in a "fake context".
Use the FastField story in the formik storybook to compare the behavior to before.

Issue #2042, opened 2019-11-16
FastField re-renders when other fields are changing


rumtraubenuss:
### Current Behavior

`FastField` in `2.0.4` re-renders when other fields update.

### Expected behavior

`FastField` only re-renders when it's own field updates.

### Reproducible example

https://codesandbox.io/s/formik-example-3tzwl

In this condesandbox you can see the effect. When changing the first input field, the second field (which is a FastField) renders as well (Every render logs a message to console, and the console's badge count increments). When you change the Formik version to `1.5.8` the `FastField` does not render, when changing the first field, as expected.

### Additional context

Is this a bug? Am I missing something here? I checked the Formik 2 Migration Guide but nothing is mentioned that the `FastField` behaviour changed.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.4
| React            | 16.11.0
| TypeScript       |
| Browser          | Chrome
| npm/Yarn         |
| Operating System |

rumtraubenuss:
The issue can be reproduced in formik's `storybook`. All three example `FastField` components render when the form's values are changing. I tried to debug the `FastField` class. What I learned is that `shouldComponentUpdate` is never called before rendering. Anyone an idea how this can be?

hrgui:
Running into this issue as well on Formik 2.0.6

Repo: https://codesandbox.io/s/formik-example-mtw4r (just click on one of the fields, and click out)

## Couple of observations:

1. shouldComponentUpdate never gets called at all. I tried putting breakpoints in every single line, it never gets invoked. I believe this may be caused by https://github.com/jaredpalmer/formik/pull/1916. 

@rumtraubenuss is correct that version 1.5.8 does not have the issue. It can be seen in the following codesandbox:

https://codesandbox.io/s/formik-example-2pppc

---

2. All fields get rendered twice due to Form validations due to https://github.com/jaredpalmer/formik/blame/master/src/Formik.tsx#L321-L332 (first call due to react, 2nd call due to `unstable_runWithPriority`.) 

This can be solved by doing:
```jsx
<Formik
      validateOnBlur={false}
      validateOnChange={false}
      validateOnTouched={false}
```
^ https://codesandbox.io/s/formik-example-gzlu1 <- compare this with the repo

jaredpalmer:
Fixed in 2.1

Issue #1759, opened 2019-08-18
FastField is rendering on every value change


thupi:
### Current Behavior
The FastField component renders even though that another values is being changed. 

### Expected behavior
FastField components should only rerender when the releated value is changed.

### Reproducible example
The very wired part of this bug is that this works as expected on codesandbox. However, when i make a new clean project using CRA it doesn't :-/

Codesandbox: https://codesandbox.io/s/formik-2-rwlhv
Clean CRA project: https://github.com/thupi/formik-fast-field-sample

### Additional context
Look for the console.log messages in both of these samples i provided above. The CRA sample console.logs both fields on every change while codesandbox works as expected.

I Thought this was due to typescript so i tried with at clean js project too but made no difference.

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.0.rc-12
| React            | 16.9.0
| TypeScript       | 3.5.2
| Browser          | Chrome
| npm/Yarn         | Yarn 1.15.2
| Operating System | MacOS Mojave

schrapel:
I created a similar issue here https://github.com/jaredpalmer/formik/issues/1739

thupi:
@schrapel I see, however in my case it is quiet consistant with only rendering all fields. it never rendered 6 times in my case :-) But indeed very similar :-)

johnrom:
I'll add some context here for whever wants to tackle this. Basically the only difference between FastField and a regular Field (if I understand it correctly, code is not in front of me!) is that when FastField itself changes, it keeps its changes locally and only propagates them to Formik when the field is blurred. This means changes to a FastField do not have a performance detriment to the Form at large. However, FastField _must_ react to any changes that come from Formik's state, or its own state will be stale. Therefore, if you have a separate, plain `Field` which updates Formik on every change, `FastField` will have to re-render each time. This is because `FastField`, like `Field`, gets a lot of objects from Formik's state, and checking if they are _actually different_ will actually slow down faster forms -- it's a trade-off. If implemented, these checks will need to be _perfect_ in order to prevent fields from getting stale.

This, of course, isn't ideal, and is a great low hanging fruit where we can improve.

@schrapel your issue is slightly different, and I'll comment on your issue.

wpq0:
Seeing that FastField is now just [an alias for Field
[...]

PR #2070, opened 2019-11-26
Fix unnecessary rendering of FastField

### LINKED ISSUE #1974: Performance issues in v2

mvestergaard:
Whenever Context changes, shouldComponentUpdate is not called, which caused FastField to always re-render whenever Context changes
This adds a wrapper around FastField that grabs the Context, and passes it down as props

Fixes #1974, #2042, #1759

May depend on #2069 to be entirely fixed

### DIFF
--- a/src/FastField.tsx
+++ b/src/FastField.tsx
@@ -9,8 +9,8 @@ import {
 } from './types';
 import invariant from 'tiny-warning';
 import { getIn, isEmptyChildren, isFunction } from './utils';
-import { FormikContext } from './FormikContext';
 import { FieldConfig } from './Field';
+import { connect } from './connect';
 
 type $FixMe = any;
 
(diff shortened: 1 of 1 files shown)


mvestergaard:
I tried my best to write a test that reproduces the problem, but it's really difficult, if not impossible in a "fake context".
Use the FastField story in the formik storybook to compare the behavior to before.

src/FastField.tsx
```diff
@@ -36,12 +36,27 @@ export type FastFieldAttributes<T> = GenericFieldHTMLAttributes &
  * Custom Field component for quickly hooking into Formik
  * context and wiring up forms.
  */
-export class FastField<Values = {}, Props = {}> extends React.Component<
-  FastFieldAttributes<Props>,
+export function FastField<Values = {}, Props = {}>(
```
jaredpalmer: Can you just use connect()?

mvestergaard: I tried, but typescript and HOC's don't play well, can spend time trying to make it work if you insist.

mvestergaard: Just tried looking at it again. I can't figure out a way to use connect() and still preserve the generic arguments.

mvestergaard: I changed it to use connect() the same way as prior to 2.0, but this approach kind of makes it pointless for FastFieldInner to have generic arguments in the first place.

Your call.

mvestergaard:
@jaredpalmer Can you take another look at this please? And possibly publish a new release including this and #2069 
Large forms are presently close to unusable

jaredpalmer:
Publishing meow

mhaylock:
@jaredpalmer what is the status of that release?

I went to look for the release attached to your comment above but I see the last release was 2.0.6, 17 days ago.

Issue #2050, opened 2019-11-20
Formik v2.05: setIn return empty error field within errors object when value is empty array


wingedearth:
### Current Behavior

I'm using a Yup validation schema with a single field that is an array. Specifically:

```
yup.object().shape({
	myField: yup.array()
		.min(minimumContent, `At least ${minimumContent} link(s) must be added.`)
		.max(maximumContent, `No more than ${maximumContent} link(s) allowed.`)
});
```

I use FieldArray to manage myField.
When I have no errors (i.e. errors: {}), and I call FieldArray's "remove" on an item in the array, Formik calls setFormikContext, which passes an empty array as the "value" to setIn when generating the errors object.

The current behavior is that setIn is returning: { myField: [ ] } as the errors, i.e. myField pointing to an empty array of errors.

### Expected behavior

I expect setIn to return empty object, { }, as the errors to setFormikState when there are no validation errors.

### Reproducible example

To reproduce, open the following sandbox and click on "remove" next to one of the content items under "myField". You'll see errors change from { } into { myField: [ ] } and you'll see isValid change from true to false.

https://codesandbox.io/s/formik-codesandbox-template-qtzzv?fontsize=14&hidenavigation=1&theme=dark

### Suggested solution(s)

The solution is for the setIn function to be able to handle an empty array as the value argument.

If "value" argument is an array with length === 0, then use undefined as value instead of using empty array.

### Your environment

| Software             | Version(s) |
| ---------------- | ---------- |
| Formik                 | 2.0.5
| React                   | 16.12.0
| ReactDOM .        | 16.12.0
| npm/Yarn            | yarn 1.17.3
| Browser .            | Chrome 78.0.3904.87

wingedearth:
I updated this issue because I tested and confirmed that this issue is in both Formik 2.0.4. and 2.0.5

aludvigsen:
I have the same issue. It was introduced in `2.0.1.rc2` version when validation was rewritten i assume. Version `2.0.1.rc1` does not have this issue at least. 

You dont even need a `validationSchema` to trigger the bug. The errors object will be populated with an empty array either way. This happens when using the arrayHelpers `pop` or `remove`.

wingedearth:
@aludvigsen  - I also noticed that the problem persists even when removing the validationSchema prop completely.

aludvigsen:
Here is another example without Yup.

[...]

PR #2072, opened 2019-11-26
handle an empty array value argument to setIn

wingedearth:
setIn checks for an undefined value argument but does not check for empty array. This is causing the bug described in:

https://github.com/jaredpalmer/formik/issues/2050

This change resolves the bug by treating a value of empty array like an undefined value for purposes of removing arrays.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -4,6 +4,10 @@ import * as React from 'react';
 
 // Assertions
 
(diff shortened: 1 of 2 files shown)


Pasalietis:
now this causing bug:
setFieldValue('field', []);
deletes key 'field'

PR #2090, opened 2019-12-06
add validationSchema to FormikContext

ldicocco:
```useFormikContext``` doesn't return ```validationSchema,``` although the type says it should do and the ```connect```in v1 did.
This PR fixes the problem and adds a specific test for this case.

### DIFF
--- a/src/Formik.tsx
+++ b/src/Formik.tsx
@@ -564,8 +564,10 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         if ((eventOrTextValue as React.ChangeEvent<any>).persist) {
           (eventOrTextValue as React.ChangeEvent<any>).persist();
         }
-        const target = eventOrTextValue.target ? (eventOrTextValue as React.ChangeEvent<any>).target : (eventOrTextValue as React.ChangeEvent<any>).currentTarget;
-        
+        const target = eventOrTextValue.target
+          ? (eventOrTextValue as React.ChangeEvent<any>).target
+          : (eventOrTextValue as React.ChangeEvent<any>).currentTarget;
+
         const {
           type,
           name,
@@ -914,7 +916,7 @@ export function Formik<
   ExtraProps = {}
 >(props: FormikConfig<Values> & ExtraProps) {
   const formikbag = useFormik<Values>(props);
-  const { component, children, render } = props;
+  const { component, children, render, validationSchema } = props;
   React.useEffect(() => {
     if (__DEV__) {
       invariant(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by josejulio** [comment]
Also hitting the same issue, if anything is needed to help, please let me know.

**Comment by desfero** [comment]
@ldicocco also it probably worth to add `validate` the same way you did with `validationSchema` because if you look at types
```
export declare type FormikContextType<Values> = FormikProps<Values> & Pick<FormikConfig<Values>, 'validate' | 'validationSchema'>;
``` 
you can see that both `validate` and `validationSchema` are used. 

@ldicocco @josejulio is there any workaround to get `validationSchema`?

@jaredpalmer is it possible to merge this one and release as otherwise formik is broken in case someone is using yup metadata to get the desired UI

**Comment by SamKirkland** [comment]
This was actually a breaking change in `v2`.
I'm updating from `v1.5.8` which used to provide validationSchema in `<FormikConsumer>`

[Example 1](https://codesandbox.io/s/formikconsumer-validationschema-example-di6zj) - change the imported formik version to `v2.1` or above and validationSchema will no longer be in context
[Example 2](https://github.com/jaredpalmer/formik/blob/606e10cef01d09b2ebaf95d7edcb2910aec29572/examples/Debug.js#L29)

**Comment by egoreburial** [comment]
cmon please merge it! There is no chance to dynamically check if come field is required or not and render it to UI !!!

**Comment by Maxim-Chugaev** [comment]
please merge it !

**Comment by Maxim-Chugaev** [comment]
@josejulio What about workaround ?

File: packages/formik/test/FormikContext.test.tsx [hunk 1]
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { render } from 'react-testing-library';
+import { Formik } from '../src/Formik';
+import { useFormikContext } from '../src/FormikContext';
+
+describe('FormikContext', () => {
+  describe('useFormikContext', () => {
+    it('should return validationContext if set', () => {
+      const validationSchema = 'validationSchema';
```
johnrom, line 9: is `string` valid for validation schema? thought it had to be a Yup object. Other tests use `{ validate: jest.fn(() => Promise.resolve()) }`.

josejulio:
> @josejulio What about workaround ?

I used the validation schema that I knew was in place directly.

IevaDrauge:
Can someone please merge/fix this? It is a very crucial missing feature right now...

Foarvad:
I have described one possible workaround here: https://github.com/formium/formik/issues/2092#issuecomment-742412094

raphaelpc:
Hello! Any news on this? Thanks!

Issue #146, opened 2017-08-30
Scroll to first invalid element


cassln:
Is there nice and recommended way to scroll to first invalid form element with formik, when form is submitted?

Andreyco:
I would go for component which will react to `errors` prop change.

1. watch change either in `componentWillReceiveProps` or `componentDidUpdate`
2. when change is detected, find first/topmost error message and scroll to message

jaredpalmer:
Yeah `componentDidUpdate` is probs the best way to do this as it is a sideeffect.

cassln:
Side effect in `componentDidUpdate` looks good. But how can I detect in `componentDidUpdate` that it was exact submit, not change. (I'm using `validateOnChange` option)

jaredpalmer:
@cassln There may be more elegant ways to do this, but I would do something like:

```js
// ... 

componentDidUpdate(prevProps) {
   // if form was submitting, but now is not submitting because it is invalid
   if (prevProps.isSubmitting && !this.props.isSubmitting && !this.props.isValid) {
      // and assume you've added refs to each input: `ref={ i => this[name] = i}` , then...
      this.[Object.keys(this.props.errors)[0]].focus() 
  }
}
```

- I think `.focus()` will trigger a scroll in most browsers, maybe janky tho
- Not 100% certain which order your `errors` keys will be in, would depend on your validation function. I believe Yup maintains shape-order (worth checking).

cassln:
@jaredpalmer this condition is working perfect. Thank you very much.

I don't use `.focus()`, because I have complex form components like gallery. I use `div` with `data-scroll-anchor` attribute and find first `div` with the attribute. But I think this is case specific part.

Issue is covered, so i'm going to close issue now. :)

lardissone:
@jaredpalmer this solution is only available on formik enhanced forms right? Is there a similar way to implement this in render props way?

Grsmto:
@lardissone you can do that as well with render props I think. Just create a `ScrollToError` component which contain the above code and after you use it this way:

```
<Formik
  render={formikProps => (
    <form>
      <ScrollToError {...formikProps} />
    </form>
  )}
/>
```

cloud-walker:
We need to be able to know at the same time about three things:
- the errors object
- the isSubmitting state
- the inputs refs

This is not actually possible at the moment, because when you submit the form, formik validate is fired, and the isSubmitting state remain false (when some field is invalid ofc)

[...]

PR #2103, opened 2019-12-10
hotfix: Revert throwing combined errors

jaredpalmer:
I confused myself this morning with 2.0.7. The change to submission is making formik unusable outside of react native. 

Revert changes in #1843

### DIFF
--- a/packages/formik-native/test/blah.test.ts
+++ b/packages/formik-native/test/blah.test.ts
@@ -1,7 +1,5 @@
-import { sum } from '../src';
-
 describe('blah', () => {
   it('works', () => {
-    expect(sum(1, 1)).toEqual(2);
+    expect(1).toEqual(1);
   });
 });
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -564,8 +564,10 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         if ((eventOrTextValue as React.ChangeEvent<any>).persist) {
           (eventOrTextValue as React.ChangeEvent<any>).persist();
         }
-        const target = eventOrTextValue.target ? (eventOrTextValue as React.ChangeEvent<any>).target : (eventOrTextValue as React.ChangeEvent<any>).currentTarget;
-        
+        const target = eventOrTextValue.target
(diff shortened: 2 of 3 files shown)


HristoKolev:
You should publish a new version, the current is obviously not working.

Now I have to pin the package to 2.0.6 in package.json. (And then forget to unpin it...)

benb1983:
+1

uriklar:
I also found out my form was breaking in 2.0.8 (updated from 2.0.7)

This is how I use formik in my app:
```
<Formik
    initialValues={...some initial values}
    validationSchema={Yup.object().shape({
        ...some validations
    })}
    onSubmit={submit}>
    {formik => {
        return (
            <Form>
                ..some regular input fields
                    <button
                        type="submit"
                        onClick={e => {
                            e.preventDefault();
                            formik.submitForm().catch(errors => {
                                formik.setErrors(errors);
                            });
                        }}
                    />
                </div>
            </Form>
        );
    }}
</Formik>
```

I know this isn't the exact documented usage, but this is the only thing that worked (the throw line that was removed was causing the form to not catch it when there were validation errors)

Anyway, maybe I need to try a different implementation of the submit, but just FYI, this setup is broken on 2.0.8 (neither the `onSubmit` callback and then errors catch are being called)

fabb:
Too bad this was reverted. I hoped there now was a better way to have side effects when submission was cancelled due to validation errors than this wrapper component:

```
export type FormikSubmitValidationErrorHandlerProps<Values> = FormikProps<Values> & {
    callback: () => void
}

export class FormikSubmitValidationErrorHandler<Values> extends React.Component<FormikSubmitValidationErrorHandlerProps<Values>> {
    componentDidUpdate(prevProps: FormikSubmitValidationErrorHandlerProps<Values>) {
        // formik currently does not call any callback in case submitting the form was aborted due to client validations, therefore we need a workaround
        if (prevProps.isSubmitting && !this.props.isSubmitting && !this.props.isValid) {
            this.props.callback()
        }
    }

    render() {
        return this.props.children
    }
}
```

https://github.com/jaredpalmer/formik/issues/146#issuecomment-327222858

PR #2107, opened 2019-12-11
Fix nav search box on big screens

madcido:
Search box is broken on the website (desktop version), it's overflowing out of the navbar


jaredpalmer:
Thanks!

PR #2122, opened 2019-12-13
Fix documentation for withFormik async validate method

mistadikay:
Banged my head against the wall until realized that documentation was not updated after #1575 

Here is the fix

-----
[View rendered docs/api/withFormik.md](https://github.com/mistadikay/formik/blob/master/docs/api/withFormik.md)


jaredpalmer:
Oof! I'm sorry about that!

Issue #1907, opened 2019-10-24
[v2] useField would love to have deprecated setFieldValue


Nomeasmo:
### Current Behavior

Currently useField returns onChange, which expects synthetic events and calls after some lengthy parsing of the event the setFieldValue method.

### Desired Behavior

In order support easier creation of custom controls as well as the use from third party libraries it would be advantageous to have a setFieldValue method again.

### Suggested Solution

I'd extend the useFields meta object with a reference to that function.

### Who does this impact? Who is this for?

People using non standard inputs, like custom fields or if third party library is not completely doing what expected.

Some examples:
```js
// Semantic UI 
<Dropdown
   {...field}
   onChange={(_, { value }) => field.onChange({ target: { id, name, value } })}
   ...
/>
// or
const onModifiedChange = (event) => {
    // Not sure either the synthetic event of semantic ui checkbox is wrong or the
    // getValueForCheckbox from Formik is not ok.
    event.target.value = event.target.checked.toString();
    field.onChange(event);
};
return (
    <Checkbox
       onChange={onModifiedChange}
// Or handcrafted
     <input
         type="button"
         onChange={() => field.onChange({ target: { id: fieldId, name, value } })}
```
### Describe alternatives you've considered

See above

### Additional context

n/a

jaredpalmer:
Word. For now, just use `useFormikContext()`

jaredpalmer:
We could do this:

```tsx
// Add within useFormik()
const getFieldHelpers = React.useCallback(
  (name: string) => {
    return {
      setValue: (value: any) => setFieldValue(name, value),
      setError: (value: any) => setFieldError(name, value),
      setTouched: (value: boolean) => setFieldTouched(name, value)
    };
  },
  [setFieldError, setFieldValue, setFieldTouched]
);
```

```diff
const MyInput = props => {
-  const form = useFormikContext();
-  const [field, meta] = useField(props);
+  const [field, meta, helpers] = useField(props);
  return (
    <input
      {...field}
      onChange={e => {
-        formik.setFieldValue(props.name, e.target.value);
+        helpers.setValue(e.target.value);
      }}
    />
  );
};
```

killjoy2013:
Hi,
My project starts to complain about `setFieldValue` after 2.0.1 upgrade.

Can I simply use 
`formikProps.setValues({[e.target.name]: e.target.value})`
instead of 
`formikProps.setFieldValue(e.target.name, e.target.value);`

[...]

PR #2124, opened 2019-12-14
Add FieldHelpers object to useField

drivasperez:
Implements the change suggested in #1907.

Adds a third object ("helpers") to the array produced by useField. This
exposes three helper functions to allow you to imperatively change the
field in question:

  1) setValue
  2) setTouched
  3) setError

E.g.:
```
const MyInput = props => {
  const [field, meta, helpers] = useField(props);

  return (
    <input
      {...field}
      onFocus={() => {
        helpers.setTouched(true);
      }}
      onBlur={e => {
        if (e.currentTarget.value === "hi") {
          helpers.setError("Hey");
        }
      }}
      onChange={e => {
        helpers.setValue(e.target.value);
      }}
    />
  );
};
```


jaredpalmer:
Add docs and this is lgtm

slorber:
+1000, was just reimplementing this in my code, as many times it's not possible to use handleChange(domEvent) (RN for example, but also dropdowns or more complex form items)

drivasperez:
@jaredpalmer How does this look?

VanTanev:
@drivasperez @jaredpalmer Was there a particular reason you decided not to include the `shouldValidate?: boolean` optional attribute for these helpers? It seems like it would have made sense to have the full functionality of the `FormikContext` methods.

drivasperez:
Oversight on my part, I think. I can add it.

MikeSuiter:
@drivasperez I just started converting from Formik 1.5.8 to 2.1.2 and since you implemented `setValue` have a question. I have our custom `Field` which adds format and normalize functionality to a field which worked in 1.5.8. Now I'm trying to convert it and `setValue` puts it in an infinite loop re-rendering `Field`. 

Here is my component and `Component` will be a formik-material-ui component. It works just fine unless the format property is set which causes the infinite loop. The parent component doesn't re-render during the infinite loop. Any ideas?

```
const Field = ({ component: Component, name, format, normalize, useContext, onChange, validate, ...otherProps }) => {
  const [field, meta, helpers] = useField(name);

  console.log('Field render');

  if (format) {
    helpers.setValue(format(field.value));
  }

  return <Component name={name} {...otherProps} />;
};
```

drivasperez:
@MikeSuiter `setValue` triggers a re-render, so unconditionally calling it in the render function will cause an infinite loop. You could do something like this:

```jsx
const Field = ({ component: Component, name, format, normalize, useContext, onChange, validate, ...otherProps }) => {
  const [field, meta, helpers] = useField(name);

  React.useEffect(() => {
    if (format && field.value !== format(value)) {
      helpers.setValue(format(value));
    }
  })

  return <Component name={name} {...otherProps} />;
};
```
But I think more likely you want something like this:
```jsx
const Field = ({ component: Component, name, format, normalize, useContext, onChange, validate, ...otherProps }) => {
  const [field, meta, helpers] = useField(name);
  
  function handleChange(value) {
    if (format) {
      helpers.setValue(format(value))
    } else {
      helpers.setValue(value);
    }
  }

  return <Component name={name} {...otherProps} onChange={handleChange} />;
}
```
Does that help?

MikeSuiter:
@drivasperez Thanks for the response and I thought it would be easy to do your second solution but the Formik Material-UI library doesn't pass onChange down to Formik. This is my first dive into hooks so I have some learning to do!

https://stackworx.github.io/formik-material-ui/docs/guide/faq

In your first solution you're using value but I'm not sure where that's coming from.

drivasperez:
Oh, sorry, that was a typo. I meant `field.value` instead of `value`. The idea is to check if the field’s value is formatted and, if not, to apply the effect.

Issue #2130, opened 2019-12-16
Removing the last element in a FieldArray sets array value to undefined


jsteiner:
### Current Behavior

If you `remove` the last element in a `FieldArray`, it sets the new value at that `name` to `undefined`.

Calling `remove(0)` on:

```
{
   foos: [{ bar: 1 }]
}
```

becomes:

```
{}
```

### Expected behavior

I'd expect the value to be set to an empty array.

Calling `remove(0)` on:

```
{
   foos: [{ bar: 1 }]
}
```

should become

```
{
   foos: []
}
```

This behavior was unexpected and let to some errors calling methods on undefined, where my application otherwise guarantees the array's presence.

### Reproducible example

https://codesandbox.io/s/formik-codesandbox-template-9q9db

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.8

Mindtraveller:
Hi, have similar issue when setting value to empty array e.g.
```
{
  foo: 'bar',
  lang: ['en']
}
```
setting value `[]` to `lang` field results in:
```
{
  foo: 'bar'
}
```

It's reproducible on Formik **v2.0.7**.
Works fine on **v2.0.6**

iobuhov:
This is the problem: https://github.com/jaredpalmer/formik/blob/448e3e7f5477400a8f77a29f2065e2b8f9a96d01/packages/formik/src/utils.ts#L130

`setIn` take care of your "empty" field and silently removes it from values, which is really confusing.

iobuhov:
#2050

imdadul:
I am also having the same problem,
Whenever i am removing the last item from my array, I am having errors in my code

```
// obj.foo is undefined
{ 
   obj.foo.map(......)
}
```

I think this is a problem.

alexcracea:
I reproduced this as well. This issue is affecting the behavior of `setFieldValue` in general as in my case I did not have any `<FieldArray>`, in my case it was a list of checkboxes that are throwing an error every time I uncheck everything.

https://codesandbox.io/s/formik-codesandbox-template-kyc9x?fontsize=14&hidenavigation=1&theme=dark

I found a workaround for this - in my case I got it temporarily fixed by rolling back package version version of formik from version 2.0.8 to version 2.0.6

meyerds:
[...]

PR #2144, opened 2019-12-20
Keep empty array values on FieldArray

viniciusdacal:
This fixes the issue reported on #2130

### DIFF
--- a/packages/formik/src/FieldArray.tsx
+++ b/packages/formik/src/FieldArray.tsx
@@ -7,7 +7,7 @@ import {
   SharedRenderProps,
   FormikProps,
 } from './types';
-import { getIn, isEmptyChildren, isFunction, setIn } from './utils';
+import { getIn, isEmptyChildren, isFunction, setIn, isEmptyArray } from './utils';
 import isEqual from 'react-fast-compare';
 
 export type FieldArrayRenderProps = ArrayHelpers & {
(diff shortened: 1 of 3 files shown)


packages/formik/src/FieldArray.tsx
```diff
@@ -157,25 +157,38 @@ class FieldArrayInner<Values = {}> extends React.Component<
       let updateTouched =
         typeof alterTouched === 'function' ? alterTouched : fn;
 
+
+      let values = setIn(
```
viniciusdacal: I have to bring values up, to be executed before  `updateErrors` and `updateTouched`. Otherwise, it would cause an error with `unshift`.

packages/formik/test/FieldArray.test.tsx
```diff
@@ -348,11 +349,29 @@ describe('<FieldArray />', () => {
...
-
+    });
+    it('should remove a value at given index of field array', () => {
       arrayHelpers.remove(1);
       const expected = ['jared', 'brent'];
       expect(formikBag.values.friends).toEqual(expected);
     });
+
+    it('should be an empty array when removing all values', () => {
+      arrayHelpers.remove(0);
```
viniciusdacal: seems weird calling 0 multiple times, but every time we call remove, the indexes get updated.

packages/formik/src/FieldArray.tsx
jaredpalmer: Add this in a comment plz above these lines

packages/formik/test/FieldArray.test.tsx
jaredpalmer: Add this in a comment

jaredpalmer:
This seems fine. Add comments and we can merge

oleksiibilous:
@jaredpalmer Can you merge it please?

PR #2148, opened 2019-12-23
Add props.validateOnChange check to cDU

crossjs:
Should NOT run `validateForm` on cDU while `props.validateOnChange` in `FieldArray` is set to `false`

### DIFF
--- a/packages/formik/src/FieldArray.tsx
+++ b/packages/formik/src/FieldArray.tsx
@@ -132,11 +132,12 @@ class FieldArrayInner<Values = {}> extends React.Component<
     prevProps: FieldArrayConfig & { formik: FormikContextType<Values> }
   ) {
     if (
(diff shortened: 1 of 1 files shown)


packages/formik/src/FieldArray.tsx
```diff
@@ -132,11 +132,12 @@ class FieldArrayInner<Values = {}> extends React.Component<
     prevProps: FieldArrayConfig & { formik: FormikContextType<Values> }
   ) {
     if (
+      this.props.validateOnChange &&
```
jaredpalmer: Is this an actual prop of FieldArray? I’m on mobile and can’t recall.

anttiahola:
I would like to see this merged as well. This is a major performace issue in Formik. The whole `<FieldArray validateOnChange={false}>` is broken because it is not used in this comparison!

PR #2159, opened 2020-01-02
Correctly handle single checkboxes and grouped ones values

eliamaino-fp:
This PR changes how the value for a checkbox is retrieved and fixes the behavior described in https://github.com/jaredpalmer/formik/issues/2160

- If the value of the checkbox was a boolean, return a boolean (using the `checked` property of the target event)
- If the value of the checkbox was not a boolean, return a boolean if the `value` property of the target event "is or looks like a boolean", return an Array otherwise.
- When an Array is returned add the new value only if not already in the Array, and remove it only if already in the Array.

The changes to the checkbox selection were made here https://github.com/jaredpalmer/formik/pull/1555

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -1114,24 +1114,39 @@ function getValueForCheckbox(
   checked: boolean,
   valueProp: any
 ) {
-  // eslint-disable-next-line eqeqeq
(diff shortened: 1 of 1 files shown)


jaredpalmer:
lgmt

jaredpalmer:
bitmoji

PR #2176, opened 2020-01-07
Use curried onChange and onBlur functions in useField

### LINKED ISSUE #1674: [v2] useField: bound name to onChange and onBlur
### Current Behavior

I'm using `useField` on react-native and since `TextInput` has no name attribute, I need to bind name to handlers manually:
```ts
const TextField: React.SFC<TextFieldProps> = ({ name, ...props }) => {
  const [field] = useField(name);

  return (
    <TextInput
      {...props}
      value={field.value}
      onChangeText={field.onChange(name)}
      onBlur={field.onBlur(name)}
    />
  );
};
```

### Desired Behavior

I spect to use `onChange` and `onBlur` handlers like this:
```ts
const TextField: React.SFC<TextFieldProps> = ({ name, ...props }) => {
  const [field] = useField(name);

  return (
    <TextInput
      {...props}
      value={field.value}
      onChangeText={onChange}
      onBlur={onBlur}
    />
  );
};
```

### Suggested Solution

Bind `name` in `formik.getFieldProps`.
I think here:
https://github.com/jaredpalmer/formik/blob/377c36e3d1fecc0f0556d0218a18725ad04a7505/src/Formik.tsx#L765-L770
It should be replaced with:
```ts
      const field: FieldInputProps<any> = {
        name,
        value: valueState,
        onChange: handleChange(name),
        onBlur: handleBlur(name),
};
```

Can this break `useField` for web users?

### Possible workarounds

Create a `useFieldNative`:

```ts
interface FieldInputNativeProps {
  value: any;
  onChangeText: (text: string) => void;
  onBlur: (e: any) => void;
}

function useFieldNative(
  name: string,
): [FieldInputNativeProps, FieldMetaProps<any>] {
  const [field, meta] = useField(name);
  const fieldNative: FieldInputNativeProps = {
    value: field.value,
    onChangeText: field.onChange(name),
    onBlur: ()=> field.onBlur(name),
  };

  return [fieldNative, meta];
}

const TextField: React.SFC<TextFieldProps> = ({ name, ...props }) => {
  const [field] = useFieldNative(name);

  return <TextInput {...props} {...field} />;
};
```

react-native users

zingerj:
Fixes #1674 and closes #2082. Bind the passed field name to the `handleChange` and `onChange` functions to remove the need for explicit binding when using `useField` with React Native while maintaining the current behavior on the web.

### DIFF
--- a/docs/guides/react-native.md
+++ b/docs/guides/react-native.md
@@ -4,15 +4,9 @@ title: React Native
 custom_edit_url: https://github.com/jaredpalmer/formik/edit/master/docs/guides/react-native.md
 ---
 
-**Formik is 100% compatible with React Native and React Native Web.** However,
-because of differences between ReactDOM's and React Native's handling of forms
-and text input, there are some differences to be aware of. This section will walk
-you through them and what we consider to be best practices.
+**Formik is 100% compatible with React Native and React Native Web.** However, because of differences between ReactDOM's and React Native's handling of forms and text input, there are some differences to be aware of. This section will walk you through them and what we consider to be best practices.
 
-### The gist
-
-Before going any further, here's a super minimal gist of how to use Formik with
-React Native that demonstrates the key differences:
+### Basic usage
 
 ```jsx
 // Formik x React Native example
@@ -28,8 +22,8 @@ export const MyReactNativeForm = props => (
     {({ handleChange, handleBlur, handleSubmit, values }) => (
       <View>
         <TextInput
-          onChangeText={handleChange('email')}
-          onBlur={handleBlur('email')}
+          onChangeText={handleChange}
+          onBlur={handleBlur}
           value={values.email}
         />
         <Button onPress={handleSubmit} title="Submit" />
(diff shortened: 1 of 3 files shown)


zingerj:
@jaredpalmer I updated the React Native guide in the docs (which closes https://github.com/jaredpalmer/formik/pull/2082) but having some difficulty with the tests. Getting the following errors:
```
  ● Field / FastField › receives { field, form, meta } props and renders element › <Field />

    expect(received).toBe(expected) // Object.is equality

    Expected: [Function anonymous]
    Received: serializes to the same string

      128 |         expect(props.field.name).toBe('name');
      129 |         expect(props.field.value).toBe('jared');
    > 130 |         expect(props.field.onChange).toBe(handleChange('name'));
          |                                      ^
      131 |         expect(props.field.onBlur).toBe(handleBlur('name'));
      132 |         expect(props.form).toEqual(getFormProps());
      133 |         if (idx !== 2) {

      at test/Field.test.tsx:130:38
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (test/Field.test.tsx:127:16)
```
Any thoughts? Also tried using `toEqual`, but no dice.

kyle-johnson:
@zingerj `handleChange('name')` creates a new object (a function), so `.toBe` and `.toEqual` will fail.

It looks like the curry is always used (whether in browser DOM or RN)? If that's the case, the test may not make sense the way it originally was coded, since the curry may remove any need to identify names via the DOM.

zingerj:
Good call @kyle-johnson, thanks for pointing that out! @jaredpalmer I think this should be good to go now. Updated the tests to check that a function is being passed to the change and blur handlers. This should make React Native usage simpler for the time being!

jaredpalmer:
@johnrom should we merge this or our parse/format?

johnrom:
I think we need to keep the old handleChange for backwards compatibility either way. If this change was merged, it would break everyone's React Native apps right? And some complex web projects as well.

I think we need to move towards the new implementation we have on parse / format PR which also fixes this issue. (right?) by providing a curried handler. 

```jsx
// use the default field onX
<input
  onChange={field.onChange}
  onBlur={field.onBlur}
/>
// OR create a new handler via handleX
<input
  onChange={field.handleBlur(fieldName)}
  onBlur={field.handleBlur(fieldName)}
/>
```

We can then deprecate `field.handleX` in favor of `[field, meta, helpers] = useField(name); helpers.handleX` or something, since it's actually not specific to `field`. That way someone could bail out and use for example `helpers.handleChange(name)(field.parse(value))`.

(I don't have the code in front of me so it's possible I'm missing something)

jaredpalmer:
We can't merge this as is. This will be tackled in v3, potentially with a codemod

PR #2179, opened 2020-01-07
Fix FieldHelperProps typings

kristenmills:
Make FieldHelperProps generic like FieldInputProps and FieldMetaProps


jaredpalmer:
Thanks! You rock!

PR #2185, opened 2020-01-09
Add submitForm to FormikHelpers to allow form re-submit

### LINKED ISSUE #1989: Re-Add submitForm to onSubmit's formikHelpers
### v1 Behavior

In v1 I used this pattern to allow retry-behavior with user interaction:

```
onSubmit={async (values, formikActions: FormikActions<MyFormData>) => {
    props.onSendMessage(values, formikActions.submitForm)
    formikActions.setSubmitting(false)
}}
```

The code in `onSendMessage` would store the `submitForm` function, and in case of an API error, show the error to the user and allow him to retry the submit. The clou was that the user can still tweak the form inputs while seeing the error message, because we want to avoid blocking user input as much as possible. Therefore when the user clickes the "retry" button, we really want Formik's `submitForm`/`onSubmit` to be triggered again and send the current values of the form.

### Current Behavior

In v2 this is no longer possible because `submitForm` was removed from the `FormikActions` (which have been renamed to `FormikHelpers`).

### Desired Behavior

I'd like to reinstate the behavior of Formik v1 and have `submitForm` as part of the `onSubmit` parameters in any way.

### Suggested Solution

Either add `submitForm` to the `FormikHelpers`, or as third argument to `onSubmit`, or introduce a new type that extends `FormikHelpers` with `submitForm` and use that for the second argument to `onSubmit`.

This is for people who want to use clean retry behavior that is triggerable by a user.

An alternative would be to have some helper prop that allows the parent component of the form that contains Formik to trigger the submit action from the outside. This is rather unwieldy and not very nice encapsulation-wise.

fabb:
Resolved #1989

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -12,6 +12,7 @@ import {
   FieldMetaProps,
   FieldHelperProps,
   FieldInputProps,
+  FormikHelpers,
 } from './types';
 import {
   isFunction,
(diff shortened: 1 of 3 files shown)


fabb:
Seems like there is a general issue with the codesandbox build, unrelated to this PR:

```
error formik-project@: The engine "node" is incompatible with this module. Expected version "12.x". Got "10.16.3"
```

Issue #1804, opened 2019-09-08
`onChange` reference can cause performance issues


3nvi:
## 🚀 Feature request

Make `onChange` have the same reference between renders

### Current Behavior

Currently, the `onChange` that gets "generated" and passed to a component via `<Field>` gets a difference reference on every render. Due to that, the component rendered from `Field` gets a prop with different reference on every render, which causes the component to be re-rendered regardless of the use of `memo` in it.

For example, let's suppose we have a Formik form which contains the following:

`<Field as={Combobox} name="combobox" />`

Any time **any field** gets rendered (not necessarily this one), the `onChange` prop that gets **automatically** passed to the `Combobox` gets a different reference. Thus, even if the `Combobox` implements a `React.memo()` it will have to re-render since its props changed.

On big forms or forms with a lot of styled components within them (a.k.a. loads of context consumers) this causes performance issues.

### Desired Behavior

Ideally, we would want the `onChange` prop passed to components whose values **have not changed** to be referentially the same as the one in the previous render cycle. This way, the component which a `memo` won't render unnecessarily.

### Suggested Solution

This issue doesn't happen with `onBlur` (which always keeps the same reference), but only with `onChange`. Seeing the code I see that it gets a new reference every time the `executeChange` function changes reference https://github.com/jaredpalmer/formik/blob/44afb37ab6af73011ef4a0b10577d56cc5a01be9/src/Formik.tsx#L545. 

The latter changes reference every time **any of the values** changes https://github.com/jaredpalmer/formik/blob/44afb37ab6af73011ef4a0b10577d56cc5a01be9/src/Formik.tsx#L483.
This is reference change happens only to cover the `checkboxes` case https://github.com/jaredpalmer/formik/blob/44afb37ab6af73011ef4a0b10577d56cc5a01be9/src/Formik.tsx#L521, so the `state.values` has been added to the `useCallback` array for something specific that doesn't necessarily apply to all forms out there.

Ideally we would want to handle this case differently. Perhaps separate the `executeChange` for the checkboxes case and keep the existing one for everything else? 

To be fully honest, I might be missing something big here, so any info will be extremely helpful. At the end of the day, my goal is to have the `Combobox` (in the above example) not re-render when a value unrelated to it changes.

### Who does this impact? Who is this for?
All users, but mainly projects with complex and/or heavy forms

P.S. Cheers for everything. Really appreciate the time you throw in this OSS.

johnrom:
[...]

PR #2187, opened 2020-01-09
Fix use event callback in handlers

deini:
Should help to keep the same reference for `onChange` and not re-render every time.

Closes https://github.com/jaredpalmer/formik/issues/1804


jaredpalmer:
Thanks.

PR #2200, opened 2020-01-13
Add validateOnSubmit

jaredpalmer:
- Add flag `validateOnSubmit?: boolean = true` to `<Formik>`/`useFormik()`/`withFormik()`. When set to `false`, Formik will not run validation or check for errors before running `onSubmit` prop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "devDependencies": {
     "lerna": "^3.19.0",
     "release": "6.0.1",
-    "got": "10.0.4"
+    "got": "10.0.4",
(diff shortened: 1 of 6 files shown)


jaredpalmer:
@CompuIves how do i fix codesandbox CI? Seems to have broken when I set node engine in package.json (This change was required by latest now.sh update because of upstream AWS Lambda Node 8 EOL).

kc980602:
support adding validateOnSubmit, this is very useful when we want to trigger validation manually during onSubmit

Issue #2034, opened 2019-11-14
resetForm handler doesn't work same at formik 2 version


krailler:
## 🐛 Bug report

With formik 1, when we use resetForm handler and we don't pass nextValues, it uses initialValues from props. You can see this behavior in https://github.com/jaredpalmer/formik/blob/version-1.5.8/src/Formik.tsx#L532

But, with formik 2, when we use resetForm handler and we don't pass nextValues, it uses initialValues from a ref and this ref is only updated when the initialValues changes and the variable enableReinitialize is true. You can see this behavior in https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L362

### Current Behavior

resetForm handler uses initialValues from ref using formik 2.

### Expected behavior

Use initialValues from props when we use resetForm handler, same as formik 1.

### Suggested solution(s)

I don't know if this is an issue or the correct behavior. I need to know if the formik 1 resetForm behavior is the correct (And fix it in formik 2) or the actual behavior is correct and it's a bug with formik 1.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.4
| React            | 16.11.0
| TypeScript       | 3.7.2
| Browser          | Google Chrome 78.0.3904.97 (64-bit)
| npm/Yarn         | npm 6.12.0
| Operating System | macOS Catalina 10.15

AlberErre:
up!

krailler:
Hi @jaredpalmer, can you see this, please? 🙏

akonyer:
I'm trying to upgrade to formik v2.0 and I'm having the same error. I think until this is sorted out I have to revert back to v1 for now

jaredpalmer:
Can you submit a PR?

sw-yx:
should close since PR merged

PR #2205, opened 2020-01-14
Adds the same behavior in the resetForm as a formik v1.

krailler:
See issue https://github.com/jaredpalmer/formik/issues/2034

In formik v1, resetForm use the initialValues from props, you can see here https://github.com/jaredpalmer/formik/blob/version-1.5.8/src/Formik.tsx#L532

In this case, I cannot use directly prop.initialValues, because the resetForm is a memorize callback and in some cases will not work properly (even adding as deps).

If you have another idea to get back the v1 behavior, feel free to comment, please 🤗

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -419,6 +419,12 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


jaredpalmer:
@krailler thanks! taking a peek now

krailler:
Hi @jaredpalmer any news? :)

Issue #2208, opened 2020-01-14
Use React.forwardRef in <Formik>


jaredpalmer:
As of 2.1.2, `<Formik>` now allows for an `innerRef` prop. However, we should instead use `React.forwardRef` + `React.useImperativeHandle`. This is more intuitive and closer to best practices. The problem, though, is that `<Formik>` takes TS generics for `Values` and I'm not sure how to express this in TS. AFAICT, you can't because forwardRef returns a constant and constants can't have generics.

drivasperez:
How about:

```tsx
export const Formik = React.forwardRef(function Formik<
  Values extends FormikValues = FormikValues,
  ExtraProps = {}
>(props: FormikConfig<Values> & ExtraProps, ref: any) {
  const formikbag = useFormik<Values>(props);
  const { component, children, render, innerRef } = props;

  // This allows folks to pass a ref to <Formik />
  React.useImperativeHandle(innerRef, () => formikbag);
  React.useImperativeHandle(ref, () => formikbag);
  ...
```

jaredpalmer:
Submit a PR?

drivasperez:
Done!

drivasperez:
Ah, never mind, I see the issue.

johnrom:
Doing a little cheating makes this possible. I didn't test too thoroughly, and honestly I'm not sure FormikContextType is the right type for FormikBag, but if not we need to set it to the correct type.

https://github.com/jaredpalmer/formik/pull/2222

drivasperez:
I couldn’t find anything that worked before that doesn’t now. I guess the main concern is: going forward, can you break it going forward in a way that the “as [...component...]” cast won’t complain about?

johnrom:
Basically, a `Formik` contributor could change the types incorrectly. I wish we could ignore a specific rule so we wouldn't have to use `as` and we could protect against all the other errors, but for now `as` is the best protection we can get. `as` won't protect against most things, but it'll at least make sure it's kind of the same signature. I made a similar comment here: https://github.com/microsoft/TypeScript/pull/30215

ProteanCode:
It touches https://github.com/jaredpalmer/formik/issues/2290

tony:
A related StackOverflow I thought would be helpful to consider: https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref ([answer with some approaches](https://stackoverflow.com/a/58473012))

[...]

PR #2209, opened 2020-01-14
Add forwardref to Formik

drivasperez:
Tentatively solves #2208, maybe.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -994,15 +994,16 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   return ctx;
 }
 
-export function Formik<
+export const Formik = React.forwardRef(function Formik<
   Values extends FormikValues = FormikValues,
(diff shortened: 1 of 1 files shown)


jaredpalmer:
This doesn't work AFAICT because of the generics.

https://github.com/jaredpalmer/formik/pull/2209/checks?check_run_id=389788052#step:5:128

drivasperez:
Yeah, I realised as soon as I made the PR. I should have run the tests locally before pushing, sorry.

Trying some really hair-raising things with casting before work, but have run out of time for the moment. Starting to think innerRef is just the better solution.

drivasperez:
@jaredpalmer This now builds and passes the existing tests, and if you pass a ref, it gets the formikbag. Is the codesandbox environment broken, or have I done something wrong?

jaredpalmer:
Codesandbox is broken.

drivasperez:
Ah, okay. Do you think something like this is worth the typescript weirdness?

PR #2215, opened 2020-01-15
Updated setError signature to string

sjsingh85:
As per comments on [here](https://github.com/jaredpalmer/formik/commit/e696e57c97a297b332be66515cbddfc8f4176321#r36783717), updated the signature of setError to take string instead of Value as param.

### DIFF
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -289,7 +289,7 @@ export interface FieldHelperProps<Value> {
   /** Set the field's touched value */
   setTouched(value: boolean): void;
   /** Set the field's error value */
-  setError(value: Value): void;
+  setError(value: string): void;
 }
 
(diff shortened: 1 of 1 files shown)


sjsingh85:
Hi Jared, 

This is a small change for typings, Can you confirm what should I do to fix this codesandbox issue?

Thanks

JamieS1211:
Had the same issue and was going to do same fix but found this. Looks perfect to me - one line change, short and sweet

Kaldy14:
Any news on this PR? Would be great if it could get fixed

JamieS1211:
Labelled stale, perhaps we should re-do this?

Kaldy14:
If you know how that would be awesome :) I dont have any experience opening / fixing PR in open source so I would probably be lost...

PR #2222, opened 2020-01-17
Try using forwardRef

johnrom:


jaredpalmer:
Does this work with ts?

johnrom:
yes, because but only because there is a cast. I left a comment on #2208 describing the pros and cons

jaredpalmer:
I'm pretty scared of this since this cast effectively removes type safety for development.

johnrom:
The type safety that is removed is specifically the safety between the two type definitions. If those type definitions go out of sync somehow, type safety will not be guaranteed. As long as those type definitions are in sync, type safety is guaranteed. It's definitely a trade-off, but is the only way to implement forwardRef that I can find. I'd recommend raising an issue on either typescript or react libraries. Coming from the owner of a large library, it might have some weight, and maybe one of the gurus would have a better solution. There's a lot of discussion [here](https://github.com/microsoft/TypeScript/pull/30215), but I don't see an issue opened specifically for the `forwardRef` case. The second best case scenario is [ignoring only specific typescript rules](https://github.com/Microsoft/TypeScript/issues/19139).

Alternatively, we could add a comment like `// DO NOT CHANGE OR THE UNIVERSE EXPLODES` as a reminder that those two type definitions must remain in sync.

mlzsolti:
Guys, if this is working, could you please write an example how to use it, as the docs don't seem clear about this? Thank you.

johnrom:
@mlzsolti this PR is experimental. Since it uses type casting, it could introduce problems with future maintenance due to types becoming out of sync. It's a gamble I'm not sure Jared is willing to take.

Issue #2219, opened 2020-01-16
Add shouldValidate param to useField's helper setTouched method


unframework:
### Current Behavior

The helper object returned by `useField` contains a `setTouched(isTouched)` method. That method accepts only the intended `isTouched` boolean value, and does not allow skipping validation while setting it.

### Desired Behavior

There is a `shouldValidate` parameter on form-wide `setFieldTouched(name, isTouched, shouldValidate)` method that allows skipping validation when updating this flag. Would be great to have the equivalent `shouldValidate` parameter on the field-specific helper `setTouched` method.

### Suggested Solution

Seems like this would be a straightforward pass-through to add in the helper function here: https://github.com/jaredpalmer/formik/blob/d71efdec147e94b66eba86ec17c715d0d4ef3d29/packages/formik/src/Formik.tsx#L893. I.e. could add the `shouldValidate` parameter to the arrow function and then passing it to the `setFieldTouched` method call.

### Who does this impact? Who is this for?

This would be handy for folks who are coding complex form inputs and are converting to use the `useField` hook.

### Describe alternatives you've considered

For now, the workaround is to get `form` object via `useFormikContext` and use the `setFieldTouched` method on that.

### Additional context

No extra notes to add.

Thanks for all the amazing work on Formik!

klimashkin:
I think the same parameter should be added to `helpers.setValues` for consistency

PR #2223, opened 2020-01-17
Add support for shouldValidate in FieldHelperProps of useField

mrmuhammadali:
I have a use case where I've disabled `validateOnChange` but wanted to run the validations using only `setValue` in case of a Checkbox as it only calls `onChange`. So I've added the support for `shouldValidate` in `FieldHelperProps`, i.e. `setValue` and `setTouched`.

Fixed Issue #2219 in this PR.

-----
[View rendered docs/api/useField.md](https://github.com/mrmuhammadali/formik/blob/master/docs/api/useField.md)

### DIFF
--- a/docs/api/useField.md
+++ b/docs/api/useField.md
@@ -150,6 +150,6 @@ An object that contains relevant computed metadata about a field. More specifica
 
 An object that contains helper functions which you can use to imperatively change the value, error value or touched status for the field in question. This is useful for components which need to change a field's status directly, without triggering change or blur events.
 
-- `setValue(value: any): void` - A function to change the field's value
-- `setTouched(value: boolean): void` - A function to change the field's touched status
+- `setValue(value: any, shouldValidate?: boolean): void` - A function to change the field's value. Calling this will trigger validation to run if `validateOnChange` is set to `true` (which it is by default). You can also explicitly prevent/skip validation by passing a third argument as `false`
+- `setTouched(value: boolean, shouldValidate?: boolean): void` - A function to change the field's touched status. Calling this method will trigger validation to run if `validateOnBlur` is set to `true` (which it is by default). You can also explicitly prevent/skip validation by passing a third argument as `false`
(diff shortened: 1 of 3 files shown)


mrmuhammadali:
@jaredpalmer / @johnrom 
Can you guys look into this PR.

johnrom:
@mrmuhammadali I appreciate your contribution, but pinging is not a good way to prioritize PRs.

johnrom:
The code looks good to me and can be approved for implementing this specific technique, but I can't help feel this isn't the right place to decide whether to validate or not. I'd prefer we exposed a method that never validated so that someone could decide on their own to `justSetFieldValue().then(shouldValidate ? validateField : Promise.resolve)`.

mrmuhammadali:
Sorry about that, I’m new to open source contribution but I’ll note that down.
Anyway, I think if setFieldValue and setFieldTouched have a third parameter of shouldValidate then setValue and setTouched should support that too, as internally these functions are using above mentioned methods.

mrmuhammadali:
In case of Dropdowns, Radio Buttons, Toggles and Checkboxes, we’ve to take the decision whether to validate the field or not, so that’s another argument to discuss. As in above mentioned cases, we’ve to call setTouched and setValue at the same time, i.e onChange.

johnrom:
No the contribution is great, I'm just wondering if Jared had purposefully left out the `shouldValidate` so that the decision to validate can be somewhere more appropriate, and if so, where that might be.

unframework:
I am the original reporter for #2219, thanks for making this PR @mrmuhammadali!

I can see how there might be a broader reason to not include shouldValidate flag, but this would be quite useful in some complex field use cases (came up twice for me now in the last couple months). As discussed above, for checkbox groups and toggles the conventional validate-on-blur/validate-on-change flags don't really apply anyway, so this flag feels like a fairly uncontroversial feature. Hope to see a decision on this soon, even though I understand that the team is busy.

Thanks folks!

PR #2237, opened 2020-01-23
+error on array helper

wallymathieu:


jaredpalmer:
I don't think this is necessary, folks can just grab the errors themselves.

wallymathieu:
Oki 😄

PR #2243, opened 2020-01-24
Fix validateOnMount useEffect issue

whazap:
in our case using validateOnMount would cause the form to always be invalid:

```javascript
useFormik({
  initialValues: { email: "" },
  validateOnMount: true,
  validationSchema: Yup.object().shape({
    email: Yup.string()
      .email()
      .required("Required")
  }),
}
```

tracked it down to this useEffect:

```javascript
React.useEffect(() => {
    if (validateOnMount && isMounted.current === true) {
      validateFormWithLowPriority(props.initialValues);
    }
  }, [props.initialValues, validateOnMount, validateFormWithLowPriority]);
```

which triggers props.initialValues and thus the useEffect hook for every render, because in our case initialValues is always a new object.

the quickfix for us was to store initialValues in useRef

```javascript
const initialValues = useRef({
  phone: '',
});
useFormik({
  initialValues: initialValues.current
  validateOnMount: true,
  validationSchema: Yup.object().shape({
    email: Yup.string()
      .email()
      .required("Required")
  }),
})
```

see this codesandbox:
Bug.js and QuickFix.js

https://codesandbox.io/s/formik-validateonmount-with-initialvalues-issue-jvkex?fontsize=14&hidenavigation=1&theme=dark

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -365,9 +365,9 @@ export function useFormik<Values extends FormikValues = FormikValues>({
 
   React.useEffect(() => {
     if (validateOnMount && isMounted.current === true) {
-      validateFormWithLowPriority(props.initialValues);
+      validateFormWithLowPriority(initialValues.current);
     }
-  }, [props.initialValues, validateOnMount, validateFormWithLowPriority]);
+  }, [validateOnMount, validateFormWithLowPriority]);
 
   const resetForm = React.useCallback(
     (nextState?: Partial<FormikState<Values>>) => {


jaredpalmer:
Fixed in 2.1.4

sgnl:
Thanks the fix, this was driving me up the wall!

TheCodeDestroyer:
@jaredpalmer shouldn't validation re-trigger if `initialValues.current` values change?

omatei-brightplan:
This issue came back again after 2.1.6

johnrom:
@omatei-brightplan please open a new issue referencing this PR reproducing the issue in a codesandbox with 2.1.6.

PR #2265, opened 2020-02-03
Properly handle nested array data in `prepareDataForValidation`.

### LINKED ISSUE #2213: Nesting arrays shows "invalid value" for first level
**"formik": "^2.0.6",**

Hello, I was trying to make a validation for my data structure but I'm having issues.

It goes like this: I have a matrix of objects from formik:

```jsx
[
  [
    { a: '123', b: 456 },
    { a: '789', b: 999 },
  ], [
    { a: 'abc', b: 0 },
    { a: 'def', b: 1 },
  ]
]
```

so my validation schema is:

```jsx
const schema = object().shape({
  myField: array().of(
    array().of(
      object().shape({
        a: string().required(),
        b: number().required(),
      }).required(),
    ).required(),
  ).required(),
});
```

But when I pass validation, on:

```jsx
{
  myField: [
    [
      { a: '123', b: 456 }
    ]
  ]
}
```

all I get is:

```jsx
{
  errors: {
    myField: "myField[0] must be a `array` type, but the final value was: `null` (cast from the value `{
      "0": {
        "a": "\"123\"",
        "b": 456
      }
    }`).
    If "null" is intended as an empty value be sure to mark the schema as `.nullable()`"
  }
}

```

Even when I pass in the whole structure as initialValue for formik.

Removing a sinle level from the nesting, makes it work again...

```jsx
const schema = object().shape({
  myField: array().of(
    object().shape({
      a: string().required(),
      b: number().required(),
    }).required(),
  ).required(),
});
```

jamesremuscat:
Previously, we would always return an object even if `values` was an array. Now, we test to see if `values` is an array and set `data` accordingly.

I've included a test for this scenario.

Fixes #2213.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -1095,7 +1095,7 @@ export function validateYupSchema<T extends FormikValues>(
 export function prepareDataForValidation<T extends FormikValues>(
   values: T
 ): FormikValues {
-  let data: FormikValues = {};
+  let data: FormikValues = Array.isArray(values) ? [] : {};
   for (let k in values) {
     if (Object.prototype.hasOwnProperty.call(values, k)) {
       const key = String(k);
--- a/packages/formik/test/Formik.test.tsx
+++ b/packages/formik/test/Formik.test.tsx
@@ -969,6 +969,18 @@ describe('<Formik>', () => {
         date,
       });
     });
+
+    it('should work correctly for nested arrays', () => {
+      const expected = {
+        content: [
+          ['a1', 'a2'],
+          ['b1', 'b2'],
+        ]
+      }
+
+      const dataForValidation = prepareDataForValidation(expected);
+      expect(dataForValidation).toEqual(expected);
+    });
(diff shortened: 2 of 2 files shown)


elmarburke:
This fix didn't went in release of 2.1.4, right? I still have the issue 😢

jamesremuscat:
@elmarburke No, it was merged just after 2.1.4 was released.

kononenko-a-m:
hi @jaredpalmer wanted to ask, if there any plan to release this fix in 2.1.5, and is there any plan in general release 2.1.5?

TMiguelT:
This is a pretty dreadful problem, since it makes nested arrays totally unusable. However I found a workaround, which is to install the CI version of Formik from just after the PR was merged: `npm install https://pkg.csb.dev/jaredpalmer/formik/commit/157603ab/formik`. You can also use newer builds listed [on this page](https://ci.codesandbox.io/status/jaredpalmer/formik), but the command above seems to work okay for me.

PR #2280, opened 2020-02-08
Parse Format PR Review Changes

johnrom:
Fixes items in review for #2255 

I didn't actually get a chance to test it on a real project yet. @jaredpalmer do you use npm link?

I think last time I was having issues with "two reacts" when using npm link. Maybe that's been fixed since then.

### DIFF
--- a/packages/formik/src/Field.tsx
+++ b/packages/formik/src/Field.tsx
@@ -208,7 +208,7 @@ export function Field({
   if (component) {
     // This behavior is backwards compat with earlier Formik 0.9 to 1.x
     if (typeof component === 'string') {
-      const { innerRef, ...rest } = props;
+      const { innerRef, parse, format, ...rest } = props;
       return React.createElement(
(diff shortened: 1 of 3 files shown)


johnrom:
Added a small change removing parse and format props from the underlying <input /> when not passed a complex component. I honestly don't think we should be passing them to complex components either, but didn't have time to think it through so I left it for now.

I was having the double-react issue again. I switched to Yarn, which seemed to help once every few builds, but then I'd get the same error again. Basically, after running yarn link (`/packages/formik`), yarn link formik (`../formik-dev`), yarn build (`/`) and yarn start (`../formik-dev`), I had to delete `/node_modules` because the `link`ed formik would traverse its real path and find a second react there.  Is there a better workflow for this?

jaredpalmer:
@johnrom re: lerna, i can commit a tsdx playground that uses parcel to alias the react install

PR #2286, opened 2020-02-13
Allow deleting error from field in types

cliedeman:


dantman:
Looks like this isn't needed anymore, the current setFieldError uses `string | undefined`.

PR #2295, opened 2020-02-19
Fix bad "scheduler" import

FredKSchott:
Scheduler is a Common.js package, with no ESM "module" entrypoint. Pulling named imports off of a CJS package aren't supported in many environments, including Rollup, Snowpack & Node.js (v13 ESM support): https://nodejs.org/api/esm.html#esm_import_statements

This change updates this import statement to use the default import, which is something that all bundlers / environments currently support.


FredKSchott:
Ah, it looks like this problem gets resolved in your ESM -> CJS package build. I'd thought formik was shipping an ESM build that more closely matched the source. Nevermind then!

PR #2299, opened 2020-02-20
Fix bad "scheduler" import

FredKSchott:
Scheduler has no export of LowPriority. It's entire interface is prefaced with "unsafe_*", to communicate its experimental nature: https://unpkg.com/browse/scheduler@0.18.0/cjs/scheduler.development.js

Scheduler is a Common.js package, with no ESM "module" entrypoint. Pulling named imports off of a CJS package aren't supported in many environments, including Rollup, Snowpack & Node.js (v13 ESM support): https://nodejs.org/api/esm.html#esm_import_statements

This change updates this import statement to use the default import, which is something that all bundlers / environments currently support.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -27,7 +27,7 @@ import {
 } from './utils';
 import { FormikProvider } from './FormikContext';
 import invariant from 'tiny-warning';
-import { LowPriority, unstable_runWithPriority } from 'scheduler';
+import scheduler from 'scheduler';
 
 type FormikMessage<Values> =
   | { type: 'SUBMIT_ATTEMPT' }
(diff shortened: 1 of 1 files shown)


FredKSchott:
ESM is now good, but CJS tests are now failing :/ I don't know enough about your build process to debug, could it be something is tsdx?

danilotorrisi:
I'm also interested in this fix, how can we fix the CJS build?

admmasters:
Interested in this fix as well - currently blocking poc of snowpack

ColinKrist:
Am also interested in this for my POC of snowpack

jakewies:
+1

FredKSchott:
For anyone coming here from Snowpack, this should no longer fail on our end as of v2.4.0 by updating to the latest version of Rollup. 

This import still seems broken, fwiw

pvormittag:
FYI, PR #2438 addresses the failed `LowerPriority` import and was recently merged. Now awaiting an NPM release.

PR #2305, opened 2020-02-24
fix: make execute blur safe when event is not provided

karszawa:
Thanks for publishing such a great library 😊

## What

- Make `handleBlur` generate a function which does not need any argument if a field name is provided as an argument.
- Make `executeBlur` safe when an event (the first argument) is not provided

## Why

In `executeBlur` function, `e` (the first argument of this function) is not used when `path` (the second argument of this function) is provided. Nevertheless, this function causes a runtime error when the user executes `handleBlur` without any argument even though it does not need it!

## Backgorund

Some of the UI components provided form UI pattern library does not send a native event to a handler. Therefore, we need to specify the name of the field manually. In this case, this function no longer needs the event object.

In my case, [InputItem](https://rn.mobile.ant.design/components/input-item/) from Ant Design React Native does not pass an event object to the handler. So, I need to pass some dummy value to `handleBlur` like:

```tsx
<InputItem
  ...
  // onBlur cause same result in this case
  onEndEditing={() => handleBlur('username')({ target: { id: "", name: "", outerHTML: "" } })}
/>
```

This is annoying.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -694,7 +694,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       if (e.persist) {
         e.persist();
       }
-      const { name, id, outerHTML } = e.target;
+      const { name, id, outerHTML } = e?.target ?? {};
       const field = path ? path : name ? name : id;
 
       if (!field && __DEV__) {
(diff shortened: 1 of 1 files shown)


TwoAbove:
I also encountered this issue when I hit Enter in the input field (and not the submit button)
image

PR #2315, opened 2020-02-26
Update useFormik.md

jackall3n:
-----
[View rendered docs/api/useFormik.md](https://github.com/jackall3n/formik/blob/patch-1/docs/api/useFormik.md)


jackall3n:
Remove the example description as it was obviously accidentally copied from `useFormikContext()`. I would have replaced it with something else, but the form is really simple and other pages with simple forms don't have a description.

PR #2317, opened 2020-02-26
Track currently focused field by name

jaredpalmer:
- Add `focus`, `handleFocus`, `setFocus(field: string)` to formik context/bag
- `useField`/`Field` now returns `onFocus` inside of field
- Calling `handleBlur`/`setFieldTouched` will remove `focus`
- All focus'ing (via refs) should be handled outside of formik with a soon to be published hook. Generalizing focus is quite difficult, but will document a useFocusableField() hook and perhaps publish next week. The nasty thing is that you actually don't want to call .focus all the time. For example, on desktop you might want to focus the input, but sometimes you want to .select(). ON mobile, you often don't want to focus the input (because that will pull up the keyboard, but instead just the area around the input. For these reasons and more, I'm omitting this functionality at the moment.


packages/formik/src/Formik.tsx
```diff
@@ -81,6 +84,7 @@ function formikReducer<Values>(
     case 'SET_FIELD_TOUCHED':
       return {
         ...state,
+        focus: undefined,
```
johnrom: I wouldn't expect `setFieldTouched` to remove focus in all scenarios. For example, if a user is entering into an input and I programatically set the field to touched so the error displays.

packages/formik/src/Formik.tsx
```diff
@@ -904,7 +944,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   const getFieldProps = React.useCallback(
-    (nameOrOptions): FieldInputProps<any> => {
+    (nameOrOptions: any): FieldInputProps<any> => {
```
johnrom: Should this be `string | FieldHookConfig<any>` as used in `useField`?

packages/formik/src/Formik.tsx
jaredpalmer: Fair. Thus the onBlur/handleBlur will set focus then.

packages/formik/src/Formik.tsx
jaredpalmer: yes

mosnicholas:
Hey @jaredpalmer any update on this PR? Running into issues with refs and field focusing, this looks like it would resolve it perfectly! :)

l0gicgate:
Any updates on this PR?

jaredpalmer:
Will be able to hit this next week

smith-it:
Great timing - this is exactly what I need as well. I'm having problems with `useRef` and setting the focus following a form reset.

3runoDesign:
Any updates on this PR?

SiSa68:
Any update?!
Is there any workaround to setFocus on input?

hennessyevan (approved):

hennessyevan:
This feature would be really nice! Has anyone found a way to integrate this without forking the whole thing?

l0gicgate:
@hennessyevan here is an example of how you could wrap the Formik form component and create your own hooks to provide which element is focused:

```jsx
import React, { createContext, useCallback, useContext, useState } from 'react';
import PropTypes from 'prop-types';
import { Form as FormikForm, Formik } from 'formik';

export const FormContext = createContext(null);

export const useFormContext = () => {
  const context = useContext(FormContext);
  if (context === undefined) {
    throw new Error('FormContext must be used within a Form');
  }
  return context;
};

export const Form = ({
  children,
  ...props,
}) => {
  const [focus, setFocus] = useState(null);

  const handleBlur = useCallback(
    e => {
      const field = e?.target?.name;
      if (field === focus) {
        setFocus(null);
      }
    },
    [focus, setFocus]
  );

  const handleFocus = useCallback(
    e => {
      const field = e?.target?.name;
      if (field === focus) {
        setFocus(field);
      }
    },
    [focus, setFocus]
  );

  return (
    <Formik
      {...props}
    >
      {form => {
        const value = {
          ...form,
          focus,
          handleFocus,
          handleBlur: e => {
            form.handleBlur(e);
            handleBlur(e);
          },
        };
        return (
          <FormikForm {...formProps}>
            <FormContext.Provider value={value}>
              {typeof children === 'function' ? children(value) : children}
              <input
                style={{ position: 'absolute', top: -9999 }}
                tabIndex={-1}
                type="submit"
                value="submit"
              />
            </FormContext.Provider>
          </FormikForm>
        );
      }}
    </Formik>
  );
};

Form.propTypes = {
  children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
};
```

Issue #2292, opened 2020-02-18
initialValues being reset causing dirty to be incorrect (v2.1.4 regression)


jamesmosier:
### Current Behavior

As of v2.1.4, `initialValues` cannot be counted on to not change during the duration of a page load. [These lines](https://github.com/jaredpalmer/formik/compare/jaredpalmer:2852cdb...jaredpalmer:747fd89#diff-13cb4898b838c74f25fbdd7bb44a139fR439-R443) introduced in `v2.1.4` will now update `initialValues` **anytime** the `props.initialValues` changes. Therefore if I have an application and the `initialValues` I pass in happen to change then Formik will update that in its internal state and therefore when doing things like determining if a form is dirty it can report unexpected results.

### Expected behavior

I would expect once I pass in `initialValues` for those to be set throughout the life of the page (as this is how it has worked previously).

### Reproducible example

https://codesandbox.io/s/formik-v214-dirty-breaking-2y6ho

1. Input a single letter into the `name` field (`dirty = true`)
2. Input a single letter into the `email` field (`dirty = true`)
3. Then remove (backspace) the value you just entered in the `email` field.
4. Notice that in Formik state `dirty = false` because `initialValues` are equal to `values`. 

If you switch the Formik version back to `2.1.3` the above test case end with `dirty = true` in step 4.

### Suggested solution(s)

Remove [this code](https://github.com/jaredpalmer/formik/blob/c85d85a3b6236a30034e73f8b821a280de8856b0/packages/formik/src/Formik.tsx#L439-L443) which was introduced in the `2.1.4` release. It was introduced via https://github.com/jaredpalmer/formik/pull/2205

### Additional context

If this new functionality is to be expected and desired then I believe this should be a new minor version bump since it is a breaking change.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.1.4
| React            | 16.11.0
| TypeScript       | n/a
| Browser          | Chrome 80
| npm/Yarn         | Yarn
| Operating System | Mac OS

PR #2318, opened 2020-02-26
Fix 2.1.4 regression with initialValues updating

jamesmosier:
This PR remove code introduced in #2205 as described in this issue #2292 

To reiterate, as of v2.1.4, initialValues cannot be counted on to not change during the duration of a page load. [This code](https://github.com/jaredpalmer/formik/blob/c85d85a3b6236a30034e73f8b821a280de8856b0/packages/formik/src/Formik.tsx#L439-L443) which was introduced in v2.1.4, updates initialValues anytime the `initialValues` passed into `<Formik>` changes. Therefore if I have an application and the initialValues I pass in happen to change then Formik will update that in its internal state and therefore when doing things like determining if a form is dirty it can report unexpected results.

This by proxy would then force https://github.com/jaredpalmer/formik/issues/2034 to be investigated/fixed again.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -437,12 +437,6 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   React.useEffect(() => {
-    if (!enableReinitialize) {
-      initialValues.current = props.initialValues;
-    }
-  }, [enableReinitialize, props.initialValues]);
-
-  React.useEffect(() => {
     if (
       enableReinitialize &&
       isMounted.current === true &&


sPaCeMoNk3yIam:
Need this fix, right now (in 2.1.4) `resetForm` is not working as expected and forces me to stay on 2.1.3.

jaredpalmer:
@jamesmosier @sPaCeMoNk3yIam agree. good looks and thanks for this

cspicuzza:
Thanks for tackling this issue.  How soon will it be released?

lnmp4000:
I've just spent ages debugging this issue with resetForm in my application. I was just about to raise the issue, so glad this is fixed.

Issue #2290, opened 2020-02-14
What is the current way of defining the ref for an innerRef attribute using TypeScript?


ProteanCode:
## ❓Question

Currently the innerRef has signature of `(instance: any) => void` which does not fit any type I can remind of, like `FormikProps` or `FormikHelpers`

Before that when I was using the 1.5 version where I could call methods like `setFieldTouched` on the ref therefore an inner Ref should be of type `FormikHelper` or at least inherit it. The `(instance: any) => void` looks broken

ProteanCode:
Since TypeScript is not really helpful i used `as` casting.

`<Formik innerRef={instance=>{formikRef=instance}}` {...otherProps}/>`

This however makes a non-constant `formikRef` variable

`let formikRef = useRef(null) as any;`

I do not think this is correct way with 2.x

johnrom:
We do currently use any. You are free to open a PR to change the innerRef to the signature that I wrote for useRef here: #2222

ProteanCode:
`TypeError: formikRef.setFieldTouched is not a function`

johnrom:
@ProteanCode are you using `innerRef`? My PR using `ref` itself is a draft, but you can take the same types I used to implement `useRef` to open a PR for innerRef, and it might be accepted, which should solve your issue with `any`.

TeoTN:
This is how I define it:

```
export type FormikRef<FormValues> = Ref<FormikProps<FormValues> | undefined>;

import React, { MutableRefObject, useCallback, useImperativeHandle, useRef } from 'react';
import { Formik, FormikConfig, FormikProps } from 'formik';
import { FormikValues } from 'formik/dist/types';
import { FormikRef } from '../models/types';

type FormikWithRefProps<ExtraProps, Values extends FormikValues> = ExtraProps & FormikConfig<Values> & {
  ref?: FormikRef<Values>,
}

export function FormikWithRef<Values extends FormikValues, Props extends object = {}>(
  { ref, ...props }: FormikWithRefProps<Props, Values>,
) {
  const formRef: MutableRefObject<FormikProps<Values> | undefined> = useRef();
  const innerRef = useCallback((formikBag: FormikProps<Values>) => {
    formRef.current = formikBag;
  }, []);
  useImperativeHandle(ref, () => formRef.current);
  return <Formik innerRef={innerRef} {...props} />;
}
```

The above is a wrapper that allows passing any ref (object/callback) to Formik.

pierroberto:
[...]

PR #2325, opened 2020-02-28
fix: Update innerRef type signature

pedroabreu:
Change of the `innerRef` prop type signature as discussed in https://github.com/jaredpalmer/formik/issues/2290

Used the type definition from @johnrom PR https://github.com/jaredpalmer/formik/pull/2222


michaldudak:
@jaredpalmer is there anything that prevents this PR from being merged in?

PR #2356, opened 2020-03-10
Add FieldProps change to v2 migration guide

### LINKED ISSUE #2350: Undocumented breaking change in FieldProps generics
## 🐛 Bug report

The `FieldProps` types has undocumented breaking changes. Previously it accepted a single generic for the `FormValues`, but now it accepts two generics with defaults. This leads to breakage since the default `any` will be used for the second generic and `FormValues` is used for the field type. This is not mentioned in the migration guide or breaking changes for v2.

### Current Behavior

`FieldProps<FormValues>` gives `field.value` typed as `FormValues`. Since `FieldProps` has default generics, the required change is not obvious.

### Expected behavior

`form.values` and not `field.value` is types with `FormValues`. Alternatively, it should be mentioned as a breaking change in the migration giude / changelog.

### Reproducible example

Pass a single generic to `FieldProps` as done in < v2.

### Suggested solution(s)

Document this breaking change.

### Additional context

The current type for `FieldProps`

```ts
export interface FieldProps<V = any, FormValues = any> {
  field: FieldInputProps<V>;
  form: FormikProps<FormValues>; // if ppl want to restrict this for a given form, let them.
  meta: FieldMetaProps<V>;
}
```

The previous type:

```ts
export interface FieldProps<V = any> {
  field: {
    /** Classic React change handler, keyed by input name */
    onChange: FormikHandlers['handleChange'];
    /** Mark input as touched */
    onBlur: FormikHandlers['handleBlur'];
    /** Value of the input */
    value: any;
    /* name of the input */
    name: string;
  };
  form: FormikProps<V>; // if ppl want to restrict this for a given form, let them.
}
```

### Your environment

| Formik           | 2.1.2
| React            | not relevant
| TypeScript       | 3.8.2
| Browser          | not executed in browser
| npm/Yarn         | yarn 1.21.1
| Operating System | Ubuntu 19.10

mhelmer:
`FieldProps` had an undocumented breaking change introduced in v2. For users that relied on the previously unary type generic, the `FormValues` parameter ended up as the type on `field.value`.

This PR adds documentation for this change in the v1 to v2 migration guide. This fixes #2350 

## [`FieldProps` in v1.6.8](https://github.com/jaredpalmer/formik/blob/v1.5.8/src/Field.tsx#L32):

```ts
export interface FieldProps<V = any> {
  field: {
    /** Classic React change handler, keyed by input name */
    onChange: FormikHandlers['handleChange'];
    /** Mark input as touched */
    onBlur: FormikHandlers['handleBlur'];
    /** Value of the input */
    value: any;
    /* name of the input */
    name: string;
  };
  form: FormikProps<V>; // if ppl want to restrict this for a given form, let them.
}
```
## [`FieldProps` in v2.1.4](https://github.com/jaredpalmer/formik/blob/v2.1.4/packages/formik/src/Field.tsx#L14):

```ts
export interface FieldProps<V = any, FormValues = any> {
  field: FieldInputProps<V>;
  form: FormikProps<FormValues>; // if ppl want to restrict this for a given form, let them.
  meta: FieldMetaProps<V>;
}
```
-----
[View rendered docs/migrating-v2.md](https://github.com/mhelmer/formik/blob/fieldprops-type-migration/docs/migrating-v2.md)

### DIFF
--- a/docs/migrating-v2.md
+++ b/docs/migrating-v2.md
@@ -42,6 +42,8 @@ resetForm({ values: nextValues /* errors, touched, etc ... */ });
 
 ### Typescript changes
 
+#### `FormikActions`
+
 **`FormikActions` has been renamed to `FormikHelpers`** It should be a straightforward change to import or alias the type
 
 **v1**
@@ -54,6 +56,22 @@ import { FormikActions } from 'formik';
 
 ```tsx
 import { FormikHelpers as FormikActions } from 'formik';
+```
+
+#### `FieldProps`
+
+**`FieldProps` now accepts two generic type parameters.** Both are optional but `FormValues` needs to be passed as the second parameter.
+
+**v1**
+
+```tsx
+type Props = FieldProps<FormValues>;
+```
+
+**v2**
+
+```tsx
+type Props = FieldProps<FieldValue, FormValues>;
 ```
 
 ## What's New?
--- a/docs/migrating-v2.md
+++ b/docs/migrating-v2.md
@@ -60,7 +60,7 @@ import { FormikHelpers as FormikActions } from 'formik';
 
 #### `FieldProps`
 
-**`FieldProps` now accepts two generic type parameters.** Both are optional but `FormValues` needs to be passed as the second parameter.
+**`FieldProps` now accepts two generic type parameters.** Both parameters are optional, but `FormValues` has been moved from the first to the second parameter.
 
 **v1**
 
--- a/packages/formik/MIGRATING-v2.md
+++ b/packages/formik/MIGRATING-v2.md
@@ -55,7 +55,7 @@ import { FormikHelpers as FormikActions } from 'formik';
 
 #### `FieldProps`
 
-**`FieldProps` now accepts two generic type parameters.** Both are optional but `FormValues` needs to be passed as the second parameter.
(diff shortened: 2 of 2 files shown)


docs/migrating-v2.md
```diff
@@ -56,6 +58,22 @@ import { FormikActions } from 'formik';
 import { FormikHelpers as FormikActions } from 'formik';
 ```
 
+#### `FieldProps`
+
+**`FieldProps` now accepts two generic type parameters.** Both are optional but `FormValues` needs to be passed as the second parameter.
```
johnrom, line 63: "Both are optional but `FormValues` has been moved from the first to the second parameter."

The wording currently sounds like "it's optional but _needs_ to be passed."

jorilallo: Would be great to elaborate what `FieldValue` means

mhelmer: @johrom: absolutely, I'll update.

@jorilallo: I agree it would be good, but I think it would be better to link into the TypeScript docs. However, it doesn't contain a description of those types.

jorilallo: @mhienle to be clear, I was going through the docs and I had no idea what `FieldValue` meant, yet alone that it's from TS definitions itself, so an example would not hurt. Same for the actual docs, there's not a single example

jaredpalmer:
LGTM

jaredpalmer:
Thanks @jorilallo

PR #2384, opened 2020-03-24
Fix type for setFieldValue

niecnasowa:
setFieldValue returns Promise with {[field_name]: field_value}

### DIFF
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -86,7 +86,7 @@ export interface FormikHelpers<Values> {
   /** Manually set values object  */
   setValues(values: Values, shouldValidate?: boolean): void;
   /** Set value of form field directly */
-  setFieldValue(field: string, value: any, shouldValidate?: boolean): void;
+  setFieldValue(field: string, value: any, shouldValidate?: boolean): Promise<Values>;
   /** Set error message of a form field directly */
   setFieldError(field: string, message: string): void;
   /** Set whether field has been touched directly */


Stas-Buzunko:
why hasn't it been merged?

danhorvath:
Is there an issue with this PR? how come it's not been merged yet?

GLObus303:
Is there a plan to merge that PR? Is that correctly marked as an issue? By very basic test I could see promise being returned from the helper.

ocho-cesar:
@jaredpalmer I think this PR is trivial and would be nice to get merged, thanks in advance

SpencerKaiser:
I found #3465 and #2648 that all come back to this same issue from 2020... what's the hold up on getting this merged??

I ended up having to use `patch-package` to make this change locally because it's preventing me from using `FormikProps<MySchema>` 😕

Renaud009:
Please merge this PR!

cactusjames:
`setFieldValue` actually returns errors from validation, if `shouldValidate` is `true` or `undefined`, so I think the type should be:

```typescript
    setFieldValue(field: string, value: any): Promise<FormikErrors<Values> | undefined>;
    setFieldValue(field: string, value: any, shouldValidate: true): Promise<FormikErrors<Values> | undefined>;
    setFieldValue(field: string, value: any, shouldValidate: false): Promise<undefined>;
```

Kepro:
3 years... (in a couple of days)

Temetias:
dancing-coffin-coffin-dance

kendallroth:
@jaredpalmer would it be possible to get this integrated into Formik? This would be very helpful, as it takes diving into source code to find out that `setFieldValue` actually returns a promise 🤷

Issue #2378, opened 2020-03-20
Formik validation works incorrectly with FieldArray `move` helper


maksimsemenov:
### Current Behavior
When I use `FieldArray` `move` helper it creates wrong `errors` object:
```
{
    arrayField: [undefined]
}
```

### Expected behavior
Validation after using `FieldArray` `move` helper, should not produce an incorrect result

### Reproducible example
https://codesandbox.io/s/blissful-fast-zpbyl

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           |2.1.4|

CoreyAR:
I did some digging around and I found that the `move` callback uses `updateArrayField` with `alterTouched` and `alterErrors` set to true. When these values are set to true they reuse the move function as the `alterTouched` and  `alterErrors` function. 

I believe it does this so that any field with errors also has its errors moved. I do agree that something is off, the "error" appears and disappears after every other move. In reading through the code this appears to be due to implementation beyond the FieldArray. Although this does appear follow documented behavior, I'm curious if there isn't a solution. 

https://jaredpalmer.com/formik/docs/api/fieldarray#fieldarray-validation-gotchas

CoreyAR:
With validation this seems to cause larger issues - https://codesandbox.io/s/still-haze-x0xf4

patrickk:
Are there any updates to this issue? I'm also getting `[null]` whenever invoking `.remove()` with `FieldArray`

pavelspichonak:
Any updates? Using `FieldArray` I have errors which looks like `{ myArrayField: [null, null, 'some error'] }` and when I use `remove(2);` errors object looks like `{ myArrayField: [null, null] }` and form is marked as invalid even though it is valid.

coreyar:
@jaredpalmer What is the level of interest in resolving this issue? I am happy to take it across the finish line

jannikbuschke:
Im also seeing this error.
On every move() I see `{ myArray: [null] }` for .5 seconds or so, and then `{ }`. Sometimes the validation error does not get removed and stays.

fwielstra:
I'm running into this issue as well and spent quite a lot of time on trying to work around it (I check the `errors` object to determine if my save button should be enabled; ended up having to flatten and filter out undefined / falsy values before checking). I would really like to see this issue fixed or an official workaround for it.

PR #2399, opened 2020-03-31
Update FieldArray error and touched

coreyar:
Fixes issue #2378 

Originally if `alterErrors` or alterTouched was not passed and the errors or touch were set to update, the alter function was being used to update the respective object. This had unindented side effects and did not produce the desired result.

This PR updates the `alterErrors` to return the validated errors by default. This should work for all methods as Yup handles the shape of the validation object.  The validated errors will be a string for array level errors (e.g. two many items in the array) or an array of strings for item level  errors (e.g. one item doesn't meet a minimum length requirement). I added a custom function to handle `alterTouched` for the `.move`. It sets the final location of the moved item as touched.

I attempted to add tests to for this update but ran into issues with the component not staying mounted so the `useRef`callback was setting `isMounted.current = false;` and thus not setting the errors. I am not so familiar with `react-testing-library` and some of the strategies in some place look like errors are getting set with internals which doesn't seem to jive with the philosophy behind `react-testing-library`.
I removed the error setting in the `.remove` test because the shape didn't match what is returned by Yup and it wasn't used in the test.


CoreyAR:
When testing this PR I noticed somethings that should probably be confirmed
- `.pop` has no validation or touch
- `.swap` creates an array the length of the list with all set to null
- `.insert` adds null to touched array
- `.replace` has no effect on touched
- `.unshift` adds null to the touched array

CoreyAR:
This PR seeks to solve issues with `.insert`. I haven't checked it out yet.

https://github.com/jaredpalmer/formik/pull/1094/files

coreyar:
@johnrom Is there interest in this PR? It was brought to my attention again because a comment was left on the issue

PR #2420, opened 2020-04-15
Fix: return onSubmit promise result in submission

### LINKED ISSUE #2419: submitForm not maintaining the result of resolved Promise
### Current Behavior

The promise returned by `props.submitForm` is not being returned from `onSubmit`.

```
onSubmit={async values => {
        console.log("Submitting", values);
        return new Promise(resolve => {
          console.log("in promise");
          resolve(777);
        });
      }}

...

onClick={async () => {
      if (!props.isSubmitting) 
         const result = await props.submitForm();
         console.log("onClick result", result);
         if (result === 777) console.log("advancing");
      }
   }}
```

results in console output
```
Submitting Object {firstName: "m", lastName: "m"}
in promise 
onClick result undefined
```

### Expected behavior

```
Submitting Object {firstName: "m", lastName: "m"}
in promise 
onClick result 777
advancing
```

The return value of `onSubmit`, if it is a resolved promise, should be the return value of `await props.submitForm()`.

An example would be doing an api call, and wanting to know the result before rejecting the submission.

### Reproducible example

https://codesandbox.io/s/agitated-yonath-nrwwo?file=/src/App.js

### Suggested solution(s)

At https://github.com/jaredpalmer/formik/blob/26c4f8627a5ecfd81ec2196c7a9687b3f39f2836/packages/formik/src/Formik.tsx#L777, we have:

```
 return Promise.resolve(promiseOrUndefined)
            .then(() => {
              if (!!isMounted.current) {
              }
            })
            .catch(_errors => {
              if (!!isMounted.current) {
                // This is a legit error rejected by the onSubmit fn
                // so we don't want to break the promise chain
                throw _errors;
              }
            });
```

In the second line, it drops the result of the resolved `promiseOrUndefined`. Changing it to 
```
 return Promise.resolve(promiseOrUndefined)
            .then((result) => {
              if (!!isMounted.current) {
              }
              return result;
            })
```

would maintain the value of the user's promise

### Your environment

| Formik           |  2.1.4
| React            |  16.12.0
| TypeScript       |
| Browser          |  Chrome, Firefox
| npm/Yarn         | 
| Operating System | Windows 10

mmartinsky:
Fixes #2419 , maintaining the value of the user's promise.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -775,10 +775,11 @@ export function useFormik<Values extends FormikValues = FormikValues>({
           }
 
           return Promise.resolve(promiseOrUndefined)
-            .then(() => {
+            .then(result => {
               if (!!isMounted.current) {
                 dispatch({ type: 'SUBMIT_SUCCESS' });
               }
+              return result;
             })
             .catch(_errors => {
               if (!!isMounted.current) {


packages/formik/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "formik",
```
johnrom, line 2: switch this back to formik

packages/formik/package.json
```diff
@@ -115,5 +115,8 @@
       "path": "./dist/formik.umd.production.js",
       "limit": "14 kB"
     }
-  ]
-}
+  ],
```
johnrom, line 118: remove

packages/formik/src/Formik.tsx
```diff
@@ -775,10 +775,11 @@ export function useFormik<Values extends FormikValues = FormikValues>({
           }
 
           return Promise.resolve(promiseOrUndefined)
-            .then(() => {
+            .then((result) => {
```
johnrom, line 778: remove parens

jaredpalmer:
Can we only format the changed lines?

johnrom:
If the formatting comes from commit hooks that might be a problem. In sourcetree there is the option to "Bypass commit hooks". I don't think this exists in Github Desktop. If it comes from prettier / vscode then it is as simple as "ctrl/cmd+shift+p -> Save without formatting" or disabling formatting temporarily.

mmartinsky:
Thanks for the quick responses - pushed up a new version with just the needed lines modified, formatting was from User level settings on my VSCode - should be taken care of now though

mmartinsky:
Any word on when this might get merged? Would love to switch back off of my forked package

cubeghost:
Also interested in seeing this get merged!

rmincling:
+1

PR #2460, opened 2020-05-06
Update field.md

gregfenton:
minor docs fix.

*Question*: the quotation section on line 30 says that "`component` ... props could also be used for rendering.".  But we are actually saying that <Field component> is valid.  This seems ... confusing.  Is it that `render` is now deprecated but `component` is not, or is the use of `component` here different than `<Field component>` ?

-----
[View rendered docs/api/field.md](https://github.com/gregfenton/formik/blob/patch-1/docs/api/field.md)

### DIFF
--- a/docs/api/field.md
+++ b/docs/api/field.md
@@ -10,7 +10,7 @@ attribute to match up with Formik state. `<Field />` will default to an HTML
 
 ## Rendering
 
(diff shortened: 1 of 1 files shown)


gregfenton:
Any thoughts to the question I asked in the original comment?

PR #2461, opened 2020-05-07
Add customization for error shape of useField FieldMetaProps

gitpash:
This is improvement for `FieldMetaProps` error typing by providing second optional `Generic` type, default is still `string` so if you are not intended to change it - no actions needed 
minimal reproducing example is [here](https://codesandbox.io/s/little-brook-4obzc?file=/src/form.tsx)
I understand that it could be solved in other way(like separating these fields) but in some cases it may be comfortable to have such feature
I was also trying to update related `.md` files, hope that cover them all

-----
[View rendered docs/api/useField.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/docs/api/useField.md)
[View rendered docs/migrating-v2.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/docs/migrating-v2.md)
[View rendered packages/formik/MIGRATING-v2.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/packages/formik/MIGRATING-v2.md)
[View rendered website/versioned_docs/version-2.0.3/api/useField.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.0.3/api/useField.md)
[View rendered website/versioned_docs/version-2.0.3/migrating-v2.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.0.3/migrating-v2.md)
[View rendered website/versioned_docs/version-2.0.4/api/useField.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.0.4/api/useField.md)
[View rendered website/versioned_docs/version-2.0.4/migrating-v2.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.0.4/migrating-v2.md)
[View rendered website/versioned_docs/version-2.1.0/api/useField.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.1.0/api/useField.md)
[View rendered website/versioned_docs/version-2.1.2/api/useField.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.1.2/api/useField.md)
[View rendered website/versioned_docs/version-2.1.2/migrating-v2.md](https://github.com/gitpash/formik/blob/feat/meta-field-error-type/website/versioned_docs/version-2.1.2/migrating-v2.md)

### DIFF
--- a/packages/formik/src/Field.tsx
+++ b/packages/formik/src/Field.tsx
@@ -11,10 +11,10 @@ import { useFormikContext } from './FormikContext';
 import { isFunction, isEmptyChildren, isObject } from './utils';
 import invariant from 'tiny-warning';
 
-export interface FieldProps<V = any, FormValues = any> {
+export interface FieldProps<V = any, FormValues = any, E extends {} = string> {
   field: FieldInputProps<V>;
   form: FormikProps<FormValues>; // if ppl want to restrict this for a given form, let them.
-  meta: FieldMetaProps<V>;
+  meta: FieldMetaProps<V, E>;
 }
 
 export interface FieldConfig<V = any> {
@@ -73,9 +73,9 @@ export type FieldAttributes<T> = GenericFieldHTMLAttributes &
 
 export type FieldHookConfig<T> = GenericFieldHTMLAttributes & FieldConfig<T>;
 
-export function useField<Val = any>(
+export function useField<Val = any, Err extends {} = string>(
   propsOrFieldName: string | FieldHookConfig<Val>
-): [FieldInputProps<Val>, FieldMetaProps<Val>, FieldHelperProps<Val>] {
+): [FieldInputProps<Val>, FieldMetaProps<Val, Err>, FieldHelperProps<Val>] {
   const formik = useFormikContext();
   const {
     getFieldProps,
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -136,7 +136,9 @@ export interface FormikHandlers {
     : (e: string | React.ChangeEvent<any>) => void;
 
   getFieldProps<Value = any>(props: any): FieldInputProps<Value>;
-  getFieldMeta<Value>(name: string): FieldMetaProps<Value>;
+  getFieldMeta<Value, Error extends {} = string>(
+    name: string
+  ): FieldMetaProps<Value, Error>;
   getFieldHelpers<Value = any>(name: string): FieldHelperProps<Value>;
 }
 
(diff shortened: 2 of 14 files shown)


packages/formik/src/types.tsx
```diff
@@ -267,11 +269,11 @@ export type GenericFieldHTMLAttributes =
   | JSX.IntrinsicElements['textarea'];
 
 /** Field metadata */
-export interface FieldMetaProps<Value> {
+export interface FieldMetaProps<Value, Error> {
   /** Value of the field */
   value: Value;
   /** Error message of the field */
-  error?: string;
+  error?: Error;
```
BrentFarese, line 276: Is this the correct type?  I just ran across an error in the types for `FieldMetaProps` when using `useField`.  Currently, the type is `string` as it is here, but it actually seems to be able to be either `string` or the shape of `Value`.  So, I was contemplating submitting a PR that changed the type to: `error?: FormikErrors<Value> | string`.  But then I ran across this PR and thought it could be done here.

The `Error` generic type parmater you have is nice for customizing the `error` that is returned (either string or some custom object that the user passes in), but I think the default of the generic `Error` type parameter should not be just `string` but `FormikErrors<Value> | string`?  

In reality, if the `Value` you use for a Formik field is not a plain string, the error object you get back matches the shape of the `Value` (I think), so it might be more helpful.  It still seems useful to have your `Error` generic type parameter, but I think the default type should be different (maybe changed to `FormikErrors<Value> | string`)...

packages/formik/src/Field.tsx
```diff
@@ -73,9 +73,9 @@ export type FieldAttributes<T> = GenericFieldHTMLAttributes &
 
 export type FieldHookConfig<T> = GenericFieldHTMLAttributes & FieldConfig<T>;
 
-export function useField<Val = any>(
+export function useField<Val = any, Err extends {} = string>(
```
BrentFarese, line 76: In line with comment below, the default type here and elsewhere might be changed to?

```suggestion
export function useField<Val = any, Err extends {} = FormikErrors<Val> | string>(
```

packages/formik/src/types.tsx
gitpash: @BrentFarese thanks for suggestions! do you mind to create reproducible example on error you'd find on `useFiled` feel free to use codesandbox example from above, or any you'd like. I'm gonna try your suggested default value as soon as will have some free time, but things are not super fast in this repo, so I guess no rush

PR #2475, opened 2020-05-16
Fix Field docs

dxts:
* Fix the field docs to show correct passed through props for `children` and `component`. 
* Make clear component is deprecated.

### DIFF
--- a/docs/api/field.md
+++ b/docs/api/field.md
@@ -15,17 +15,16 @@ There are 2 ways to render things with `<Field>`.
 - `<Field as>`
 - `<Field children>`
 - ~~`<Field render>`~~ _deprecated in 2.x. Using these will log warning_
-- `<Field component>`
+- ~~`<Field component>`~~ _deprecated in 2.x. Using these will log warning_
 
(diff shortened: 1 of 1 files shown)


oleg-brizy:
If `component` is deprecated. How to access field errors? Because they are not accessible in `Field.as`.

dxts:
You can use a callback with `children`, as it passes the `meta` prop to the callback, which includes the error. You can find an example [here](https://jaredpalmer.com/formik/docs/api/field#children).

oleg-brizy:
I already have a component and don't want to wrap it in callback to forward exactly the same props.

Before: `<Field component={MyComponent} />`
After: `<Field additionalProp="foo">{props => <MyComponent {...props} />}</Field>`

----

Do you know why the `component` prop is deprecated?

jaredpalmer:
I think Field component is not deprecated. It was in an early beta / rc of 2.x

dxts:
The documentation should make it clear that `meta` is not passed to component, and there was a typo in the `component` method, making it seem like the field helpers were passed to it.
Is this good to go now?

PR #2485, opened 2020-05-21
added onSubmitCancelledByFailingValidation 

fabb:
h1. Motivation

It is desirable to execute business logic when the user submits the form, but validation fails. 

We need the functionality in order to send analytics information and other side effects on failed form validation when the user tried to submit.

h1. Proposed Solution

This PR adds an additional config parameter to `FormikConfig`: `onSubmitCancelledByFailingValidation`.

```
    const formProps = useFormik({
        initialValues: myInitialValues,
        validate: myValidateFunction,
        onSubmit: mySendForm, // is only executed when validation succeeds
        onSubmitCancelledByFailingValidation: myFallbackOnFailedValidation,
    })
```
(The same when using `Formik`, and similar when using `withFormik`)

This would be backwards-compatible, and also it would separate the general validation (that can happen when e.g. just typing in an input, or blurring a field) from validation that is triggered by trying to send the form.

h1. Alternatives Considered

h2. Using a custom subcomponent to observe state change in FormikHelpers

Like this: https://github.com/jaredpalmer/formik/pull/2103#issuecomment-574565897

This is not very nice and might break when React Concurrent Mode lands, and [Automatic batching of multiple setStates happens (which even happens in Blocking Mode)](https://reactjs.org/docs/concurrent-mode-adoption.html#feature-comparison).

h2. A second parameter to the `validate` function

An alternate solution idea would be to add a second parameter to the `validate` function to signal that the validation is being executed because the user is trying to send the form: `validate?: (values: Values, validationTrigger: 'field-changed' | 'field-blurred' | 'submit') => void | object | Promise<FormikErrors<Values>>`

```
    const formProps = useFormik({
        initialValues: myInitialValues,
        validate: (values, validationTrigger) => {
            const result = myValidateFunction()
            if (validationTrigger === 'submit') {
                myFallbackOnFailedValidation()
            }
            return result
        },
        onSubmit: mySendForm, // is only executed when validation succeeds
    })
```

This would also be backwards-compatible, but not as elegant on the client side, and also would not work for a Yup `validationSchema` if used.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -136,6 +136,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   isInitialValid,
   enableReinitialize = false,
   onSubmit,
+  onSubmitCancelledByFailingValidation,
   ...rest
 }: FormikConfig<Values>) {
   const props = {
@@ -143,6 +144,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     validateOnBlur,
     validateOnMount,
     onSubmit,
+    onSubmitCancelledByFailingValidation,
     ...rest,
   };
   const initialValues = React.useRef(props.initialValues);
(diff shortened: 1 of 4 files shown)


packages/formik/src/Formik.tsx
```diff
@@ -739,8 +741,13 @@ export function useFormik<Values extends FormikValues = FormikValues>({
 
   const submitForm = useEventCallback(() => {
     dispatch({ type: 'SUBMIT_ATTEMPT' });
-    return validateFormWithHighPriority().then(
-      (combinedErrors: FormikErrors<Values>) => {
+    return validateFormWithHighPriority()
+      .catch(error => {
```
fabb, line 745: the new callback is called both when the validation returns field errors, and also when the validation throws. this is debatable. should a throwing validation also call the callback, or is it rather a sign that the client forgot to implement proper error handling? also a thrown validation error would be catchable by `handleSubmit`, so the question is if "duplicate" handling by the new callback makes sense here.
also IF it makes sense, should the caught error also be returned as parameter to the callback?

fabb:
@jaredpalmer please review

Issue #2490, opened 2020-05-26
 `onSubmit` or `handleSubmit` swallows all errors.


adamerose:
### Current Behavior

- Any errors occurring inside `onSubmit` get automatically caught and replaced with a nondescript console warning.  

![image](https://user-images.githubusercontent.com/10884874/82859446-bf73f080-9ee4-11ea-8e01-19e3b80df344.png)

### Expected behavior

- Errors should be thrown normally and propagated up

### Reproducible example

- https://codesandbox.io/s/formik-example-mkfcp?file=/src/App.js

### Additional context

- Only happens in versions `2.0.9` and onward
- This is the code causing it https://github.com/jaredpalmer/formik/blob/f117c04738ed218b5eb8916d7189e0849962d50d/packages/formik/src/Formik.tsx#L841
- This breaks global error handlers or anything that tries catching errors outside the `onSubmit` method itself

### Suggested solution(s)

- Revert that change. I don't think Formik should silently swallow errors, or at least there should be an option to make it not do that.

johnrom:
As described on the linked PR, this is a non-issue. `handleSubmit` is meant as an event handler that catches errors and doesn't interrupt the user in the case of catastrophic failure.

If you want to handle errors in some custom way, you should use the `submitForm` method instead, as demonstrated in the sandbox I linked in the PR.

adamerose:
> As described on the linked PR, I think this is a non-issue. `handleSubmit` is meant as an event handler that catches errors and doesn't interrupt the user in the case of catastrophic failure.
> 
> If you want to handle errors in some custom way, you should use the `submitForm` method instead, as demonstrated in the sandbox I linked in the PR.
> 
> Custom error handling is possible like this:
> 
> https://codesandbox.io/s/formik-example-sqyxe?file=/src/App.js

After considering your alternative and the problems with my pull I agree, thanks for taking a look at this.

PR #2491, opened 2020-05-26
Fix `onSubmit` or `handleSubmit` swallows all errors. #2490

adamerose:
Fixes issue #2490

When I ran `yarn test` some tests failed but I'm pretty sure they're unrelated to my code and just popped up because this issue was just hiding the errors

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -830,12 +830,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         }
       }
 
-      submitForm().catch(reason => {
-        console.warn(
-          `Warning: An unhandled error was caught from submitForm()`,
-          reason
-        );
-      });
+      submitForm()
     }
   );
 


packages/formik/src/Formik.tsx
```diff
@@ -830,12 +830,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         }
       }
 
-      submitForm().catch(reason => {
-        console.warn(
-          `Warning: An unhandled error was caught from submitForm()`,
-          reason
-        );
-      });
+      submitForm()
```
johnrom, line 833: Wouldn't we need to return the Promise here in order to catch anything?:

https://codesandbox.io/s/formik-typescript-playground-dg32w?file=/pages/basic.tsx

Also, trailing semicolon please.

johnrom:
In addition to the changes that were requested above, `handleSubmit` is _not_ meant for this functionality. `handleSubmit` is meant to be an event handler which swallows errors so that the user can continue even if the submit function fails.

_the `submitForm` function should not cause an error_ unless one is thrown from the onSubmit callback (which you should instead act on within the onSubmit callback). But if it's a possibility that you want to catch, you should use the `submitForm` function instead as demonstrated in the sandbox that I link above. I recommend instead catching everything within submitForm and performing error handling there.

PR #2521, opened 2020-06-10
add validate and validationSchema to formikBag

### LINKED ISSUE #2458: Formik bag doesn’t contain `validationSchema`
### Current Behavior

`useFormikContext()`’s return value does not include `validationSchema`.

### Expected behavior

`useFormikContext()`’s return value includes `validationSchema`.

### Reproducible example

(pending)

### Suggested solution(s)

`useFormik` returns a formik bag which does not include `validationSchema`:

```
  const ctx = {
    ...state,
    // --- SNIP ---
    validateOnBlur,
    validateOnChange,
    validateOnMount
    // -- SNIP ---
  };

  return ctx;
}
```

Adding `validationSchema` to the `ctx` would make it available in `useFormikContext`.

### Your environment

| Formik           | master |
| React            | 16.13.1
| TypeScript       | 3.8.3
| Browser          | (any)
| npm/Yarn         | yarn 1.22.4
| Operating System | macOS Catalina 10.15.4 (19E287)

evertbouw:
These properties should be part of the context according to the [types](https://github.com/jaredpalmer/formik/blob/master/packages/formik/src/types.tsx#L245)

fixes #2458

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -990,6 +990,8 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     validateOnBlur,
     validateOnChange,
     validateOnMount,
+    validate: props.validate,
+    validationSchema: props.validationSchema,
   };
 
   return ctx;


bash-plentific:
This would be a great contribution, it's currently blocking us. Could we please get this release in a new revision?

johnrom:
In v3 these values will likely come from `FormikConfigContext` as their inclusion in the normal FormikContext will cause the entire context to update frequently in the event the developer generates their validate function inline in their component.

You can see this in action here: #3121

bash-plentific:
@johnrom thanks for prompt response! is there a timeline for when v3 will be launched?

johnrom:
No idea, I put forward a PR for it but I'm not a maintainer.

PR #2554, opened 2020-06-25
Update migration guide

### LINKED ISSUE #2220: Change in behaviour of isValid
### Current Behavior
isValid returns true when the form is just rendered (i.e not dirty) as errors object is empty (no errors).

### Previous behavior
isValid depended on isInitialValid when not dirty, hence by default returned false.

Since isValid is meant to be used to make buttons and stuff disabled, now this change has brought a change. And provided, there are too many instances of isValid in an application, making the upgrade would require a lot of effort.

### Suggested solution(s)

Ideally, a form should not be valid when rendered as it requires some user interaction at the very least. Hence, when the form is not dirty, it should not be valid. 2 solutions:
1. Do not deprecate isInitialValid.
2. set isValid = false, when the form in not dirty. (But the user cannot change this behaviour, so we may need something like setIsValid?). 

Also, I think this deprecation detail should be added in the migrating docs. I found out this inconsistency after the tests failed :p 

### Your environment

| Formik           | 2.x
| React            | 16.8
| TypeScript       | 3.x
| Browser          | All
| npm/Yarn         | Both
| Operating System | MacOS

maxaggedon:
Warn about breaking changes

Fixes #2220

### DIFF
--- a/docs/migrating-v2.md
+++ b/docs/migrating-v2.md
@@ -40,6 +40,18 @@ resetForm(nextValues);
 resetForm({ values: nextValues /* errors, touched, etc ... */ });
 ```
 
+### `isValid`
+
+This property does not take the value of `dirty` into account anymore. This means that if you want to disable a submit button when the form is not `dirty` (i.e. on first render and when values are unchenged), you have to explicitly check for it.
+
+```tsx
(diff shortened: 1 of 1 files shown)


jaredpalmer:
Moved to updated docs #2612

PR #2556, opened 2020-06-25
Extend setValue to use React.SetStateAction functionality

PatoBeltran:
This is a continuation of https://github.com/formik/formik/pull/1513 which was closed for inactivity. This PR has addressed all the comments and reused React.SetStateAction types.

With this, formik consumers could pass a value or a function to setValues just like you can with React's setState dispatchers.


james-mckenzie-cko:
is there any news on this being merged? We've just run into an issue using setValues where state isn't updated properly because of React state batching.

If anyone runs into issues in the meantime, as a temporary workaround, I'm using this:

```
const useSetFormikValues = <T>() => {
  const { values, setValues } = useFormikContext<T>();
  const [formState, setFormValues] = useState(values);

  useEffect(() => {
    setValues(formState);
  }, [formState, setValues]);

  return { setFormValues };
};
```

jaredpalmer:
This makes a lot of sense to me. Can you fix merge conflicts?

leothorp:
Would it make sense to extend this to setFieldValue as well?

PatoBeltran:
@jaredpalmer fixed merge conflicts

Ronny25:
Hi all!
Any news on this?
I think it's quite an important change for a lot of Formik users.

Ronny25:
@jaredpalmer maybe it also makes sense to have the same functionality for **setValues** helper method from **useField** hook?
As sometimes you may want to be subscribed to nested object changes and update only certain properties in there.
Would be also nice to have similar to setState logic there too.
At least, from my point of view.

johnrom:
I think instead of providing this shortcut, we should provide an API for accessing the latest ref of `values` in Formik so it is never stale. So one could do something like the following. React's useState hook doesn't provide this type of API. Sometimes I still wish it did, though. However, I think the below is much more explicit, we know the exact moment that we got the latest values.

```js
const onChange = async((event) => {
  if (event.target.name === 'sourceField') {
    const sourceValue = event.target.value;
    const destinationValue = await calculateDestinationField(event.target.value);

    setValues({ 
      ...formik.getValues(), // or formik.valuesRef.current ??
      sourceValue,
      destinationValue,
    });
  }
}, []);
```

jaredpalmer (approved):

jaredpalmer:
Can you re-open, this was merged and then reverted?

jaredpalmer:
Nvm. I actually am going to revert- the revert. 

@johnrom it's documented in the types that you can't return a promise, just a callback. You can't do what you've just described with setState, so I don't think the async nature of it matters. Stale values is a different issue.

Ronny25:
@jaredpalmer I updated Formik to 2.1.7 and there are no changes related to `setValues` 🤷 
Is there any trick or something?

nemanjacosovic:
@jaredpalmer @SarahEW1206  @ehicks05  can you please check your distribution because the discrepancy in the dist version?

**DISTRIBUTED: Missing an option to pass a function to setValues.**
Screenshot 2020-10-07 at 13 03 11

**GitHub: Change implemented in latest 2.1.7 PR**
Screenshot 2020-10-07 at 13 01 38

rmacklin:
⚠️ A note to anyone who ends up on this pull request ⚠️ 

Despite allowing you to pass a function to `setValues`, this does not actually ensure that the "`prevValues`" passed to your function is the latest form state. **_It can be stale_**:
- https://github.com/jaredpalmer/formik/issues/3047

In other words, it does **_not_** provide the same guarantees you get when passing a function to [React's `setState` method](https://github.com/reactjs/reactjs.org/blob/e4aa5076c2f8b9bf2b035c24a28cb80e6166333f/content/docs/reference-react-component.md#setstate-setstate) (even if the pull request title makes you think it would).

PR #2586, opened 2020-07-06
Remove empty hooks in production build output

wmzy:
`__DEV__` is a constant. It won't change the order of hooks.

image


jaredpalmer:
Smart!

Issue #2606, opened 2020-07-14
Tutorial: getFieldProps example breaks styles


rahulbhooteshwar:
## 📖 Documentation
[getFieldProps Example](https://formik.org/docs/tutorial#getfieldprops) removes type attribute from form fields & thus breaks the existing styles.
It may lead to confusion for the beginners & they can loose focus from the main agenda of the tutorial.

Issue #2604, opened 2020-07-14
Tutorial: Formik Installation step is missing - Local Development Environment Setup


rahulbhooteshwar:
## 📖 Documentation
[Tutorial - Local Development Environment Setup](https://formik.org/docs/tutorial#setup-option-2-local-development-environment)
Setup steps tells about creating new React app, deleting existing src demo code & creating the starter code (styles.css & index.js).
After that, users are told to run the project using yarn start.

These instructions misses very important step of ***installing Formik***

That step should be added to avoid confusions for beginners.

PR #2607, opened 2020-07-14
Fix Docs: Tutorial, getFieldProps example & local development environment setup

rahulbhooteshwar:
Fixes: #2606 
Fixes: #2604
* Formik installation step is missing in basic development environment setup
* Due to missing type attribute styles are breaking for input fields in current documentation.


rahulbhooteshwar:
It also fixes #2604

PR #2627, opened 2020-07-17
Fix field prop types

m-rutter:
Presently `Field` using `FieldAttributes<any>` as its prop type. Because `FieldAttributes<T>` uses its generic parameter as part of its top level intersection this effectively means that `Field`'s prop type is `any`.

https://github.com/formium/formik/issues/2524

### DIFF
--- a/packages/formik/src/Field.tsx
+++ b/packages/formik/src/Field.tsx
@@ -126,7 +126,7 @@ export function useField<Val = any>(
   ];
 }
 
-export function Field({
+export function Field<T extends object = {}>({
   validate,
   name,
   render,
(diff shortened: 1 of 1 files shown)


m-rutter:
Ah, I was hoping that this was a quick fix, but given the CI results it looks like this problem is a little harder.

johnrom:
If I remember correctly, not only do the types have to be carried carefully all the way through the types, Formik had an issue at least in v1 where the generic of FieldProps<T> referred to the FieldValue in one type and the FormValues in another type.

johnrom:
`T` here stands for both `FieldValue` in FieldConfig, and Props on FieldAttributes itself. The thing is -- what did people use FieldAttributes for before? `Props` or `FieldValue`? I want to say that because FieldValue cannot never be assigned to FieldAttributes, we should go with Props. We'll need further testing for it though.

```ts
export type FieldAttributes<T> = GenericFieldHTMLAttributes &
  FieldConfig<T> &
  T & { name: string };
```

PR #2673, opened 2020-08-07
Update DocSearch v3 implementation

francoischalifour:
This updates the DocSearch v3 integration to the latest version:

- Support for typing query when the search button is focused
- Improved accessibility

The DocSearch integration on this website uses two instances: one for mobile, one for desktop. This resulted in a double modal being shown when hitting <kbd>Ctrl + K</kbd>, and then users has to click twice on the overlay to close it. I solved this by checking that no other DocSearch instance is showing before opening a new one. We might want to do this check in DocSearch itself in the future.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -12,7 +12,8 @@
   },
   "dependencies": {
     "@babel/preset-typescript": "^7.10.1",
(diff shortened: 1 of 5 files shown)


jaredpalmer:
Thanks!

PR #2705, opened 2020-08-17
Make onSubmit/handleSubmit optional

### LINKED ISSUE #2675: Please make onSubmit optional for use cases that don't involve submitting forms
### Current Behavior

The `Formik` component's `onSubmit` is required. Failing to use it causes a runtime error when submitting and a TypeScript error (if you use it). You can work around this by passing a no-op function (`onSubmit={() => {}}`), but I think it's misleading for beginners and makes it seem like you need some specific code inside `onSubmit`.

### Desired Behavior

 `Formik` should allow the `onSubmit` prop to be omitted.

### Suggested Solution

Make the type optional in TypeScript (using `undefined` or `?:` in the prop type) and simplify disable any submitting functionality when `onSubmit` is not given. Any validation and other synthetic events that don't rely on submitting should still work.

Any form state management use case that doesn't involve submitting forms, for example:
- Learning how to use Formik without a server
- Creating a prototype of an app that doesn't have a server implemented yet
- Calculating results based on form fields immediately as the user changes/blurs fields (I have multiple calculator apps that do this)
- Persisting a bunch of fields to local storage without having to manage form state manually

There could be another library for managing state of form fields that aren't for submitting, but personally I think Formik is good enough at that already despite this issue that it wouldn't be very worth the effort.

### Additional context

When I first learned about Formik, I avoided it because I saw the `onSubmit` prop and assumed I had to have some sort of REST/GRAPHQL server/service in order to use it. I didn't realize until later that I could just use it for state management when someone told me about the no-op prop.

nickmccurdy:
This allows the `onSubmit` prop of `Formik` to be omitted, both at runtime and in TypeScript. This makes Formik easier to learn and use in situations where it's used for form state management without submission.


nickmccurdy:
I've noticed somewhere in the docs that `setSubmitting(false)` is recommended for synchronous `onSubmit` handlers. In this pull request, nothing happens when `onSubmit` is `undefined`. Should `setSubmitting(false)` or anything else still be called when submitting isn't necessary?

nickmccurdy:
I have updated this with master and added a changeset.

nickmccurdy:
Merged in from master

johnrom:
I don't think I understand the purpose of this PR really. Are you saying that when a user hits a submit button, nothing should happen? I think that is covered by a constant void function like `doNothingOnSubmit`.

I think instead of letting onSubmit be optional internally, the onSubmit can default to `noop`. (I copied the following from an experimental branch so it is a little bit different than the main branch)

```ts
export const useFormik = <Values extends FormikValues = FormikValues>(
  rawProps: FormikConfig<Values, FormikRefState<Values>>
): FormikRefApi<Values> => {
  const {
    validateOnChange = true,
    validateOnBlur = true,
    validateOnMount = false,
    enableReinitialize = false,
-    onSubmit,
+    onSubmit = noop,
    ...rest
  } = rawProps;
  const props = {
    validateOnChange,
    validateOnBlur,
    validateOnMount,
    onSubmit,
    ...rest,
  };
```

PR #2707, opened 2020-08-17
Use low priority validation for FieldArray

mhiley:
We've been experiencing issues with laggy text input fields in our application, specifically when using `FieldArray`, even on forms with a small number of fields (less than 10). Using `FastField` has been an acceptable workaround in certain cases but has its own drawbacks, so I tried out the suggestion in https://github.com/formium/formik/issues/2542 in our application and confirmed that using low priority validation does fix our performance issues (or is at least a very significant improvement).

I don't believe this change affects the current API at all - it just exposes `validateFormWithLowPriority` so that it can be used in place of `validateForm` (alias of `validateFormWithHighPriority`) in `FieldArray`. This appears to be working well in our application and I can't see any obvious drawbacks, but let me know if there are any issues with running the validation with low priority here.

Thanks for taking a look!


pietmichal (approved):

mhiley:
Thanks @pietmichal - btw, we've been running this as a patch in our application for about two weeks now and it's been working well so far.

jaredpalmer (changes requested):
Can you rename the prop to `unstable_validateFormWithLowPriority` as I don't think I want to fully document this yet?

mhiley:
> Can you rename the prop to `unstable_validateFormWithLowPriority` as I don't think I want to fully document this yet?

@jaredpalmer For sure, will try to get to it soon.

mhiley:
@jaredpalmer It looks like this PR is no longer relevant as of https://github.com/formium/formik/pull/2891? Do you have any ideas on how to address performance concerns with `FieldArray` now that low priority validation has been removed?

jaredpalmer:
Formik 3 alpha which I'm about to release in a few minutes

mhiley:
Closing since this is no longer a valid strategy as of https://github.com/formium/formik/pull/2891

PR #2437, opened 2020-04-23
added: ability to pass submit context to onSubmit handler

avasuro:
This closes #214, closes #1792

-----
[View rendered docs/api/formik.md](https://github.com/avasuro/formik/blob/feat/add-submit-context/docs/api/formik.md)

PR #2710, opened 2020-08-18
Include FormikState in onReset/onSubmit callback

### LINKED ISSUE #2709: onSubmit function signature to include all of FormikState, not just "values"
### Current Behavior

In the `onSubmit` handler, you only have access to `values` and `formikActions`. It is impossible to access e.g. the `touched` object.

### Desired Behavior

It would be great to be able to access `touched` in the `onSubmit` handler. One possible use case is to only send changed form values to an API (re-sending unchanged form values generally increases the risk of write conflicts, when others submit the same form with stale values).

### Suggested Solution

Pass a reference to `touched`, or all of `FormikState` to `onSubmit`. Either as a 3rd argument, or change the type of the 2nd argument to `FormikHelpers & FormikState` or so, both should maintain backwards compatibility.

Anyone wanting to utilize the entirety of state handling that Formik provides.

An alternative is to avoid `onSubmit` entirely (you need at least an empty implementation as it's a non-optional property), and have the submit button of the form call a custom function that handles the form submit, as you generally have access to the `touched` object where the submit button is rendered. I don't think that is the design goal of this library however.

jonaswitt:
Resolves #2709 

Let me know if anything else should be changed, I tried to update the docs as best as possible.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -416,7 +416,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       if (props.onReset) {
         const maybePromisedOnReset = (props.onReset as any)(
           state.values,
-          imperativeMethods
+          imperativeMethodsAndState
         );
 
         if (isPromise(maybePromisedOnReset)) {
@@ -841,7 +841,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     }
   );
 
-  const imperativeMethods: FormikHelpers<Values> = {
+  const imperativeMethodsAndState: FormikHelpers<Values> & FormikState<Values> = {
     resetForm,
 
     validateForm: validateFormWithHighPriority,
@@ -856,10 +856,12 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     setValues,
     setFormikState,
     submitForm,
+
+    ...state,
   };
 
   const executeSubmit = useEventCallback(() => {
-    return onSubmit(state.values, imperativeMethods);
+    return onSubmit(state.values, imperativeMethodsAndState);
   });
 
   const handleReset = useEventCallback(e => {
--- a/docs/src/pages/docs/2.1.4/api/formik.md
+++ b/docs/src/pages/docs/2.1.4/api/formik.md
@@ -318,21 +318,19 @@ an input from uncontrolled to controlled.
 Note: `initialValues` not available to the higher-order component, use
 `mapPropsToValues` instead.
 
-### `onReset?: (values: Values, formikBag: FormikBag) => void`
+### `onReset?: (values: Values, formikHelpersState: FormikHelpers & FormikState) => void`
 
 Your optional form reset handler. It is passed your forms `values` and the
 "FormikBag".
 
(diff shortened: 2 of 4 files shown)


amaechler:
This is also somewhat related to #2365 which adds access to the `initialValues` in the submit handler.

jonaswitt:
> This is also somewhat related to #2365 which adds access to the `initialValues` in the submit handler.

Good point, thanks! I was not aware of that PR when I started this one. The change in #2365 would have probably been sufficient for me to reach my initial goal (although I was looking for the `touched` data, having `initialValues` would be just as good). 

Now, who can get either of these PRs merged?

johnrom:
In v3 the Formik API will have a function called `getState()` which will return the entireity of Formik's internal state and can be used in a callback like this.

#3231 

```jsx
onSubmit={(values, formik) => {
  // note that `values` above have been validated
  // whereas the values returned from `getState()` below may be future unvalidated values
  const { errors, touched, initialValues, ...etc } = formik.getState();
}
```

On the contrary, using the state where this PR is using it is completely unsafe (in theory), as it will only be as current as the render which assigned the callback to onSubmit().

packages/formik/src/Formik.tsx
```diff
@@ -856,10 +856,12 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     setValues,
     setFormikState,
     submitForm,
+
+    ...state,
   };
 
   const executeSubmit = useEventCallback(() => {
-    return onSubmit(state.values, imperativeMethods);
+    return onSubmit(state.values, imperativeMethodsAndState);
```
johnrom, line 864: I would combine them here, like `{ ...imperativeMethods, ...state }`, instead of doing it on every render.

packages/formik/src/Formik.tsx
```diff
@@ -416,7 +416,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       if (props.onReset) {
         const maybePromisedOnReset = (props.onReset as any)(
           state.values,
-          imperativeMethods
+          imperativeMethodsAndState
```
johnrom, line 419: I would combine them here, like `{ ...imperativeMethods, ...state }`, instead of doing it on every render.

johnrom (commented):
Since this is a slightly different use case than v3's getState, I would support adding this functionality, however I wouldn't spread state on every render -- only when the callbacks are executed.

I might also consider separating it into its own prop, like:

```
(values, helpers, snapshot)
```

Then you could do something like:

```
if (snapshot === getState()) {
  // state has not changed
}
```

Or if we were doing this with a major version + breaking changes, I might do:

```
(snapshot, helpers) // values is already within snapshot
```

johnrom:
Another thing to consider is we already have a request to add a prop to onSubmit for context, #2437, basically just any old information that needs to be sent, like click event or whatever.

jonaswitt:
> Since this is a slightly different use case than v3's getState, I would support adding this functionality, however I wouldn't spread state on every render -- only when the callbacks are executed.

Thanks for your feedback on this PR and sorry I lost track of it for a little bit. I have pushed the suggested changes. Maybe it can still be helpful before Formik 3 ships.

PR #2719, opened 2020-08-26
Bugfix/console

adebiyial:
Fix SVG attribute names


docs/.sample.env
```diff
@@ -1,2 +0,0 @@
-NOTION_TOKEN=XXXX
-BLOG_INDEX_ID=XXXX
```
jaredpalmer, line 2: Can you add the files back? Don’t want to delete these

adebiyial: It's done now.

PR #2726, opened 2020-09-01
a little cleaner

AminAtaee:


LautaroLobo12 (approved):

johnrom:
@AminAtaee thank you for your PR. Sorry that nobody reached out regarding this. Generally, we don't accept PRs that simply lint or refactor code for the sake of refactoring. If it was done as part of a larger issue, it's definitely a valid change, but with limited resources we need to focus our releases on solutions for issues folks are having or new features.

Thanks again!

PR #2748, opened 2020-09-17
corrected the spelling of automatically in the <Field /> description

nefejames:


jaredpalmer:
This is intentional actually. It’s a play of the word automatic and magic

nefejames:
Ah. Okay then

Issue #2327, opened 2020-02-28
FastField not updating Checkbox value on toggle


simkessy:
I'm trying to control if my points input is enabled or disabled. When I click on checked, it disabled as expected. When I uncheck it remains disabled. I first tried setting `disabled={isChecked}` then I created it's own state I'd toggle using `useEffect` neither work. Once it's disabled, it won't toggle back to enabled. 
   
      const [isChecked, setIsChecked] = useState(false);
    
      const handleChecked = event => {
        handleChange(event);
        setIsChecked(checked => !checked);
        setFieldValue(`questionData.${idx}.answers.${answer}.point`, 0);
        setFieldTouched(event.target.name);
      };

        <FormControlLabel
            control={
              <Checkbox
                name={`field1`}
                onChange={handleChecked}
                value={isChecked}
              />
            }
            label="Disable points"
            style={{ marginTop: 15, marginLeft: 8 }}
        />
        <FastField
          as={InputField}
          name={`field2`}
          label="Points"
          type="number"
          inputProps={{ min: '0', max: '100', step: '1' }}
          disabled={isChecked}
          handleError={handleError}
        />

```
export const InputField = memo(props => {
  const { handleError = () => {}, name, ...attr } = props;

  const field = name.split('.').pop();
  return (
    <TextField
      size="small"
      fullWidth
      color="secondary"
      error={!!handleError(field)}
      helperText={handleError(field)}
      name={name}
      {...attr}
    />
  );
});
```

I was able to solve this by changing from FastField to Field. I'm not entirely sure why it works now.

maddhruv:
Try the latest formik (`>= 2.2.3`), fixed with https://github.com/formium/formik/pull/2874

Issue #945, opened 2018-09-29
FastField and Field meta is undefined from v1.2


edward-sia:
## Current Behavior
I have recently upgraded formik from v1.1 to v1.3 and found that the meta prop in Field no longer working. Downgrading formik realizes that the issue starts to show up since v1.2 onwards. I assume this is a bug as it was a feature introduced in v1.1 and no change log that describes it has changed since.

## Steps to Reproduce
By switching to v1.2

In another component the following `FastField` is used
```
<FastField name='firstName' label='First name *' component={StyledTextField} />
```
StyledTextField component
```
import React from 'react'
import TextField from '@material-ui/core/TextField';
import FormHelperText from '@material-ui/core/FormHelperText';
import FormControl from '@material-ui/core/FormControl';

const StyledTextField = props => {
  const { field, form, label, meta } = props;

  console.log('Styled ', field, form, meta) // meta is undefined in v1.2 and v1.3
  return (
    <FormControl error>
      <TextField
        {...field} 
        error={meta.touched && !!meta.error}
        label={label}
      />
      { (meta.touched && meta.error) && <FormHelperText>{meta.error}</FormHelperText> }
    </FormControl>
    )
}

```

**CodeSandbox Link**:
https://codesandbox.io/s/lx9kpz2jl9

---

- Formik Version: 1.2.0
- React Version: 16.5.2
- Browser and Version: Google Chrome 69
- OS: MacOS
- Node Version: v10.11.0
- Package Manager and Version: NPM 6.4.1

jaredpalmer:
Ahh, when I rewrote FastField i forgot that we had added this. It's probably time to add it back (as a breaking change) in v2.

jaredpalmer:
Fo reference, Field meta, never existed, just FastField (which is why I forgot).

prichodko:
I can update my long-standing PR - https://github.com/jaredpalmer/formik/pull/579 which should fix the issue.

charlax:
I think this is still important, especially since the doc mentions it:

https://jaredpalmer.com/formik/docs/api/fieldarray

> _NOTE_: In Formik v0.12 / 1.0, a new `meta` prop may be added to `Field` and `FieldArray` that will give you relevant metadata such as `error` & `touch`, which will save you from having to use Formik or lodash's getIn or checking if the path is defined on your own.

maddhruv:
Closing as stale

Issue #450, opened 2018-02-22
Make FastField work with React Native


jaredpalmer:

Andreyco:
@jaredpalmer could you share what is the issue here? I could possibly take this one (since working with RN only for last couple of months) if you agree

daviscabral:
Any news on this? I can contribute on this too.

daviscabral:
I'll take a stab on this :-)

daviscabral:
Just a heads up. Not totally happy with my approach - but I hope to get something to show here soon.

brodybits:
looks relevamt

brucedesa:
Updates?

hemedani:
any update ...,
please after making work, create an example for how to this

jaredpalmer:
Seems legit

TomasUhrik:
For anyone struggling with this in the meantime until this gets fixed. Don't forget that you can manually control when should your component rerender. In class components with [shouldComponentUpdate](https://reactjs.org/docs/react-component.html#shouldcomponentupdate) and in functional component with [React.memo](https://reactjs.org/docs/react-component.html#shouldcomponentupdate).

In my case I had my input components as functional components so I did something like this:
`
export const InputFieldPrimary = React.memo(
  ({ field }) => {
    const { name } = field

    return (
      <View>
        <TextInput
          value={field.value}
        />
      </View>
    )
  },
  (prevProps, nextProps) => prevProps.field.value === nextProps.field.value
)
`

PR #2749, opened 2020-09-17
Render optimization

### LINKED ISSUE #1007: Optimizing validation for FastFields
### Current Behavior
Currently `FastField` does rendering optimization, but still its `onChange` handler triggers validation for the whole `validationSchema`, which can be costly for really huge forms.

### Desired Behavior
It would be perfect if `onChange` and `onBlur` of `FastField` could validate only slice of `validationSchema`

### Suggested Solution
`yup.reach` is the answer - https://github.com/jquense/yup#yupreachschema-schema-path-string-value-object-context-object-schema Fast fields could call a new Formik method, which validates only schema corresponding to a given `FastField`. For instance, `<FastField name="someName.nested" />` could do `schema.reach('someName.nested')` and use this for validation instead of validating the whole schema. Then, the validation result could be merger with Formik errors.

All users caring for the best possible form performance, who use Formik together with Yup, especially for big forms.

I have really big dynamic forms, sometimes +100 fields and I tried `FastField` - it is better now but for such a big form it is Yup which becomes a bottleneck - each key stroke validates 100+ fields, which could be especially costly for more expensive validators.

Alternatively I could use Field level validation, but I really like to keep validation logic only in yup, it is easy to test then and also it nicely keep good separation between yup and formik. Also, I dont like to repeat myself defining `initialValue` and `schema.cast()` does this automatically form me (provided I have all fields defined inside schema, not as field validators)

### Additional context

N/A

jaredpalmer:
This begins preparing Formik for concurrent mode. 

- [x] Add `unstable_useStrictField` and `unstable_StrictField` to exports. They actually work like `FastField` and a `useFastField` (if it existed). I left FastField as is for now because it's not 1:1 with `StrictField`, although one could argue that it really _is_ since state is stale in FastField's render props on other fields. Anyways, we can and need to debate naming and deprecations going forward.
- [x] Add a formik-codemod package in preparation for name and import changes
- [ ] Clean up and dedupe `handleChange` and `handleBlur` internals once we figure out where things are going

Closes #1007, #450, #945, #2327 

Proof of Concept: https://codesandbox.io/s/formik-use-context-selector-2-hso7i?file=/src/App.tsx

## Basic custom field

**Stays the same**

```jsx
// this is what useField does today.
// meta is safe for now, because it just contains error, touched, value and initials.
import { unstable_useStrictField as useField } from "formik";

export const CustomField = ({ label, name }) => {
  const [field, meta] = useField(name);
  return (
    <>
      <label>{label}</label>
      <input {...field} />
      {meta.error && meta.touched && <div>{meta.error}</div>}
    </>
  );
};
```

## Custom Imperative handler of the same field

**Bad (but, how it works today)**

```jsx
import { unstable_useStrictField as useField, useFormikContext } from "formik";

export const CustomField = ({ label, name }) => {
  const [field, meta] = useField(name);
  // This is bad news bears, but pretty common. Calling `useFormikContext`
  // means this component will re-render on every update to Formik,
  // even with this new PR.
  const onChange = value => {
    formik.setFieldValue(name, value);
  };
  return (
    <>
      <label>{label}</label>
      <CustomInput {...field} onChange={onChange} />
      {meta.error && meta.touched && <div>{meta.error}</div>}
    </>
  );
};
```

**Good**

```jsx
import { unstable_useStrictField as useField } from "formik";

export const CustomField = ({ label, name }) => {
  const [field, meta] = useField(name);
  // By making handleChange bound to the field and smarter about events
  // people will be able to avoid using setFieldValue(name, ...) and just do this:
  const doSomethingSpecial = () => {
    field.onChange("zop");
  };
  return (
    <>
      <label>{label}</label>
      <input {...field} />
      {meta.error && meta.touched && <div>{meta.error}</div>}
    </>
  );
};
```

## Setting the value of another field or other Formik state

**Bad**

```jsx
import { unstable_useStrictField as useField, useFormikContext } from "formik";

export const CustomField = ({ label, name }) => {
  const [field, meta] = useField(name);

  // This is bad news bears, but pretty common. Calling useFormikContext
  // means this will update on every key stroke / update to Formik,
  // even with this new PR.
  const {
    setFieldValue,
    values: { otherFieldName }
  } = useFormikContext();
  
  React.useEffect(() => {
    if (field.value === "jared") {
      setFieldValue("otherFieldName", "is cool");
    }
  }, [field.value, otherFieldName]);

  return (
    <>
      <label>{label}</label>
      <input {...field} />
      {meta.error && meta.touched && <div>{meta.error}</div>}
    </>
  );
};
```

**Good**

```jsx
import { unstable_useStrictField as useField, useSetFieldValue } from "formik";

export const CustomField = ({ label, name }) => {
  const [field, meta] = useStrictField(name);
  // By using this new hook (which uses the new context selector under the hood),
  // this component now only re-renders when there are changes
  // from useField (instead of on every key stroke and update to Formik).
  const setOtherValue = useSetFieldValue("otherFieldName");

  React.useEffect(() => {
    if (field.value === "jared") {
      setOtherValue("is cool");
    }
  }, [field.value]);

  return (
    <>
      <label>{label}</label>
      <input {...field} />
      {meta.error && meta.touched && <div>{meta.error}</div>}
    </>
  );
};
```


johnrom:
Why don't we expose all the stable methods from a single API hook like `useFormikApi()`, and expose the unstable values from something like `useFormikState()`. If someone needs both, they can do `const [state, api] = useFormikState();` since adding the API isn't expensive, but adding state to API would be. @fredyc and I fleshed out a really nice API for it, and useFormikApi and useFormikState were his suggestions. [Check out the discussion here](https://github.com/formium/formik/issues/1677#issuecomment-511559743)

The API would support useContextSelector nicely I think.

jaredpalmer:
We can have `useFormikActions()`. However, as mentioned, a big `useFormikState()` isn't performant. It will subscribe to all changes. With this PR, we can use selectors to limit updates. `useFieldValue(firstName)` will only update when that `firstName` value changes (not when `firstName.touched` or `firstName.errors` changes or when `lastName.values` changes). This solves the re-rerendering woes in the same way that recoil does: with selectors.

jaredpalmer:
@johnrom you are correct that we can just have `useFormikApi` instead of a bunch of little helpers for setters. I still had my Recoil hat on. But since we don't have atoms to pass around this isn't necessary.

sibelius:
we tested this PR and it is still rerendering everything

so the idea is to rebuild most of the hooks using `unstable_` prefix?

jaredpalmer:
Yeah i didn't implement any of the re-rerendering optimizations yet.

packages/formik/src/Formik.tsx
```diff
@@ -631,7 +631,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       if (!isString(eventOrTextValue)) {
         // If we can, persist the event
         // @see https://reactjs.org/docs/events.html#event-pooling
-        if (!!(eventOrTextValue as React.ChangeEvent<any>).persist) {
+        if ((eventOrTextValue as any).persist) {
```
johnrom: This confuses me. eventOrTextValue is already `string | React.ChangeEvent<any>`, so if `!isString(eventOrTextValue)` then eventOrTextValue will automatically be inferred as React.ChangeEvent<any>, which means there shouldn't be a cast here at all. Are we using an old typescript version? Or the wrong definition of `isString` which does not use [type predicates](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates)

```diff
-if ((eventOrTextValue as any).persist) {
+if (eventOrTextValue.persist) {
```

jaredpalmer: Idk this didn’t have anything to do with this PR it just showed up in ts. reality is that the types might be lying (if we are in RN) and also that in react 17 this is a no op anyways.

tooppaaa:
Hello, happy to help contributing to this PR with your guidance.

Please let me know if you need anything !

johnrom:
> With this PR, we can use selectors to limit updates. `useFieldValue(firstName)` will only update when that `firstName` value changes (not when `firstName.touched` or `firstName.errors` changes or when `lastName.values` changes).

I think we should consider dependent fields if we're going to take this approach. We could add an [`include=`](https://github.com/formium/formik/issues/512#issuecomment-719592805) Field-level prop for dependent fields, like:

```jsx
<Field name="fieldA" />
<Field name="fieldB" include={{ fieldA } => { fieldA }} validate={
  (value, dependentFields) => value * dependentFields.fieldA < 100
} />
```

Then this field level validation can be triggered when either fieldA or fieldB changes.

jaredpalmer:
> > With this PR, we can use selectors to limit updates. `useFieldValue(firstName)` will only update when that `firstName` value changes (not when `firstName.touched` or `firstName.errors` changes or when `lastName.values` changes).
> 
> I think we should consider dependent fields if we're going to take this approach. We could add an [`include=`](https://github.com/formium/formik/issues/512#issuecomment-719592805) Field-level prop for dependent fields, like:
> 
> ```js-jsx
> <Field name="fieldA" />
> <Field name="fieldB" include={{ fieldA } => { fieldA }} validate={
>   (value, dependentFields) => value * dependentFields.fieldA < 100
> } />
> ```
> 
> Then this field's render and field-level validation can be triggered when either fieldA or fieldB changes.

@johnrom This is a good idea. Any thoughts on whether we should call this StrictField or FastField vs. break Field?

johnrom:
I keep going back and forth on the API I suggested above. Seems like it could be done with useField and a custom prop. If we went this route, I'd prefer not to branch out into another component like FastField -- we've all seen the fallout of deprecating it. I feel like there should be one Field. I wonder if there is a way we can opt in to the new behavior at build time with a feature flag? This Context Selector business is very cool, but it's likely to be deprecated if the React team ever releases a solution which optimizes these renders via another method. For example, the input lag issue would probably be solved by Concurrent Mode.

We may want a way for users to opt into this behavior so long as they need it, and if it ever becomes the default that's great. But if it gets deprecated users should be able to opt into it for the next major version or two instead of having to rewrite their apps. Where we could put a feature flag is a whole different question. Webpack plugin? Formik prop?

Apparently I contribute more questions than answers, lol

jaredpalmer:
@johnrom Technically speaking, this implements FastField and useFastField _correctly_. Instead of having to tell people in the fast field example, "don't get other form values from in here!" this just removes them entirely from the render props. I agree it would be nice to have one version of Field, and I think we can get there, but I don't think people want to rewrite their apps. Unfortunately most fields should just work, but obviously some will not. 

A couple of options,

- Make this FastField and useFastField, breaking old FastField (which was janky anyways). Tell people to start using FastField everywhere. Deprecate Field with warning?
- Move existing Field and FastField to a new `formik-compat` package. Add a codemod that rewrites imports to LegacyField and LegacyFastField and field to LegacyField.
- Explore feature flags?

johnrom:
Codemods sound cool if webpack/babel plugin & feature flags are just a crazy idea. I've never looked at how to build them, just thought it would be a modern approach for deprecating commonly used functionality. .Net uses these kinds of flags a lot in the compilation process which allows them to support tons of new functionality while not breaking old code.

FilipK999:
Hey @jaredpalmer, I'm glad you've just finished this feature as I'm building a really huge form in Formik and this is needed a lot, since it's really laggy. Could you tell me roughly when can we expect this change released?

jj-crypto (approved):

PR #2761, opened 2020-09-22
added back missing default shouldUpdate check (see https://github.com…

nico1510:
same fix as in https://github.com/formium/formik/pull/1317 but it got lost in this commit https://github.com/formium/formik/commit/c93a6d4ad059bb8684f48c2d54524f417add2030. Fixes https://github.com/formium/formik/issues/1238 (again)


johnrom:
I disagree with this change. This basically means a user cannot optimize _further_ than the default FastField allows.

For example, the default implementation is:

```js
return props.name !== this.props.name 
  || getIn(props.formik.values, this.props.name) !== getIn(props.formik.values, this.props.name) 
  || getIn(props.formik.errors, this.props.name) !== getIn(props.formik.errors, this.props.name)
  || getIn(props.formik.touched, this.props.name) !== getIn(props.formik.touched, this.props.name)
  || Object.keys(this.props).length !== Object.keys(props).length 
  || props.formik.isSubmitting !== this.props.formik.isSubmitting;
```

If, for example, my field doesn't care about `touched` or `errors`, I might pass:

```diff
<FastField shouldUpdate={(props, prevProps) => 
  props.name !== prevProps.name 
  || getIn(props.formik.values, prevProps.name) !== getIn(props.formik.values, prevProps.name) 
- || getIn(props.formik.errors, prevProps.name) !== getIn(props.formik.errors, prevProps.name)
- || getIn(props.formik.touched, prevProps.name) !== getIn(props.formik.touched, prevProps.name)
  || Object.keys(prevProps).length !== Object.keys(props).length 
  || props.formik.isSubmitting !== prevProps.formik.isSubmitting;
```

With this change, that wouldn't work.

johnrom:
I think a better option might be exposing the Formik default FastField check. FastField might be deprecated in the future if we can optimize Field enough, so maybe it wouldn't be worth even focusing on this. But if instead you did:

```ts
const defaultFastFieldShouldUpdate = <Values, Props>(
  props: FastFieldInnerProps<Values, Props>, 
  prevProps: FastFieldInnerProps<Values, Props>
) => 
  props.name !== prevProps.name 
  || getIn(props.formik.values, prevProps.name) !== getIn(props.formik.values, prevProps.name) 
  || getIn(props.formik.errors, prevProps.name) !== getIn(props.formik.errors, prevProps.name)
  || getIn(props.formik.touched, prevProps.name) !== getIn(props.formik.touched, prevProps.name)
  || Object.keys(prevProps).length !== Object.keys(props).length 
  || props.formik.isSubmitting !== prevProps.formik.isSubmitting;
```

```ts
// in FastField
shouldComponentUpdate(props: FastFieldInnerProps<Values, Props>) {
    const shouldUpdate = this.props.shouldUpdate ?? defaultFastFieldShouldUpdate;

    return shouldUpdate(props, this.props);
}
```

Then in your component you can choose to import defaultFastFieldShouldUpdate from Formik and do:

```jsx
<FastField shouldUpdate={
  (props, prevProps) => 
    myCustomShouldUpdate(props, prevProps) || defaultFastFieldShouldUpdate(props, prevProps)
  }
/>
```

jaredpalmer:
FastField should be deprecated after we implement use-context-selector and strict useField. I don't see a huge point in adding new features to it.

PR #2766, opened 2020-09-23
Add function parameter to setValues

MMDF:
You can now pass on a function to setValues just like a React hook.


MMDF:
This is a really nice comfort feature imo. I needed it in my app so i came to add it here.

PR #2774, opened 2020-09-25
Add enableReinitialize to the context

chrisneven:
In some component down the tree in my app I tried to check if `enableReinitialize`  was turned on. However, its value is always undefined. I think it should be added to the formik bag because there could be scenarios where it's needed. Nevertheless, the typings would make you believe it's accessible but it's actually not.


johnrom (commented):
I'm not against this, enableReinitialize is a boolean so it wouldn't be bad for performance or anything. I have, however, suggested we should re-do enableReinitialize so that we can pass a custom reinitialization function that can solve some issues like choosing whether to reset touched during reinitialization or whatnot. This might set the wrong precedent if we went that route. Then again, enableReinitialize still be a boolean we pass if onReinitialize or whatever is truthy, so I'm still fine with adding it.

chrisneven:
Oh a custom reinitialization function? That'd be cool! Well, if this might set the wrong precedent, then I understand you won't add it in. I've found another way to know if the `enableReinitialize` has been set, so it's not much of a problem. The only thing that got my mind a little twisted is that the types would make you believe it's available. Also, I do think it could be handy to see whether `enableReinitialize` has been set to true, or in the future: the reinitialization function has been set.

PR #2791, opened 2020-10-01
Changed image for Benchmark to require the url

colbygarland:
Just a simple change to fix the broken logo for Benchmark UAE.


jaredpalmer:
Fixed here f464b3f

PR #2794, opened 2020-10-01
Extend setValue to use React.SetStateAction functionality

jaredpalmer:
Reverts formium/formik#2793

### DIFF
--- a/.changeset/great-grapes-clean.md
+++ b/.changeset/great-grapes-clean.md
@@ -1,0 +1,10 @@
+---
+"formik": minor
+---
+
+`setValue` can now optionally accept a function as a callback, exposing `React.SetStateAction` functionality. Previously, only the entire object was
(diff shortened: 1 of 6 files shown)


rmacklin:
⚠️ A note to anyone who ends up on this pull request ⚠️ 

Despite allowing you to pass a function to `setValues`, this does not actually ensure that the "`prevValues`" passed to your function is the latest form state. **_It can be stale_**:
- https://github.com/jaredpalmer/formik/issues/3047

In other words, it does **_not_** provide the same guarantees you get when passing a function to [React's `setState` method](https://github.com/reactjs/reactjs.org/blob/e4aa5076c2f8b9bf2b035c24a28cb80e6166333f/content/docs/reference-react-component.md#setstate-setstate) (even if the pull request title makes you think it would).

PR #2812, opened 2020-10-09
Added `isFirstIn` utility function

henkerik:
Added the `isFirstIn` utility function. This function is useful to implement a scroll to first error behaviour. For example, one could write a hook using this utility function: 

```
export const useOnFirstError = (name: string, scrollRef: React.RefObject<HTMLElement>) => {
  const { errors, isSubmitting, isValidating } = useFormikContext()

  useEffect(() => {
    if (isSubmitting && !isValidating && isFirstIn(errors, name)) {
      scrollRef.current?.scrollIntoView()
    }
  }, [errors, name, isSubmitting, isValidating, scrollRef])
}
```

The `isFirstIn` function works a bit similar to the `getIn` function. It accepts an object (like the Formik errors object) and a formik field name. Given these two values it determines if the given name is the first to occur in the given object. 

The order of the keys in the errors object is determined by the order of the fields in the Yup validationSchema, so one can order these in the same order as the fields appear on the page. 

Finally, one can use this hook in the following way: 

```
import React, { useRef } from 'react'
import { useField, useFormikContext } from 'formik'

import { useOnFirstError } from './useOnFirstError'

type Props = { 
  name: string
}

const InputField = ({ name }: Props) => {
  const [field, meta, helpers] = useField(name)
  const inputRef = useRef<HTMLInputElement>(null)

  useOnFirstError(name, inputRef)

  return <input {...field} ref={inputRef}  />
}

<Formik ...>
  <InputField name="firstName" />
  <InputField name="lastName" />
</Formik>
```

The `isFirstIn` function is a small function, but it would be beneficial to include this function in Formik. This will allow consumers to use it without requiring a direct dependency on Lodash. Secondary, it's behaviour is strongly coupled to the behaviour of `setIn` en `getIn`, changing one of these will like also mean changing the `isFirstIn` function, so it would make sense to keep these three functions closely together :) 

A more complete version of the `useOnFirstError` hook can be found here:

https://github.com/dhlparcel/formik/pull/2 

We can add this hook to this PR if desired.


jaredpalmer:
I spoke with Dan Abramov about this exact thing use case yesterday. We agreed on the following:

The reality is that this functionality really can't be easily done in pure React with refs because we actually don't know if the order of the fields in the errors object actually matches the order in the UI. We also don't know if the field is still mounted just by looking at the errors object either (becuase of conditional rendering or suspense).

The way to do this is with the DOM. One solution we came up with is to add data attribute to mark a scrollable area and then use an effect to scroll to the dom node when you need to. This works well with conditional logic and is safe in Concurrent Mode. 

```
<Form>
  <div data-scrollable="name">
    ...
  </div>
  {something && 
	<div data-scrollable="name">
     ...
    </div>
  }  
  <div data-scrollable="address">
    ...
  </div>
</Form>

// Somewhere else...
useEffect(() => {
	// get the list of fields in the order they show in the DOM
	const orderedFieldsElements = document.querySelectorAll('[data-scrollable]')
    smoothScrollIntoViewIfNeeded(orderedFieledElenets.some(el => ... ))
}, [errors])
```

PR #2813, opened 2020-10-09
Hook to scroll to the first error of a Formik form

harshadpandit:
This PR adds the functionality to scroll to the first error of a formik form. This is done using a react hook upon formik validation. When there are validation errors as defined in the Yup validation schema then this hook checks in the `errors` object from `FormikContext` to see if the given element is the first one to appear and if it is then does the necessary action of scrolling to that element and optionally bringing it in focus if it is an `input` element. This hook can be placed selectively on any `input | div | span` but not limited to these aforementioned elements.

The order of the keys in the formik errors object is determined by the order of those fields in the Yup validationSchema, so one can order these in the same order as the fields appear on the form.

The hook can be used in the following way:
```
import React, { useRef } from 'react'
import { useField, useFormikContext } from 'formik'

import { useOnFirstError } from './useOnFirstError'

type Props = { 
  name: string
}

const InputField = ({ name }: Props) => {
  const [field, meta, helpers] = useField(name)
  const inputRef = useRef<HTMLInputElement>(null)

  useOnFirstError(name, inputRef)

  return <input {...field} ref={inputRef}  />
}

<Formik ...>
  <InputField name="firstName" />
  <InputField name="lastName" />
</Formik>
```
This PR uses a function `isFirstIn` from the other PR which can be found [here](https://github.com/formium/formik/pull/2812)

This is a very common use case which can be supported in Formik forms.

### DIFF
--- a/packages/formik/src/useOnFirstError.ts
+++ b/packages/formik/src/useOnFirstError.ts
@@ -1,0 +1,19 @@
+import { useEffect } from 'react'
+import { useFormikContext } from 'formik'
+
+import { isFirstIn } from './utils'
+
+export const useOnFirstError = (
+  name: string,
+  scrollRef: React.RefObject<HTMLElement>,
+  focusRef?: React.RefObject<HTMLInputElement>
+) => {
+  const { errors, isSubmitting, isValidating } = useFormikContext()
+
+  useEffect(() => {
+    if (isSubmitting && !isValidating && isFirstIn(errors, name)) {
+      focusRef?.current?.focus()
+      scrollRef.current?.scrollIntoView({ behavior: 'smooth', block: 'center', inline: 'center' })
+    }
+  }, [errors, name, isSubmitting, isValidating, scrollRef, focusRef])
+}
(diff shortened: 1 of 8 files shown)


packages/formik/src/useOnFirstError.ts
```diff
@@ -0,0 +1,31 @@
...
+  scrollRef: React.RefObject<HTMLElement>,
+  focusRef?: React.RefObject<HTMLInputElement>
+) => {
+  const { errors, isSubmitting, isValidating } = useFormikContext();
+
+  useEffect(() => {
+    if (isSubmitting && !isValidating && isFirstIn(errors, name)) {
+      focusRef?.current?.focus();
+      scrollRef.current?.scrollIntoView({
+        behavior: 'smooth',
```
jaredpalmer, line 25: smooth-scroll-into-view-if-needed is a cleaner way to do this

jaredpalmer (changes requested):
I’m not certain this is a safe or reliable  implementation. The way we solve this functionality when we need it is to fallback to a vanilla dom query or use https://github.com/palmerhq/react-register-nodes (a lib we wrote specifically for this use case)

jaredpalmer:
Going to close for now as a workaround already exists, but also very likely to change in concurrent mode.

PR #2828, opened 2020-10-19
Separate State and API surface

johnrom:
This does the basic work of separating Formik into both Refs and Reducer State and will eventually separate the Formik API into both render-specific state using `useFormikState` and render-independent API surface using `useFormikApi`. Developers can then opt into the API version when they do not need render-specific values.

I want to rebase this away from the useContextSelector stuff, because I don't think they need to be entangled the way they are now -- it was an accident 😞. This is mostly an experiment to see if this can work and if this API makes sense.


packages/formik/src/Formik.tsx
```diff
@@ -52,6 +51,9 @@ type FormikMessage<Values> =
       payload: FormikState<Values>;
     };
 
+const isChangeEvent = (value: any): value is React.ChangeEvent<any> =>
+  typeof value === 'object' && typeof value.persist === 'function';
+
```
jaredpalmer, line 56: isSyntheticEvent maybe better name

packages/formik/src/Formik.tsx
```diff
@@ -171,9 +173,46 @@ export function useFormik<Values extends FormikValues = FormikValues>({
...
+  });
+
+  /**
+   * Breaking all the rules, re: "must be side-effect free"
+   * BUT that's probably OK.
+   *
+   * The only things that should use stateRef are side effects themselves --
+   * those things which need the latest value in order to compute their own latest value.
+   */
+  const refBoundFormikReducer = React.useCallback((
```
jaredpalmer, line 202: I wonder if this is safe in concurrent mode? I mean it’s unlikely people are suspending in forms, but it’s not impossible. For example, you might suspend to lazy load a date picker or something. Again, above my understanding of CM, but something we should ask React core team about

packages/formik/src/Formik.tsx
johnrom: def a better name

packages/formik/src/Formik.tsx
johnrom: good question. I have no idea, I'll have to look and see how CM works whenever I get some free time to look at this again. From what I remember, I think it means a ref could be updated but then that render abandoned, causing tearing between the two states?? which would definitely not be good for this technique.

johnrom: I had a look at Concurrent Mode and this technique should work as long as we never call `refBoundFormikReducer` outside of Effects or Setters.

What this means, however, is that users can break this pattern by doing something like:

```
const MyForm = () => {
  const { values, setValues } = useFormik(formikProps);
  const dependentValues = calculateDependentValues(values);

  // uh oh, if a render is abandoned after calculating dependentValues 
  // because someone typed another letter, this will be incorrect
  setValues(dependentValues);
  // instead, this MUST use `useEffect`.
  useEffect(() => {
    const dependentValues = calculateDependentValues(values);

    // never incorrect because this render MUST have occurred
    setValues(dependentValues);
  }, [values, setValues]);
}
```

So while it's technically incorrect, it's a _super breaking change_.

johnrom: I try to think of it this way. User interactions should _never_ be abandoned. If a user types a, then b, both value = a and value = ab should be committed. And in this scenario they will.

However, the validation for "a" _can_ be abandoned, and an effect calculating dependent values for "a" can be abandoned, simply by encapsulating useEffect, and that would still work in this scenario.

What differs in this scenario is that the validation for "a", triggered by onChange when validateOnChange is set to true, could actually end up validating "ab", then be followed by a second validation for "ab". I am having trouble reasoning about whether that is more or less correct than the setup we current have.

johnrom:
closed in favor of #2931

PR #2844, opened 2020-10-27
feat: useFormik will return the same object

0x6368656174:
Previously, the `useFormik` hook would return a new object each time. Because of this, its result could not be used as a dependency for other hooks. Now `useFormik` will return the same object if them state is not changed due to the fact that it is wrapped in `useRef`.


johnrom:
The implications of this are pretty broad, and it would be a breaking change for anyone who is currently using the Formik object as a dependency for "if Formik state has changed".

Personally, I think it makes way more sense to return a new object everytime for the purposes of listening for "Formik state has been updated" (as there would be no other possible way to achieve this), and for all hooks to be dependent on the bits of Formik that they use. If a bit of functionality needs access to all of Formik arbitrarily, it should itself be a hook that can decide its own dependencies:

```
const MyForm = () => {
  const formik = useFormik(formikProps);
  useCustomFormikFunctionality(formik);
}

const useCustomFormikFunctionality = (formik) => {
  const { values } = formik;
  useEffect(() => console.log(values), [values]);
}
```

packages/formik/src/Formik.tsx
```diff
@@ -1020,14 +1020,22 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     validateOnMount,
   };
 
-  return ctx;
+  // We want useFormik to always return the same object reference
+  // so that it can be used as a dependency for other hooks.
+  const ctxRef = React.useRef<typeof ctx>({} as any);
+  Object.assign(ctxRef.current, ctx);
```
jaredpalmer, line 1026: You can't write to refs in render in concurrent mode. So this strategy won't work in future versions of React

0x6368656174:
@johnrom Yes, I agree. It might make sense to return a new object every time the state has changed. But now ʻuseFormik` returns a new object every time it is called, which is unlike the behavior of other React hooks, which usually return the same objects between calls.

packages/formik/src/Formik.tsx
0x6368656174: Thank, @jeffbski. I don't know its=) I found a good article about it if someone who will read this comment is interesting https://frontarm.com/daishi-kato/use-ref-in-concurrent-mode/#:~:text=Concurrent%20Mode%20lets%20React%20apps,%2C%20text%20input%20or%20hover).

0x6368656174: But it seems like how exactly in this case the competitive mode will not break due to the use of refs, since several renders will undermine the return of similar objects. But, I agree, we need to fix it =)

0x6368656174: As far as I understand, there will be a problem with concurrent mode if the reference changes to a different value during render. But in our case, no matter how many times the render is launched, the link will contain a similar object. Plus, before, between renders, a new object was returned every time.

There is an article about it https://frontarm.com/daishi-kato/use-ref-in-concurrent-mode/#:~:text=Concurrent%20Mode%20lets%20React%20apps,%2C%20text%20input% 20or% 20hover). And it clearly shows that the problem arises due to the fact that between renders there are changes in the reference to a new value. There is no such thing. But maybe I don't understand something =))

0x6368656174:
@johnrom I corrected the code so that it only returns a new object if the state of the form has changed. Which is more expected and shouldn't break backward compatibility.

johnrom:
@0x6368656174 isEqual is a deep comparison and is pretty expensive to perform on each render. I think we need to solve this issue on a deeper level.

In [this PR](#2828) I propose separating the state and API into two separate pieces. I think if I can get the reducer itself to return all of FormikState, then we return the following, it should solve the problem entirely:

```
// where state is the current snapshot returned by FormikReducer, and not a new object
const [state, api] = useFormikState();
```

0x6368656174:
@johnrom  The comparison will not work much longer than each render is currently in progress and a new `ctx` object is created. Since it will compare the values ​​of fields from `ctx` with fields from `ctxRef.current` and if the form has not changed, it will go through the keys from the 1st level of the object and exit the comparison, since the references of the key values ​​will match. Those are precisely the "deep" comparison will occur only in the case of a change in the state of the form, and then, the comparison will go only until 1 distinguishing, after which it will be finished.

Therefore, I think that the performance will suffer from the use of `isEqual` no more than from the fact that now, for each render, the `ctx` is being re-created.

But this fix will allow using the result of `useFormik` in other hooks, as well as passing it down the component tree in props.

0x6368656174:
We use the common form object not only in the hook dependencies, but we also want to transfer it entirely in the child widgets so as not to inflate the number of props of the components, as does `withFormik`. But, no problem=) I will then close the PR.

0x6368656174:
@johnrom After I closed the PR, on the way home, I realized that it is possible to speed up the algorithm by comparing only the state formic. I made a commit in this PR, but I will not reopen it. In general, there is a high probability that with such an implementation it will work even faster than before since every render will not create `ctx` object, but the previous one will be used if the status has not changed.

0x6368656174:
After I closed the PR, on the way home, I realized that it is possible to speed up the algorithm by comparing only the state formic. In general, there is a high probability that with such an implementation it will work even faster than before since every render will not create `ctx` object, but the previous one will be used if the status has not changed.

I reopened the PR to pull out the fix commit.

0x6368656174:
@johnrom After I closed the PR, on the way home, I realized that it is possible to speed up the algorithm by comparing only the state formic. In general, there is a high probability that with such an implementation it will work even faster than before since every render will not create `ctx` object, but the previous one will be used if the status has not changed.

I reopened the PR to pull out the fix commit.

johnrom:
> @johnrom   We use the common form object not only in the hook dependencies, but we also want to transfer it entirely in the child widgets so as not to inflate the number of props of the components, as does `withFormik`. But, no problem=) I will then close the PR.

@0x6368656174 you can use `useFormikContext` in any functional child widget to get any bit of Formik state that you need without adding any child component props. Hope it's helpful!

0x6368656174:
The PR was opened only to fill in the fixed commit.

PR #2846, opened 2020-10-27
v3

jaredpalmer:
- All hooks that return an initial thing return a single variable.
- `useIsDirty`, `useIsSubmitting`, and `useIsValid` return booleans
- All other hooks that return state also return updaters (e.g. `const [values, setValues] = useValues()`)
- Hooks that are name `useSetXXXX` just return the setter (and have no perf overhead)

Todo:

- [x] `useFieldInitialValue`
- [x] `useFieldInitialTouched`
- [x] `useFieldInitialError`
- [x] `useInitialValues`
- [x] `useInitialErrors`
- [x] `useInitialTouched`
- [x] `useInitialStatus`
- [x] `useFieldTouched`
- [x] `useFieldValue`
- [x] `useFieldError`
- [x] `useSetFieldValue`
- [x] `useSetFieldTouched`
- [ ] `useSetFieldError`
- [x] `useValues`
- [x] `useSetValue`
- [x] `useSetErrors`
- [x] `useErrors`
- [x] `useTouched`
- [x] `useSetTouched`
- [x] `useStatus`
- [x] `useSetStatus`
- [x] `useSubmitForm`
- [x] `useResetForm`
- [x] `useIsSubmitting`
- [x] `useIsValid`
- [x] `useIsDirty`
- [x] `useValidateForm`
- [x] `useValidateField`
- [ ] `useActions` ??? (bag of all methods)?
- [x] Provide way to call `validateForm` and `validateField` with hooks
- [x] Reimplement `Form`, `ErrorMessage`, and `Field`
- [ ] Fix typescript types of `Field` and `FastField` render props
- [ ] Docs
  - [ ] Update examples
  - [ ] TypeScript type explorer like https://formium.io/docs/react
  - [ ] Migration guide


johnrom:
I have a v3 competitor in my mind, but I haven't had time to put pencil to paper and get it done. This is the theory, and a pseudo code example.

The model that Formik uses is fundamentally flawed because it reverses the relationship between form state and rendering. Formik currently uses state via useState, which means that form state is a side effect of a render. However, form state is never dependent on renders in reality. Instead, Form state is a state machine, and renders are a side effect of mutations of form state. Form state must exist outside of the render, and only snapshots of it are sent to useState for the purposes of triggering renders.

Form state exists only as a side effect of user interaction, as well as any other effects a developer adds to those user interactions, applied on top of a calculated initial state. Its connection to React should only be via bindings which connect it to React's API.

The casualties of the reversal of responsibilities are evident in the issues. 

1. Too many renders. Every update to Formik requires a render in order to update state and trigger effects. This isn't a requirement, it is simply a consequence of our implementation. Using special context selectors doesn't fix this issue -- it only creates incremental micro-optimizations which might eventually make the inefficiency unnoticeable.
2. Async is problematic. Because of the nature of renders and snapshotting, Formik doesn't have a render-independent state. And because of this, the values within an async function are inherently stale and we must use workarounds in order to try and be async-like. The redux-like pattern of `if (getState().matchesExpectations) doAThing();` doesn't exist. The most important part of an async pattern is to have access to future state to check for cancellation.
3. It's nearly impossible to match touched and errors in a single render without forcing renders to wait for validation to complete. Because validation depends on state which depends on renders, it creates a race condition which takes several renders to resolve in the best case, and in the worse case ends up with stale state information presented to the user.

What is the solution? We create a Formik Core that is independent of React's implementation details and simply handles applying user interaction and effects to Formik's internal state, and then connect that state to React's functional APIs using a completely separate, replaceable process that can evolve with React. Here are two examples:

1. [useMutableSource](https://github.com/reactjs/rfcs/blob/master/text/0147-use-mutable-source.md). Formik state should not necessarily be "mutable" in the classic sense, but it should be a reducer that exists outside of renders which updates a mutable snapshot object which can trigger new renders. All effects from a developer standpoint should be added to the reducer at the time that next state is applied. The reducer result can then be applied to the mutable source in _one implementation_ of Formik's connection to React. We shouldn't use this unless you know, it is merged to 
2. `beReduxLike`. It's funny, the initial argument which I agree with from Formik is that form state is not global, so Redux isn't the place for form state. I still agree with this. However, really what this should mean is that the redux pattern should apply at the component-state level. We functionally create a mutable "formik" source, which triggers subscriptions after each reducer dispatch. Those subscriptions are subscribed to using `useFormikContext()` and `useField()`. useFormikContext now becomes something like:

```
const useFormikApi = () => {
  // FormikContext _does not expose things like .values, .touched, and other non-stable references_
  return useContext(FormikContext);
}
const useFormikContext = () => {
  const formik = useFormikApi();
  // stable reference to Formik State which will only update via subscriptions
  const [formikState, updateFormikState] = React.useState(formik.stateRef.current);
  const [subscription, setSubscription] = React.useState(null);
  // this calls updateFormikState after a reducer applies new changes. 
  // if it applies multiple times before a render, we don't care. we'll only render the latest version to the user
  useEffect(
    () => {
      setSubscription(formik.addFormEffect(formState => updateFormikState(formState)));
    }, 
    () => {
      formik.removeFormEffect(subscription);
    }
  );
  return [formikState, subscription];
}
```

UseField does the same thing, but it only does `addFieldEffect("name", includesCallback)`, where includesCallback is like `include={fields => fields.name}` and will add things like touched, values, and errors to the field state for dependent fields.

The reducer can be decorated by adding actions which affect only the reducer return value, including setting up dependent fields and things like that. I haven't had time to think that through completely.

leothorp:
@johnrom Great points, glad to see some discussion on those issues. What would be the user-facing changes to the Formik external API that would result from this? I may have misunderstood, but would the user manually handle adding and removing subscriptions to fields, like in that code snippet?

From my perspective as a Formik user, the following feel like the two biggest issues around handling/responding to Formik updates.

1. It isn't possible to write an imperative Formik update that depends on the latest value of a field to determine the next value
2. it isn't possible to know when an imperative update has completed, to run side effects after it.

Using `setFieldValue` as an example, here are the API changes I'd propose to resolve both of these. these are fairly simple extensions of the current API.

1. make setFieldValue optionally accept an "updater" function for the value, which would be passed the latest value of the field, and return the next value.

2. make setFieldValue return a promise which resolves with the newest value of the field (post-update). (Confusingly, it already does return a promise, but this is undocumented. Not sure what the current behavior is around when exactly that resolves.)

TLDR example setFieldValue usage showing both suggested changes: 
```
          setFieldValue("count", (prevCount) => prevCount + 1).then((updatedCount) => {
            console.log("inside side effect chained after state change.");
            console.log("latest value", updatedCount); //latest value of values.count
          });
```

Also, here's a somewhat more involved codesandbox example showing what usage of both of these would look like (the SampleUsage component), as well as an example implementation of it with a simplified "Formik" component (only "values" and "setFieldValue" are implemented). 

https://codesandbox.io/s/quizzical-bohr-rrwuf?file=/src/App.js

Note that, as in the sandbox, these can actually both be achieved simply by using class component setState in Formik- #2 in particular is not possible with useState. I'd written a comment about that on the other issue, but wanted to respond in more detail here. This approach still solves the "requires a re-render to trigger a side effect" issue you mention; you can see in the console logs in the sandbox when you click the button that the side effect does not depend on a rerender to have access to the latest value. But of course that's just an internal implementation detail. 

Thanks for all your great work on this library. happy to discuss more if you have any thoughts on the above, and happy to help out with any eventual implementations of changes like the ones we're discussing.

johnrom:
@leothorp in my recommended approach, there would no longer be a need for an effect. That's because the value would immediately be updated within Formik's state. The only time there would be a delay is for the value that is rendered to the form -- that would wait until the next render. If you need an action there, that would be the time for an effect. Otherwise, you would just use something like:

```
// this is a snapshot of values as the time getValues is called, 
// not the stale value realized during the render which triggered this effect
const currentCount = formik.getValues().count;
setFieldValue("count", currentCount + 1);

console.log("latest value", formik.getValues().count); // latest value of values.count
```

leothorp:
@johnrom I see, so the update would just be treated synchronously like in redux. Seems great, though that would seem to imply a major breaking change right? (i.e., no more "values" prop in formikBag).

`If you need to wait for validation to complete, that would be another issue.` for this aspect, having a callback/promise returned would still be an intuitive API IMO (in addition to having the sync value update you describe)

muqg:
@johnrom What you described in your first comment above is also how I imagine a highly effective and performant form implementation.

A lot of Formik's current issues can easily be solved by extracting a core Formik class and then using a subscription model to integrate it with React's primitives. Such aproach would still allow for selective input rerenders and even allow to skip renders when nothing has actually changed.

Don't know if it is really relevant, but I once made myself a POC implementation like this https://github.com/muqg/react-model/blob/master/src/Model.ts This is far from what Formik provides, but I found this implementation to never need more than a single render, allow for selective rerenders, and allow all form methods to always operate on latest form values/errors.

You are probably already familiar with these ideas, but just thought I'd leave this here anyway. Keep up the good work, guys!

johnrom:
@jaredpalmer I think the lack of what I will call "dynamic slicing" is going to be a problem as the Formik API continues to evolve. The basic issue is this: You need to know what hooks need to be called in order to build a component with these useContextSelector-based hooks. You must be able to write a component which selects specific bits of Formik's state. You can't subscribe to a "random cross-section". And you cannot change those hooks based on dynamic props. If you need a random cross-section, you'll have to subscribe to all of Formik's state and select it yourself, leading back to where we are today.

A specific API which I'm thinking would be impossible with this is one I've been thinking of for a while:

```
<Field 
  type="number"
  name="width" 
  include={fields => fields.height} 
  validate={(value, height) => value * height.value > 100 
    ? "Max of 100 square feet."
    : ""
  }
/>
```

I believe my subscription based approach [here](https://github.com/formium/formik/pull/2931#issuecomment-763166978) would enable that API in a performant way.

jaredpalmer:
Why wouldn't useContextSelector be able to handle this (assuming you memoized the selector)?

johnrom:
1. From what I understand, useContextSelector is a pretty simple subscription which is limited to using `Object.is` comparison, which means if the selector returns an object, the object will never match, and the context will always update. Anytime `values`, `errors`, or `touched` are updated, all of the Fields would in theory update due to a new object created by useField(). However, your PR gets around this limitation by breaking useField into separate micro-hooks, like `useFieldValue`, `useFieldTouched`, etc. This optimization cannot be used for dynamic slices of Formik's state, unlike they can with a custom subscription. My Subscription PR allows a user to pass a comparer like `Object.is` (default), but could also use `_.isEqual` or any other simple comparison callback. useField would now consist only of one subscription instead of dozens. 

2. Due to using micro-hooks, in this PR every use of `<Field />` uses 18 separate subscriptions to the useContextSelector provider, as well as other subscriptions in other places to access the API, which isn't necessary in my PR because I've made sure the API itself is stable. You can check all the subscriptions here for a simple form with 3 fields. Note there are (50) total listeners to useContextSelector.

image

Meanwhile, in my PR, each Field only uses one subscription, as well as a normal Context subscription to access the API, which never updates due to stable references. Note there are 3 total subscribers, with 1 dispatch effect each.

image

And if multiple fields were all listening to, for example, the same `selectFieldMetaByName("firstName")`, there would only be one subscription checking for equality, and when it is determined to have change, it can simply trigger all three listeners with the same value instead of checking each individually. Note this is 1 total subscriber, with 3 dispatch effects.

image

dai-shi:
For reference: https://github.com/dai-shi/use-context-selector/issues/19#issuecomment-766467359

johnrom:
@dai-shi custom equality is only one difference between my implementation and useContextSelector. 

1. does useContextSelector provide direct access outside of Context?

not using the context (except to access the API) means you can use the internal values wherever you have access to the API, with my PR you can do something like:

```
const formik = useFormik(...props);
// you can subscribe to the "mutable source" without a context, much like you will be able to after useMutableSource lands
const formikSlice = useFormikStateInternal(formik, selectIsSubmitting, Object.is);
```

Or you could even pass a ref to formik and subscribe to it from any context you want (well, in theory, haven't tried). 

2. my subscriber provides the underlying object via getState() which is guaranteed to always have the latest value, without waiting for React commits. is that available in useContextSelector?

this particular latest value guarantee is necessary due to the fact that in many cases form state is not a function of React effects, but a function of user input which does not wait for effects to complete.

```tsx
onSubmit={async (values, api) => {
  const result = await checkApiThing();
  // make sure these form values aren't stale
  const stillRelevant = isEqual(values, api.getValues());
  
  if (result.success && stillRelevant) {
    doImportantStuff(values);
  }
}}
```

dai-shi:
@johnrom sorry, i'm not familiar with context of the discussion here in formik.
I'm a bit confused. It sounds like the two points you are asking are not like React state and React context.

`useContextSelector(ctx, selector)` is just like `selector(useContext(ctx))` semantically, and only adds optimization. Like you can only read it in render. So, I guess answers to the questions are both "no"s?

johnrom:
@dai-shi thanks for your response. I'm familiar with the power of useContextSelector in optimizing state and suggested its use for Formik many months ago now. But after a lot of thinking, I have come to the conclusion that Formik should avoid using Context except to pass down its API and instead provide state access via direct and internally controlled subscriptions. There are plenty of packages that do this out of the box, like useSubscription, redux, etc, but the implementation in my PR is specifically crafted to handle the large number of similar subscriptions and updates that Formik generates by having dozens or more components subscribing to reusable cross-sections of state.

dai-shi:
Got it. Yeah, form store makes sense if that's the requirement. (I mean I know both needs. like, some of my libs are for react state and some are for module state.) It's just a bit unfortunate..

johnrom:
@dai-shi would you be interested in reviewing my subscription implementation? I wrote the subscription part completely independent of Formik's general API and it is suitable to be broken into a separate package. There is overlap between this and useContextSelector, and I wonder if they can work together as a method to subscribe to Context.

The two files related to subscriptions are here:

[subscription-helpers.ts](https://github.com/formium/formik/pull/2931/files?file-filters%5B%5D=.ts#diff-aedf5631785750bddec44440cdbdd40acf28e2c9669fd6534ee7ec1ef8492371)
[useSubscriptions](https://github.com/formium/formik/pull/2931/files?file-filters%5B%5D=.ts#diff-509c38ecfcb55353c481c1202ee97151e49002f45f92896a289758877e0b9510)

jaredpalmer:
Is there a material performance difference between sharing a subscription and creating new ones with useContextSelect for forms with < 100 fields? Does this actually matter to end users?

johnrom:
I didn't set out to optimize the subscription mechanism per se when I created my PR. I set out to include a subscription API that can allow us to:

1. use custom equality comparisons to prevent Field re-renders
2. dynamically slice formik state
3. get the current state at any time (`getState()`), and
4. provide access to the subscriptions without a context at all, as described [here](https://github.com/formium/formik/pull/2846#issuecomment-766896142). 

I couldn't find anything that fit the bill (including useMutableSource), so in order to proof that, I had to write a new subscription mechanism. Could definitely ship with useContextSelector level of performance (as long as common subscriptions don't use deep equality), but performance is not my first goal here.

dai-shi:
> The two files related to subscriptions are here:
> 
> [subscription-helpers.ts](https://github.com/formium/formik/pull/2931/files?file-filters%5B%5D=.ts#diff-aedf5631785750bddec44440cdbdd40acf28e2c9669fd6534ee7ec1ef8492371)
> [useSubscriptions](https://github.com/formium/formik/pull/2931/files?file-filters%5B%5D=.ts#diff-509c38ecfcb55353c481c1202ee97151e49002f45f92896a289758877e0b9510)

It's more complicated than expected. Maybe because it's tree structure with subscriptions to avoid duplicated invocation of selectors?
I didn't go details, but it's the store outside React. Looks close to zustand, another library I maintain.
Honestly, I don't see any fundamental issues with this approach in Legacy Mode.
To prepare for Concurrent Mode, you want to consider the migration path to useMutableSource. (It's still questionable if it's really necessary, depending on use cases and requirements.)

dantman:
The API appears to be missing `useIsValidating`.

dantman:
The various `use*` hooks need explicit return values, **especially** the ones that use tuples.

Take useValues for example:

```ts
export function useValues<Values>() {
  const state = useFormikContextSelector<
    Values,
    FormikContextType<Values>['values']
  >(ctx => ctx.values);
  const update = useFormikContextSelector<
    Values,
    FormikContextType<Values>['setValues']
  >(ctx => ctx.setValues);
  return [state, update];
}
```

The type that TypeScript infers is `(Values | ((values: SetStateAction<Values>, shouldValidate?: boolean | undefined) => void))[]`. What we actually want is `[Values, ((values: SetStateAction<Values>, shouldValidate?: boolean | undefined) => void)]`.

PR #2854, opened 2020-11-03
Fix low priority validation race condition.

### LINKED ISSUE #2841: Changes made to Formik 2.2.1 causing validation to briefly show error onBlur even though no error exists
### Current Behavior

My form is set to validate onBlur. After first field is blurred, the form is validated and all fields are marked with appropriate errors, but remain untouched (so no error is displayed in UI). However, when updating any subsequent fields, those fields have their touched property set to true, which takes place before field is validated and error can be removed. Within milliseconds, validation is run and error is removed.

Effect is to make error message flicker in field. This never happened before updates made to 2.2.1.

Formik

### Expected behavior

During onBlur event, Touched should not be set to true until after validation has occurred.

### Suggested solution(s)

Revert or refine whatever changes were made in version 2.2.1

### Your environment

| Formik            | 2.2.1 |
| React            | 16.14.0 |
| TypeScript       | N/A |
| Browser          | Chrome (latest) |
| npm/Yarn         | 6.14.8 |
| Operating System | Windows 10 Pro |

umidbekkarimov:
Fixes #2841

I also bumped `react`, `react-dom`, and `scheduler`.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -46,6 +46,10 @@ type FormikMessage<Values> =
   | { type: 'SET_FIELD_ERROR'; payload: { field: string; value?: string } }
   | { type: 'SET_TOUCHED'; payload: FormikTouched<Values> }
   | { type: 'SET_ERRORS'; payload: FormikErrors<Values> }
+  | {
+      type: 'SET_LOW_PRIORITY_ERRORS';
+      payload: { values: Values; errors: FormikErrors<Values> };
+    }
   | { type: 'SET_STATUS'; payload: any }
   | {
       type: 'SET_FORMIK_STATE';
@@ -72,6 +76,23 @@ function formikReducer<Values>(
       }
 
       return { ...state, errors: msg.payload };
+    case 'SET_LOW_PRIORITY_ERRORS':
+      if (
+        // Low priority validation can occur after high priority validation and
+        // this will create stale validation results, e.g:
+        // SET_FIELD_VALUE-------------------SET_LOW_PRIORITY_ERRORS
+        //   SET_FIELD_VALUE-------------------SET_LOW_PRIORITY_ERRORS
+        //       SET_ISVALIDATING-SET_ERRORS-SET_ISVALIDATING
+        //
+        // So we want to skip validation results if values are not the same
+        // anymore.
+        state.values !== msg.payload.values ||
+        isEqual(state.errors, msg.payload.errors)
+      ) {
+        return state;
+      }
+
+      return { ...state, errors: msg.payload.errors };
     case 'SET_STATUS':
       return { ...state, status: msg.payload };
     case 'SET_ISSUBMITTING':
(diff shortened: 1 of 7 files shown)


jaredpalmer:
I think this looks good.

jaredpalmer (changes requested):
See comment above.

packages/formik/src/Formik.tsx
```diff
@@ -72,6 +76,23 @@ function formikReducer<Values>(
...
+    case 'SET_LOW_PRIORITY_ERRORS':
+      if (
+        // Low priority validation after high priority validation, and this can
+        // stale validation results, e.g:
+        // SET_FIELD_VALUE-------------------SET_LOW_PRIORITY_ERRORS
+        //   SET_FIELD_VALUE-------------------SET_LOW_PRIORITY_ERRORS
+        //       SET_ISVALIDATING-SET_ERRORS-SET_ISVALIDATING
+        //
+        // So we want to skip validation results if values are not the same
+        // anymore.
```
umidbekkarimov, line 88: @jaredpalmer I did my best, please add suggestions if you have any.

packages/formik/src/Formik.tsx
```diff
@@ -72,6 +76,23 @@ function formikReducer<Values>(
...
+    case 'SET_LOW_PRIORITY_ERRORS':
+      if (
+        // Low priority validation can occur after high priority validation and
+        // this will create stale validation results, e.g:
+        // SET_FIELD_VALUE-------------------SET_LOW_PRIORITY_ERRORS
+        //   SET_FIELD_VALUE-------------------SET_LOW_PRIORITY_ERRORS
+        //       SET_ISVALIDATING-SET_ERRORS-SET_ISVALIDATING
+        //
+        // So we want to skip validation results if values are not the same
+        // anymore.
```
umidbekkarimov, line 88: @jaredpalmer I did my best, please add suggestions if you have any.

jaredpalmer (approved):

jaredpalmer:
Thanks for this!

Krasnopir:
After update to 2.2.2 the form became strange. On the authorization form, I select a value from the suggested autocomplete in the input field, and the password field is automatically filled in. But the submit button does not become active, because the login field is in focus and the form is not marked as valid. After clicking outside the form, the password field is highlighted in error, despite the fact that it is filled, and only then the errors disappear and the form becomes valid.
So I back to 2.2.0.

umidbekkarimov:
> After update to 2.2.2 the form became strange. On the authorization form, I select a value from the suggested autocomplete in the input field, and the password field is automatically filled in. But the submit button does not become active, because the login field is in focus and the form is not marked as valid. After clicking outside the form, the password field is highlighted in error, despite the fact that it is filled, and only then the errors disappear and the form becomes valid.
> So I back to 2.2.0.
> 
> ```
>   const validationSchema = yup.object().shape({
>     email: yup
>       .string()
>       .email(t`Введите корректный email`)
>       .required(t`Обязательное поле`),
>     password: yup.string().required(t`Обязательное поле`),
>   });
> 
>   const formikConfig = {
>     validationSchema,
>     validateOnMount: true,
>     initialValues: {
>       email: '',
>       password: '',
>     },
>     onSubmit: handleLogin,
>   };
> ```

Fixed in https://github.com/formium/formik/pull/2865

PR #2865, opened 2020-11-05
Fix low priority validation for browser password autofill.

umidbekkarimov:
Follow-up fix for the https://github.com/formium/formik/pull/2854#issuecomment-722529976

Some edge cases impossible to test without mocking `scheduler` internals, but since all the `scheduler` methods considered unstable I think the best way is to test inside the browser.


jaredpalmer (approved):

jaredpalmer:
Ahh shoot can you open another pr with the changeset for this one. It should be a patch. You just need to run `yarn changeset` and it will walk you through it

Issue #2860, opened 2020-11-04
<form> usage in examples is a security risk when JavaScript fails


WestonThayer:
## 📖 Documentation

Take the example on https://formik.org/docs/overview:

```jsx
<form onSubmit={handleSubmit}>
  <input name="email" ... />
  <input name="password" ... />
  <button type="submit" ... >Submit</button>
</form>
```

If you were to ship that code to production, and a client had JavaScript disabled, or there were unhandled JS exceptions that prevent `handleSubmit` from calling `e.preventDefault`, the `<form>` will fall back to it's default behavior, which is `method="get"`, which will navigate to `/sign-in?email=user%40example.com&password=Password!` (assuming the above code is on the `/sign-in` route), exposing the password as plain text in the URL bar.

This could leak the plain text password to any sort of analytics tracker.

I suggest adding `method="post"` to the `<form>`.

maddhruv:
Hey @WestonThayer we would really appreciate if you can raise a PR updating the example code 🤝

WestonThayer:
@maddhruv sure thing, see https://github.com/formium/formik/pull/2873.

PR #2873, opened 2020-11-06
docs: update form example security

WestonThayer:
Context in #2860.

Questions for reviewers:

- Does adding `method="post"` require explanation?
- Should the comment cite the GitHub issue?

If you agree with this approach, I'll look through the rest of the docs and also update the `Form` component to set `method="post"` by default.

### DIFF
--- a/docs/src/pages/docs/overview.md
+++ b/docs/src/pages/docs/overview.md
@@ -145,7 +145,10 @@ const Basic = () => (
         isSubmitting,
         /* and other goodies */
       }) => (
-        <form onSubmit={handleSubmit}>
+        <form
+          onSubmit={handleSubmit}
+          method="post" /* Don't leak password if JS disabled (#2860) */
+        >
           <input
             type="email"
             name="email"


johnrom:
I don't think we need to document specific security concerns or cite github issues in comments. If we need to describe in detail how to avoid security issues with Forms (that's not necessarily our place, as that info will not be nearly complete), we can create an FAQ area related to form security.

Instead, I think we might want to avoid using a username / password form as an example -- it sets a bad precedent to begin with. Authentication-related forms require many well-thought-out layers, and probably shouldn't be oversimplified like this.

WestonThayer:
I agree with that approach. Do you want me to update this PR to swap username/password for something benign like favorite food/favorite dessert?

A FAQ on form security would be great. Do you want to track in a GitHub issue?

johnrom:
I have no problem keeping some authoritative info related to form security in the docs somewhere. But we should link to something like OWASP and encourage users to read up fully on security because when things aren't as benign as favorite ice cream, form stuff is very serious. Even things that seem benign can be very serious, maybe your favorite ice cream is a security question on your credit card password reset form.

https://owasp.org/www-project-top-ten/

johnrom:
I like the idea of switching the form to a "benign" form, btw, anything is better than user / pass.

WestonThayer:
I think both some authoritative resources and formik-specific tips are good (unless there's an authoritative resource out that that talks about the dangers of writing a JS-only form + `<form>`'s default behavior, I couldn't find one).

docs/src/pages/docs/overview.md
```diff
@@ -145,7 +145,10 @@ const Basic = () => (
         isSubmitting,
         /* and other goodies */
       }) => (
-        <form onSubmit={handleSubmit}>
+        <form
+          onSubmit={handleSubmit}
+          method="post" /* Don't leak password if JS disabled (#2860) */
```
maddhruv, line 150: Can we remove this comment and reference to the issue

maddhruv (changes requested):
Please remove the comment

jaredpalmer:
Agree. Let’s remove the username password example

PR #2874, opened 2020-11-07
fix[FastField]: initial values in case of radio and checkbox

maddhruv:
Resolves - https://github.com/formium/formik/issues/2297

As FastField doesn't set initial values when the input type is radio or checkbox

### DIFF
--- a/.changeset/tough-dragons-dream.md
+++ b/.changeset/tough-dragons-dream.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
+
+fix FastField initial value when input type is radio or checkbox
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Can you add a changeset ?

maddhruv:
@jaredpalmer added the changelog

jaredpalmer (approved):

maddhruv:
@all-contributors please add maddhruv for bug code content doc example ideas projectManagement question review

johnrom:
@maddhruv @jaredpalmer this changeset migrated FastField from using formik.handleChange to field.onChange which could technically be a breaking change. Just a note in case we see FastField issues pop up.

jaredpalmer:
Oof you’re totally right. Shoot

royhadad:
how can I get this fix in my own project? is there a specific version I need to install? currently using 2.2.4

PR #2885, opened 2020-11-10
Low-priority validation state management

johnrom:
This manages low-priority validation via the reducer.

With this change, `validateField()` can cancel `validateFormWithLowPriority()`, resulting in partially validated forms. That's how it already works, since it abandons validation when the values don't match the props. However, if values previously _did_ match, this could cause an issue.

```js
const breakValidation = () => {
  validateFormWithLowPriority(values);

  // this matches the above, so Formik would previously apply lowpri validation after this completes. 
  // but now, validateField will unschedule the low priority validation.
  validateField(values.myValue);
}
```

theoretically, there may always be another low priority validation waiting for change, blur, or submit, but I need to think about it more. what we may need to do instead, is that if validateField is being called on the same values as were submitted to a currently schedule low-priority validation, validateField is cancelled instead.

here's the sandbox for the issue this solves: https://codesandbox.io/s/formik-codesandbox-template-forked-rozm4?file=/index.js

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -47,6 +47,14 @@ type FormikMessage<Values> =
   | { type: 'SET_TOUCHED'; payload: FormikTouched<Values> }
   | { type: 'SET_ERRORS'; payload: FormikErrors<Values> }
   | {
+      type: 'SCHEDULE_LOW_PRIORITY_VALIDATION';
+      payload: { values?: Values; };
+    }
+  | {
+      type: 'UNSCHEDULE_LOW_PRIORITY_VALIDATION';
+      payload: { values: Values; };
+    }
+  | {
       type: 'SET_LOW_PRIORITY_ERRORS';
       payload: { values: Values; errors: FormikErrors<Values> };
     }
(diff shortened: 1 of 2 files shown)


jaredpalmer:
Looks legit

jaredpalmer:
I guess my only wonder is if we should be even doing the scheduling in the first place. If we just got rid of async validation like in #1524 then I don't even think we need this?

johnrom:
I don't have an example proving it wrong, but I'm not committed to the idea of async validation. Using event callbacks for validation don't seem right to me, because how do you call all of them at once if you need to elsewhere? In my mind, the way we handle validation is not ideal, but in the opposite way. This is how I imagine validation in pseudocode:

```jsx
const Formik = () => {
  // this is validation as I imagine it in React
  // using refs that are only updated by user interactions or side effects, 
  // and NOT RENDERS. see #2828 
  const validate = React.useCallback(async () => {
    if (state.values !== refs.lastValidatedValues.current) {
      dispatchLastValidatedValues(state.values);
      const validationResult = await runAllValidationsEtc();
      // check if "future values" are equal to the values we validated or bail
      if (state.values === refs.values.current) {
        setErrors(validationResult.errors);
      }
    }
  }, state.values);
  // imaginary throttle function that only occurs once every Xms, 
  // but always calls the last call
  const lowPriorityValidationCallback = useThrottle(
    validate, 
    [formik.validateOnChangeDebounce ?? 250]
  );
  useEffect(
    // return void, not Promise
    () => { lowPriorityValidationCallback() }, 
    [formik.values, validateOnChange, lowPriorityValidationCallback]
  );
}
```

Then again, I don't think I've ever used async validation, so it's possible I'm overthinking it. If I did need it, I wouldn't want it to be in the onChange hook.

johnrom:
I wasn't able to find any instances where I've used async validations. This PR still solves an issue I have with current Formik release and it should probs be fixed without any API changes. I'm still trying to figure out if the `validateField()` cancellation would be an issue.

jaredpalmer:
I've been speking with Dan for the last hour or so, we are going to get rid of low priority validation in master since he doesn't believe it's working as expected. We instead going to have a CM mode branch that uses `startTransition` instead

johnrom:
After some good sleep I definitely rescind my useEffect version of low pri validation.

I think it works _kind of_ as expected. It prevents the validation itself from holding up the next render. This problem may be resolved in concurrent mode, but I think expecting users to adopt concurrent mode to get validations working correctly and performantly is a big ask.

Here's how I see this issue:

Low Pri validation is scheduled on React's internal scheduler which, to my knowledge, occurs before the next render when possible, or after the next render if the callback takes too long, triggering a second render. For renders, Errors and Touched are expected to be in sync. When Touched is changed, the validations must not be stale or there will be a flash when the scheduler isn't able to update validations between renders, because `touched` will be true but the stale error will remain in `errors`. To me, there are a few solutions to this.

1. Trigger touched and validation via the same `schedule`. This is questionable to me because users may want to know whether the field was touched independent of validation state.

2. Expand validation API. Validations return a ValidationResult with a shape like `{ errors: string, isUserRelevant: true }`. Or add that info anywhere else, like a sibling to `touched`. Whether a field is touched and whether a user expects feedback can be two different things, for example a user may only expect feedback after initial form submission attempt, which is a technique I use sometimes, and it prevents me from using touched properly.

  ```jsx
  const ErrorMessage = ({ name }) => {
    const field = useField(name);
    return field.validationResult.error && field.validationResult.isUserRelevant && <p>{validationResult.error}</p>
  }
  ```

  Now, `validateFormWithLowPriority()` can set 

  ```js
  formik.validationState = fields.each(field => 
    ({ error: validateField(field.name), isUserRelevant: getIn(touched, field.name) });
  ```

3. Check whether validation is scheduled, using the changes in this PR:

```
const ErrorMessage = ({ name }) => {
  const isTouched = getIn(formik.touched, name);
  const isValidationScheduled = getIn(formik.touched, name);
  const [isUserRelevant, setIsUserRelevant] = useState(isTouched);

  useEffect(() => {
    if (!isUserRelevant && isTouched && !isValidationScheduled) {
      setIsUserRelevant(true);
    }
  }, [isTouched, isValidationScheduled]);
}
```

johnrom:
Closing for now, but if we ever return to low priority state management my last comment has the solutions we need.

jaredpalmer:
Agreed. Boomarking this PR for future

johnrom:
@jaredpalmer `validationScheduledFor` can still be used for normal `isValidating` to prevent race conditions: #2898

validateField would need to wait until a scheduled validation has resolved to apply over the previous errors.

```
name: john                                                                                                                
age: 21                                  age: 22
    |                                         \ validateField:start      validateField:complete            | validateField:commit /
    \ validateForm:start                                                                  validateForm:end / 
```

Anytime after validateField:start, a change to validationScheduledFor would stop field-level validation from committing.

PR #2896, opened 2020-11-12
build: add install-types and fix missing dependencies

maddhruv:


jaredpalmer (commented):
Why is the postinstall script in Formik package needed?

maddhruv:
> Why is the postinstall script in Formik package needed?

@jaredpalmer so whenever any new packages is being installed, we don't miss the types for that

PR #2898, opened 2020-11-12
Remove unnecessary deep equality when dispatching errors

KhaledSakr:
This is already checked for in the reducer. It can also cause race conditions if a new validation result arrives before the dispatcher has had a chance to process the action.

### DIFF
--- a/.changeset/thick-penguins-smile.md
+++ b/.changeset/thick-penguins-smile.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
+
+Remove unnecessary deep equality when dispatching errors
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -325,9 +325,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       return runAllValidations(values).then(combinedErrors => {
         if (!!isMounted.current) {
           dispatch({ type: 'SET_ISVALIDATING', payload: false });
-          if (!isEqual(state.errors, combinedErrors)) {
-            dispatch({ type: 'SET_ERRORS', payload: combinedErrors });
-          }
+          dispatch({ type: 'SET_ERRORS', payload: combinedErrors });
         }
         return combinedErrors;
       });


johnrom:
To actually prevent race conditions, this pattern should probably be adopted:

```js
dispatch({ VALIDATION_STARTED, values: formikValues });

validateAll(formikValues)
  .then((errors) => {
    dispatch({ VALIDATION_COMPLETE, values: formikValues, errors });
  }).catch(() => {
    dispatch({ VALIDATION_FAILED, values: formikValues });
  });

const reducerLogic = (state, action) => {
  switch (action.message) {
    case VALIDATION_STARTED:
      return {
        ...state,
        validationScheduledFor: action.values,
      };
    case VALIDATION_COMPLETE:
      // strict equals, not isEqual
      if(state.validationScheduledFor === action.values) { 
        return {
          ...state,
          validationScheduledFor: undefined,
          errors: action.combinedErrors,
        };
      }
    case VALIDATION_FAILED:
      // strict equals, not isEqual
      if(state.validationScheduledFor === action.values) { 
        return {
          ...state,
          validationScheduledFor: undefined,
          errors: action.combinedErrors,
        };
      }
  }
}
```

KhaledSakr:
I agree that this is an overall better solution. Is it planned to be included in an upcoming release?

One comment I have is that we probably shouldn't catch errors on the `validateAll` promise. If one of my validation functions threw an error for any reason, I'd expect that error to propagate back to me instead of being swallowed by formik.

johnrom:
@KhaledSakr we should probably log those errors in development, but in Production client-side validation is generally viewed from the context of user experience. a user would expect the form to function normally and be able to submit even if the app failed to validate -- since one would be validating on the server side anyway.

I think from a programmatic standpoint, if you were to manually call `validate()`, I agree that we shouldn't catch those errors. But in terms of the validation triggered within Formik, I think it makes sense to swallow errors in production, and I think changing that would break a lot of peoples' expectations.

KhaledSakr:
Is this the current behavior though?

Check https://codesandbox.io/s/formik-example-forked-fccxw?file=/index.js

johnrom:
I guess not? I'll have to defer to someone else on the choices made here then.

jaredpalmer (dismissed):

PR #2902, opened 2020-11-13
fix: empty string conversion to undefined

maddhruv:
Resolves:
- https://github.com/formium/formik/issues/2528
- https://github.com/formium/formik/issues/2047
- https://github.com/formium/formik/issues/1520
- https://github.com/formium/formik/issues/1697


burtek (commented):
Do we even need `prepareDataForValidation` at this point? It looks like it does literally nothing at the moment.

johnrom:
it looks like it does a deep clone to pass the object to Yup immutably? isn't there a lodash func for that?

jaredpalmer:
Isn't this a massive breaking change?

johnrom:
definitely a breaking change. it looks like specifically for this, it will break min/max validations at the root level which currently accept no value.

https://github.com/formium/formik/issues/1520

This is how I understand it working from reading the PR (could be outdated):

If a validation schema like Yup.min(1).max(5) is entered, it will currently accept a value of ''.
If a validation schema like Yup.required().min(1).max(5) is entered, it will currently not accept a value of ''.
If a validation schema like Yup.min(1).max(5) is entered with the new changes, it will validate that '' is not between 1 and 5.

Can we get a unit test for this?

packages/formik/src/Formik.tsx
```diff
@@ -1092,13 +1092,13 @@ export function prepareDataForValidation<T extends FormikValues>(
           if (Array.isArray(value) === true || isPlainObject(value)) {
```
johnrom, line 1092: can we get rid of these `=== true` while we're in here?

maddhruv:
@jaredpalmer can you have a look

burtek:
@jaredpalmer ?

@maddhruv this needs update afaik (and response from Jared)

burtek:
ping again? c'mon...

tparvi:
We just ran into this issue (empty strings being passed as undefined to yup) so I'll share my opinion. We are using draft-js editor in our form:

```
type FormValues = {
  name: string;
  text: RawDraftContentState; // from draft-js
}

```
Because formik replaces **all** the empty strings **recursively** with `undefined` and passes that data to yup, it breaks the draft-js model that is received by my validation function:

```
function draftjsValidation(context: yup.TestContext, value: RawDraftContentState,
): yup.ValidationError | boolean {

  // draft-js convertToRaw fails because strings inside the value are undefined instead of empty
  const content = convertFromRaw(value);
```

In my opinion formik should pass the form values as is to yup but then again there are probably historical reasons why formik does this and changing it would be a breaking change.

felixmeziere:
Need this urgently please!

tparvi:
> Need this urgently please!

If you are using yup, you can run the validation yourself. Note the `yupToFormErrors `comes from formik.

```
<Formik validate={validate}>

function validate<T>(
  validationSchema: yup.SchemaOf<T>,
  values: T,
  context: ValidationContext
): FormikErrors<unknown> | undefined {
  try {
    validationSchema.validateSync(values, {
      abortEarly: false,
      context: context,
    });
  } catch (err) {
    return yupToFormErrors(err);
  }

  return undefined;
}
```

burtek:
#unstale
TBH this is a needed fix

codeclown:
Based on @tparvi's example, here's a ready-to-use wrapper function:

```ts
import { FormikErrors, yupToFormErrors } from "formik";
import yup from "yup";
export default function yupFormikWrapper(
  validator: yup.AnySchema
): (values: unknown) => Promise<FormikErrors<unknown>> {
  return async function validate(values) {
    try {
      await validator.validate(values, {
        abortEarly: false,
      });
    } catch (err) {
      return yupToFormErrors(err);
    }
    return undefined;
  };
}
```

Usage:

```ts
// Old
validationSchema={validator}
// New
validate={yupFormikWrapper(validator)}
```

PR #2903, opened 2020-11-13
Refactor v3 for alpha 

jaredpalmer:
- Renames `unstable_StrictField` to `FastField`.
- Deprecates `<FastField shouldUpdate>` prop with warning. If you need this functionality, use `useFormikContext()` and `useField()` in a custom component wrapped in `React.memo()` instead. 
- **Breaking**: `FastField` is no longer passed a `form` object in any render prop. If you still need to access the `form` object in the render prop use `Field` (which updates on every render).


jaredpalmer:
I am going to merge so that we can cut a release with these changes. More work needed to update tests

PR #2918, opened 2020-11-18
fix: unable to change checkbox with initial value 1 or 0.

abouolia:


johnrom:
Can this not be done with `parse`? 

I'd recommend opening an issue explaining the issue that you're having, which this PR solves, and linking to that issue. That way we can have more context related to this. To me, it seems like this would break checkboxes with an actual number value.

abouolia:
I've opened issue before the PR. 
https://github.com/formium/formik/issues/2917

"Can this not be done with parse?"
number as string is unlikely in many cases (my perspective), and unlikely boolean data comes from backend as string type, so there're three cases boolean, number (1, 0) or array.

johnrom:
That's my bad, I didn't realize the parse fn isn't releasing until v3. I believe your fix would change the outcome of this sandbox:

https://codesandbox.io/s/formik-checkbox-v212-forked-hsogq?file=/index.js

Initial value is 0, but user can select 0, 1, or both. Technically initial values is incorrect and it would be a developer error, because it should be [0] or ["0"], but the value could have been returned from an API and nobody noticed it changed to an array.

Because this has the potential to break production code, I'd recommend pushing this to a major release like v3.

PR #2931, opened 2020-11-25
Separate Formik State and API Surface

jawnrom:
Replaces #2828 

Creates a @formik/core package which contains the central Formik API `useFormikCore`, then creates a new Ref-loving version of Formik which uses that API. Isolating the Core API will allow us to experiment with switching between useContextSelector, useMutableSource, refs, etc. It will also let us unit test the individual bits of Formik instead of using integration tests as unit tests.

Up next: 

- Move tests to @formik/core using the new selectors.

### The new Core API `@formik/core`

The new Formik Core API is being used in the background by all Formik wrappers. This Formik Core API is meant to be separate from the actual React bindings (though it still has a dependency on React, for now), and be interchangeable between different implementations (useMutableSource, useContextSelector, etc). It takes four parameters, `getState`, `dispatch`, `formikProps` and `isMounted`, and from only those four parameters, it generates the following API surface.

```js
return {
  handleBlur,
  handleChange,
  handleReset,
  handleSubmit,
  resetForm,
  setErrors,
  setFormikState,
  setFieldTouched,
  setFieldValue,
  setFieldError,
  setStatus,
  setSubmitting,
  setTouched,
  setValues,
  submitForm,
  validateFormWithLowPriority,
  validateFormWithHighPriority,
  validateField,
  unregisterField,
  registerField,
  getFieldProps,
  getFieldMeta,
  getFieldHelpers,
};
```

Anywhere the API needs to get a value from Formik's state, it calls getState, and anytime it needs to dispatch an update, it calls dispatch. This is a redux-like inversion of control.

### The new implementation framework

Any implementation of the Formik API will expose the FormikCore API as well as the following helpers:

```js
return {
  ...formikCoreApi,
  getState,
  validateForm, // generally, formikCoreApi.validateFormWithHighPriority
  // dispatch, // not sure if we should expose ??
  // addFormEffect, // this is specific to this refs implementation
}
```

I've added a new Formik implementation, `@formik/reducer-refs` which uses a redux-like state machine to dispatch changes to its internal state (via dispatch) to React's UI state (via useState). As of right now, that implementation isn't complete, BUT! it provides a great example of how I will modify the original `formik` to use the new `@formik/core` API. Basically, Formik would just pass getState() as a callback `() => state` from useReducer, and the dispatch as is.

The implementation now exposes a stable `useFormikApi` which doesn't trigger UI updates (or it won't, when I'm done!), and `useFormikContext` decorates that API with state which will trigger UI updates. I hope eventually to deprecate useFormikContext in favor of `const [state, api] = useFormikState();` because it is more React-like.

> But how does the new Reduxy API work?!

Great question! As described #2846 , the new reduxy API works like this:

#### Form effects

```ts
const useFormikContext = () => {
  const formikApi = useFormikApi<Values>();
  const [formikState, setFormikState] = React.useState(formikApi.getState());
  
  React.useEffect(() => {
    // every time the reducer is updated and committed, the next useEffect will pass that state to each listener
    return formikApi.addFormEffect(setFormikState);
  }, []);

  return {
    ...formikApi,
    ...formikState,
  };
};
```

#### Field effects

```ts
const useField = <Val>(props: FieldHookConfig<Val>) => {
  const formik = useFormikApi();
  const fieldMetaRef = React.useRef(getFieldMeta<Val>(fieldName));
  const [fieldMeta, setFieldMeta] = React.useState(fieldMetaRef.current);

  const maybeUpdateFieldMeta = React.useCallback<FormEffect<Values>>((formikState) => {
    const fieldMeta = selectGetFieldMeta(() => formikState /* hint: getState */)(name);

    if (!isEqual(fieldMeta , fieldMetaRef.current)) {
      fieldMetaRef.current = meta;
      setFieldMeta(meta);
    }
  }, [name, /* never update >> */ setFieldMeta, fieldMetaRef]);

  React.useEffect(() => {
    // every time the reducer is updated and committed, the next useEffect will pass that state to us
    // and we'll slice it and update our field values
    return formik.addFormEffect(maybeUpdateFieldMeta);
  }, [maybeUpdateFieldMeta]);

  return [formik.getFieldProps(props), fieldMeta, formik.getFieldHelpers(props)];
}
```

The useField implementation is one example of how we can optimize things and also let users optimize things their own queries to Formik's parameters. For example, if a user only needs to re-render if `isValid` changes, they can do this. Consequently, this would be the exact definition of `useIsValid` from the new useContextSelector-based API.

```js
const [isValid, setIsValid] = useState(formikApi.getState().isValid);
const updateIsValid = React.useCallback((formikState) => setIsValid(formikState.isValid),  []);

React.useEffect(() => {
  return addFormEffect(updateIsValid);
}, [addFormEffect, updateIsValid]);
```

We will also provide the full set of hooks added by v3 in the useContextSelector changes, which will use this method.


johnrom:
Updated the description to include a heck of a lot of details on how the Core API will work, and how my experimental refs implementation will work. 

@jaredpalmer

johnrom:
I just optimized a few things and now the performance is close to v3. @jaredpalmer how did you get timing metrics for testing performance?

In my last set of changes, I rewired the /app so that it compiles the Formik packages while running, and breakpoints etc are accessible via a vs code Chrome launch task. No more yarn link. I didn't touch the tsdx build process.

To try it out, pull this PR, run `npm start` and navigate to localhost:3000. Or run the vscode "launch chrome" profile and set some breakpoints to test.

I added 2 pages, one that tests 500 inputs using v3 and one that tests 500 inputs using @formik/reducer-refs. I couldn't tell which was faster by feeling alone.

johnrom:
Switched `/app` to integrate with Lerna Workspace.
Set up unit test environment for vscode.
Set up dev environment for vscode.
Wrote example unit test for `@formik/core`.

johnrom:
The current issue with is a TypeScript inference issue. It seems like for some reason there are two `@types/react` and that one is used for inference but isn't locally installed, resulting in the error:

```
semantic error TS2742: The inferred type of 'Form' cannot be named without a reference to 'hoist-non-react-statics/node_modules/@types/react'. This is likely not portable. A type annotation is necessary.
```

Seems like https://github.com/microsoft/TypeScript/issues/29808

johnrom:
Resolved typescript issue with `resolutions` in package.json.

johnrom:
Simplified `useFormikStateSlice` per https://github.com/formium/formik/issues/2433#issuecomment-759844122

And fixed all the Jest Tests. Now just need to fix the cypress tests.

johnrom:
OMG a fully working build! 🥇

johnrom:
Here's a comparison of the underlying features of each implementation:

| Implementation | Render-optimized | Access to "future" state | Batching | Versioning | Notes |
| v2 | ❌| ❌| ✅ | ✅ | Use of Context provides automatic batching and versioning, preventing tearing. |
| useContextSelector | ✅ | ❌| ✅ | ✅ | useContextSelector uses both unstable_runWithPriority and unstable_batchedUpdates to achieve batching. getState() is render dependent and cannot be used to get future state until the next render is committed |
| reducer-refs | ✅ | ✅| ❌ | ❌ | getState() provides ref access to the underlying current state, but does not currently use batching or versioning. May risk tearing and extra renders without further optimization |

johnrom:
RE: https://github.com/formium/formik/issues/2433#issuecomment-761273638 @dantman

> I'm not concerned about tearing in legacy mode.

I will be advocating using a different version of Formik for concurrent mode, and with the way things are looking the underlying implementation will be with useMutableSource. Jared and I discussed using Feature Flags in webpack or some other method of opting into concurrent mode. I like the feature flag method because it doesn't require understanding the underlying implementation or changing code, but the API is yet to be finalized. My PR here isn't focused on those details, but aimed at identifying a common API we can use in both sync and concurrent mode, one which in the case of `useFormStateSlice` discussed on the issue above, would use different underlying implementations of subscriptions to that state depending on sync or concurrent.

I am actively advocating avoiding "drinking the kool-aid" like we did for v2 where React came out with Context and we said "React says it's OK so let's go all in!" and it resulted in a less performant Formik experience in v2. I don't want to over-prepare for concurrent mode at the cost of today's dev experience out of a sheer trust in React to implement it within a reasonable period of time, but I do want to be prepared enough that if they launched it tomorrow we'd be able to "fall forward" relatively easily.

To this end, I've split the bulk API of Formik into one package, and created two implementations (Jared's useContextSelector and my "reducer-refs", for lack of a better name, implementation). This change supports the creation of a useMutableSource implementation as well. There are some kinks to work out, but I think this is enough for a proof of concept.

> Optimizations gained by context splitting however use an API currently available in React 17, but have no tearing issues. i.e. The same optimized code works in both stable react and experimental concurrent mode, but doesn't introduce tearing between components in concurrent mode like subscriptions do.

But these optimizations have a trade-off -- we have to maintain all of those contexts, each of those contexts has its own render pipeline, and partial use of straight context and partial use of subscriptions, those are the technical ones. However, the biggest one I see is one of dev experience -- a developer reading through their Dev Tools Components through a list of 20 Contexts per Formik instance to debug an input will be extremely frustrated by the development experience of nested Contexts.

> Do you have any reason to anticipate that a setValue and a setTouched made at the same time in the same batch would result in a component receiving one but not the other? To my understanding that's not supposed to happen.

Batches are currently only automatically used in React-based event handlers and other special React functions, otherwise they must be wrapped in `unstable_batchedUpdates`. So it is my understanding that if you did something like:

```
// Formik users love async code like this
setTimeout(() => {
  formik.setIsSubmitting(true);
  formik.setValues(myNewValues);
});
```

Then React could commit the result of isSubmitting (the reducer `state` that results), and `setValues` at separate increments. Normally I think it would be so fast a callback as to hit in the same render, but I would be concerned that if one value is received via `Context` and another is a subscription (the result of an effect on the resulting `FormikState`), this would increase the chance that they didn't hit during the same render. I don't have anything but a gut feeling to back that up, but would love proof for or against.

------

I've been thinking all weekend however about what you said, and I agree that running memoizations for every single subscription that is added is a little bit overboard. That's why I'm going to take the following steps with this PR and update Formik's state in a more performant way.

- For the bulk of Formik's users, we're going to recommend using micro-hooks, like useIsSubmitting, useField, etc.
- As an advanced topic, I still think we should expose the underlying State Slice mechanism, but it will be a little bit more complicated to use.
- Formik's hooks will use this State Slice mechanism internally in a highly optimized fashion.

Formik's context will keep track of listeners on a per-update-callback basis, in a shape that looks like this:

```
// note, this will start off an empty object with no updates, and each call to useFormikX will add listeners here, and the first addition of each one will create the prevSliceRef.
const formikListeners = {
  // all references to `useIsSubmitting` are here. this is a VERY fast function, using ===
  memoizedUseIsSubmittingUpdaterSymbol: {
    selector: selectIsSubmitting,
    updater: updateIfNotReferenceEquals,
    listeners: [],
    prevSliceRef: {
      current: selectIsSubmitting(state),
    }
  },
  // all listeners for  `useField("myField")` are here. All references to `useField("myOtherField")` would be in another context.
  // this is relatively slow, but as optimized as possible
  memoizedUseFieldUpdaterSymbol: {
    updater: updateIfNotDeepEquals,
    selector: selectGetFieldMeta,
    listeners: [],
    prevSliceRef: {
      current: selectIsSubmitting(state),
    }
  }
}
```

Now instead of a flat list of listeners that apply to every update, We can first diff state, and if the state differs for a given set of listeners, we can apply them all at once. This is the theoretical equivalent of how nested Contexts would work, but centralized within Formik.

I'm gonna prototype this up and see how it works. I'm going to create a new page in the dev app that is 500 inputs of the SAME FIELD. It should currently be pretty slow, and implementing this should make it lightning.

johnrom:
Created a dev app page with 500 of the same input. It's pretty slow.

dantman:
> However, the biggest one I see is one of dev experience -- when debugging in Dev Tools through a list of 20 Contexts per Formik instance I think nested Contexts would be borderline unusable.

This should NEVER be a factor considered. Use as many contexts as makes sense. The react team has made it clear you shouldn't be minimizing contexts expecting to optimize things in a different way if you find that more contexts is the more optimal. React's DevTools have an option for hiding contexts precisely because of this.

> * For the bulk of Formik's users, we're going to recommend using micro-hooks, like useIsSubmitting, useField, etc.

👍

> * Formik's hooks will use this State Slice mechanism internally in a highly optimized fashion.

Was a bit strange to wrap my head around the example. But looks good. Separate subscription listener systems (which are effectively just an array of functions to call) definitely seem cheaper than running object memoization on everything.

Instead of having `useFormikX` dynamically append a listener. How about creating something based on the `create*` pattern React uses. e.g. `createFormikListener`. This could offer guarantees that all listeners of the same type use the same config and there are no name collisions. Perhaps it could also optimize the testing of whether a type of listener already exists. e.g. Maybe it could reduce the formikListeners list to a Set (actually, maybe Map) that you only need to run a super fast .has on to know if the listener is already available on that formik instance.

e.g. Something like:
```js
const isSubmittingSelector = createFormikSelector({
	selector: selectIsSubmitting,
	updater: updateIfNotReferenceEquals,
});

function useIsSubmitting() {
	return useFormikState(isSubmittingSelector)
}

// somewhere in <Formik>, probably an addFormikSelector
if (!formikSelectorListeners.has(selector)) formikSelectorListeners.set(selector, new Set());
const listeners = listeners.get(selector);

listeners.add(listenerHandler)
const unsubscribe = () => listeners.delete(listenerHandler);
```

> ```js
> const updateIfNotReferenceEquals = (selector, listeners, prevSliceRef, newState) => {
>   const newSlice = selector(newState);
>   const prevSlice = prevSliceRef.current;
>   if (prevSlice !== newSlice) {
>     prevSliceRef.current = newSlice;
>     listeners.forEach(listener => listener(newSlice));
>   }
> }
> ```

I'm not sure the updater should be responsible for actually executing listener handlers. What if a formik listener from the author's perspective was made up of just a selector function to get the slice and a fast equality function to test if listeners should be called:

```js
const selectIsSubmitting = state => state.isSubmitting;
const referenceEquality = (prevSlice, nextSlice) => prevSlice === nextSlice;

const isSubmittingSelector = createFormikSelector({
	selector: selectIsSubmitting,
	equals: referenceEquality,
});
```

Then the majority of that update function isn't needed since it's boilerplate that applies to all formik listeners and can just be in formik somewhere as this:

```js
// S=The createFormikSelector result,
// or this could be part of createFormikSelector and {prevSliceRef, newState} are passed in
const nextSlice = S.selector(newState);
const prevSlice = prevSliceRef.current;
if (!S.equality(prevSlice, nextSlice)) {
  prevSliceRef.current = nextSlice;
  listeners.forEach(listener => listener(nextSlice));
}
```

johnrom:
> This should NEVER be a factor considered. Use as many contexts as makes sense. The react team has made it clear you shouldn't be minimizing contexts expecting to optimize things in a different way if you find that more contexts is the more optimal.

But what if a dev needs to debug Formik's context? Still seems very weird to me. I haven't read anything from the React team specifically about using dozens of contexts but if it's not in the docs I probably haven't read it.

I still have a concern about subscriptions + raw contexts being mixed together, but I'd be really interested to compare performance between these two methods.

> How about creating something based on the create* pattern React uses. e.g. createFormikListener.

YES. I really like the idea of `createFormikSelector`. The reason I split args into its own separate dependency hook is because when you using something like, the following, you're not actually using a constant value function, but a dynamic one. 

```ts
const [field] = useField("myField");
```

Here, the underlying mechanism you're actually using would be something like:

```
// selectGetFieldMeta is a global constant, but the return value isn't
const selector = selectGetFieldMeta(name);
```

So we'd have to separate args in order to create a single instance across components. There are a few methods I have in mind:

### 1.

```ts
// the identity is now selector+[...args]
const selector = api.createSelector(selectGetFieldMeta, [name]);
// both args must be optimized
const fieldMeta = useFormikState(selector, _.equals);
```

### 2.

```ts
// the identity is now selector+comparer+[...args]
const selector = api.createSelector({
  selector: selectGetFieldMeta,
  comparer: _.equals,
  args: [name],
});
const fieldMeta = useFormikState(selector);
```

### 3.

```ts
// the identity is now selector+comparer, but selector can be optimized separately
const subscriber = api.createSubscriber({
  // selector can still be cleaned up because it is scoped to API.
  // additionally, this could just be a constant function like selectIsSubmitting
  selector: api.createSelector(selectGetFieldMeta, [name]),
  comparer: _.equals,
});
const fieldMeta = useFormikState(subscriber);
```

I guess the final thing I'm concerned about is that we've moved some implementation details out from behind `useFormikStateSlice()`, which are currently nicely tucked away in this PR.

johnrom:
Looks like they basically document optimizing selectors here: https://github.com/reactjs/rfcs/blob/master/text/0147-use-mutable-source.md#redux-stores

They don't really give a concrete example across Components, but inferring from their examples, it looks like createSelector will still be useful for establishing identity. It also doesn't seem like there is a way to configure whether to re-render when the result of `getSnapshot` has a certain type of equality, so we'll have to thoroughly test performance when moving to `useMutableSource`. getSnapshot doesn't get passed a previous value to check against, and I don't believe refs will be safe in that context.

johnrom:
It was challenging to get TypeScript to play nicely with this system, BUT! Alas, it is done. I've simplified the Subscriptions API in to more manageable chunks, using the useSubscriptions API.

The new API is like so:

```ts
import { isEqual } from 'lodash';

const MySubscriber = () => {
  const { createSubscriber, createSelector } = useFormikApi<FormValues>();

  /**
   * Create a subscriber. Memoizing it is the most important thing you can do.
   * I think if I can memoize the 
   */
  const subscriber = React.useMemo(
    () =>
      createSubscriber(
        createSelector(selectFieldMetaByName, [fieldName]),
        isEqual
      ),
    [createSelector, createSubscriber, fieldName]
  );

  // GEE it was hard to get `fieldMeta` automatically typed.
  const fieldMeta = useFormikStateSubscription(subscriber);
}
```

I think the footprint can get even smaller, but I'm pretty tired of battling typescript at this point, lol. Similar to what is discussed, batched updates just do this:

```
    unstable_batchedUpdates(() => {
      subscriptionsRef.current.forEach(subscription => {
        const prevState = subscription.prevStateRef.current;
        const newState = subscription.selector(state);

        if (!subscription.subscriber.comparer(prevState, newState)) {
          subscription.prevStateRef.current = newState;
          subscription.listeners.forEach(listener => listener(newState));
        }
      });
    });
```

johnrom:
The other alternative I can think of to find existing Subscriptions is the following. Not sure exactly how performance would differ without testing.

```js
// as exists in the PR
const subscriptions: SubscriptionsRef = useRef([]);
// complex mapping object to keep track of what selectors / comparers / args map to which subscriptions
const subscriptionMap = new Map();
subscriptionMap.push(
    selectIsSubmitting, // selector
    new Map(
      Object.is, // comparer
      {
        subscription: subscriptions[0] // for example
        nextArgs: undefined,
      }
   )
);
subscriptionMap.push(
    selectFieldMetaByName, // selector
    new Map(
      _.isEqual, // comparer
      {
        subscription: undefined, // there should be no version of this without args
        nextArg: new Map(
          "firstName", // first arg
          {
            subscription: subscriptions[1],
            nextArg: undefined, // unless there are more args passed in createSelector
          }
        )
      }
   )
);
```

johnrom:
I've simplified the API to look like this:

```
useFormikState(selectIsSubmitting);
useFormikState([selectFieldMetaByName, "firstName"], compareFieldMeta);
```

304s (dismissed):

johnrom:
The reducer-refs API enables easy optional subscriptions, such as the `<Form />` subscribing to `values` only when `onChange={values => dependents(values)}` is passed to it.

johnrom:
closed in favor of #3089

Issue #2751, opened 2020-09-17
Get validationSchema or required fiedls list from formik


SIN3d73:
## ❓Question

Is there any option? It would be very helpful for required indicator  `*` .
According to [https://github.com/formium/formik/issues/1241](https://github.com/formium/formik/issues/1241) there is no way.

PR #2933, opened 2020-11-28
Add missing validationSchema to Formik context

JoseLion:
Hi @jaredpalmer!

I noticed the `validationSchema` was missing when using `useFormikContext`. The types say it's there, but it always returns `undefined`. I took a look, and it seems it was missing from the context value definition, so I decided to open this small PR to lend a hand with this.

Hope this helps 🙂


johnrom:
This is ok by me especially since it exists in the types already. The only issue I see with it is that some users may create a new Yup object on each render. and if that's the case, this prop will change on every render, and cause more updates. Not sure how much of a problem that is. I don't use Yup, so I'll defer to someone else whether that's an issue or not.

johnrom:
related https://github.com/formium/formik/issues/2751

JoseLion:
@johnrom thank you for your response 🙂 I think the most common use-case will be exactly what is mentioned in #2751. As I think about it, it seems unlikely that a new Yup object is created on each render, but it could happen. I hope it's not much of a problem because that is the only way of getting validation info into abstracted field components. 😅

JoseLion:
@jaredpalmer any news on this one? It's not super critical, but I think it would be great to have this 🙂 

Cheers!!

noisyscanner:
Any movement on this? Would be very useful to us

JoseLion:
I just updated the branch, so this is ready to merge if there're no concerns 🙂

johnrom:
I don't think we should pass through this or the `validate` callback through the default Formik React context because of the high chance it has not been optimized. If it isn't optimized, it will trigger re-renders in every single component which uses Context to access Formik.

I'd be open to passing these values via a secondary Context which contains only Formik Configuration:

```jsx
const Formik = () => {
  // formikApi is completely memoized
  // formikConfig is more volatile based on usage
  const [formik, config] = useFormik(...formikProps);

  return <FormikProvider value={formik}>
    <FormikConfigProvider value={config}>
      {/* children */}
    </FormikConfigProvider>
  </FormikProvider>
}

const useFormikConfig = () => useContext(FormikConfig.Consumer);
```

@jaredpalmer if this is something we want to do, I can make these changes to my v3 branch

JoseLion:
I updated this branch just in case. Thanks again @johnrom for looking into this 🙂

noisyscanner:
@github-actions plz no

johnrom:
FYI this change is available in my v3 PR #3231

JoseLion:
@johnrom I updated this branch just in case. However, it's great to know this is coming in v3, is there an ETA for v3 yet? 🙂

PR #2938, opened 2020-12-01
fix: remove lodash-es

maddhruv:


TrySound:
Maybe better completely get rid from lodash?

TrySound:
Btw does it play with tsdx?

johnrom:
To test the impact of a change like this we need to test the impact of file size on the following for both a CJS and ESM project:

- impact on a project that only imports formik and react
- impact on a project that imports formik, react and `lodash`
- impact on a project that imports formik, react, and `lodash-es`
- impact on a project that imports formik, react, and lodash and uses the webpack lodash plugin
- impact on a project that imports formik, react, and lodash-es and uses the webpack lodash plugin
- impact on a project that imports formik, react, and `lodash` and uses the alias technique below
- impact on a project that imports formik, react, and `lodash-es` and uses the alias technique in reverse:

```
// webpack.config.js or whatever
webpackConfig.resolve.alias = {
  lodash: 'lodash-es',
};
```

https://github.com/formium/formik/issues/2602#issuecomment-658296674

My theory is that we go ahead and use `lodash` because it is the standard and users are far more likely to have imported it than lodash-es. However, the caveat is that we should all of our imports within Formik should use the following format:

```
import join from 'lodash/join';
```

I think if we combine those two changes, and instruct users to use the webpack lodash plugin, that may be the silver bullet solution to solving this issue in userland for v3. But we'll want to test and make sure webpack lodash plugin can correctly alias `lodash/join` type imports first.

Evidence that lodash is more popular than lodash-es: https://www.npmtrends.com/lodash-vs-lodash-es

johnrom:
(we also may get rid of lodash completely depending on the API of v3)

maddhruv:
@johnrom I was just trying formik in a webpack project and was seeing both lodash being included in the final bundle. Not much impact on bundle size, as were named imported but no sense of having duplicate codes.

johnrom:
@maddhruv that is one of the many different package configurations I listed, all of which will have different effects.

if your project used `lodash-es` (which some certainly do) you would received duplicate packages with this PR, for example

I still agree with this change, though as mentioned above I think we should go above and beyond to limit the scope of our imports since `lodash` isn't tree-shakeable (I don't think?) and make sure that webpack lodash plugin can optimize them to create the lowest total package size across all common project types. This will also affect people's packages in a big way so this is likely something we should punt to v3.

jaredpalmer:
TSDX intelligently bundles lodash and treeshakes properly. 

https://tsdx.io/optimization#using-lodash

However, I would like to get rid of lodash altogether. I haven't yet found an equivalent of `toPath` that supports brackets.

TrySound:
@jaredpalmer I mean will tsdx continue to work when lodash-es is removed?

jaredpalmer:
Haven't tried, my guess is it will not.

jaredpalmer:
AFAIK implementation is still your original one from back when formik handled its own build.

palfrey:
lodash-es hasn't had a release since 4.17.15 and so is subject to a bunch of vulnerabilities v.s. lodash 4.17.20 which has none of them. See https://snyk.io/test/npm/lodash/4.17.15

TrySound:
Someday they may lost an access to `lodash` package as well 🤷‍♂️

johnrom:
I don't think anyone is advocating that we keep lodash-es for no reason, but I do recommend a bunch of mitigations above that need to be considered before merging this, and this should probably be shipped with v3 and not v2 for those who have already solved the duplication in their projects another way, unless we can prove in all situations I listed above this doesn't have a negative impact.

jaredpalmer:
I am open to removing lodash in v3 but we need to do it gracefully with warnings in v2 if there are any backwards incompatible changes (such as bracket path references)

palfrey:
> lodash-es hasn't had a release since 4.17.15 and so is subject to a bunch of vulnerabilities v.s. lodash 4.17.20 which has none of them. See https://snyk.io/test/npm/lodash/4.17.15

This bit looks fixed now. See https://github.com/lodash/lodash/issues/4879#issuecomment-748571123

l0gicgate:
> Maybe better completely get rid from lodash?

@TrySound I'm 💯 in favor of this. 75kb (ungzipped) is a lot..

lodash-es

PR #2939, opened 2020-12-01
Add useSetFieldValue and useSetFieldTouched hooks (v3)

seropaski:
Hi everyone!

In my work I fint out that I often repeat some selectors like `useFormikContextSelector(ctx => ctx.setFieldValue);`. So I've added hooks for `setFieldValue` and `setFieldTouched`. Seems it may be useful for other developers.

### DIFF
--- a/.changeset/afraid-buses-kiss.md
+++ b/.changeset/afraid-buses-kiss.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
+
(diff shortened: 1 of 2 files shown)


jaredpalmer (changes requested):
Can you add a changeset with `yarn changeset`?

seropaski:
Done

jaredpalmer:
Thanks.

PR #2955, opened 2020-12-09
fix(docs): fix link to docs in console warning

jkbktl:
This PR fixes link to Docs in console warning which is showed when developer forgets to pass identifier. There are two anchor links, both tested, both works well with new link.


jaredpalmer (approved):
Thanks!

PR #2980, opened 2020-12-26
chore(docs): Change functional component type in Typescript example

Freddixx:
Hi all 👋 ,

I believe `React.FC` is better avoided as it doesn't offer too many benefits (see https://github.com/facebook/create-react-app/pull/8177). If you are required to explicitly declare a type because your project settings urge you to do so, then  `ReactElement` is the safe shot imo.

Note that this is an opinionated PR, please let me know what you think. 

Happy holidays 🙂 🎄

### DIFF
--- a/docs/src/pages/docs/guides/typescript.md
+++ b/docs/src/pages/docs/guides/typescript.md
@@ -26,7 +26,7 @@ interface MyFormValues {
   firstName: string;
 }
 
-export const MyApp: React.ReactElement = () => {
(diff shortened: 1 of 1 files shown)


johnrom:
We can just remove the return type annotation as it will be inferred by TypeScript. We may have added it here either for demonstrative purposes or to test to make sure TSX was working at some point, but it isn't necessary and is probably confusing if anything. I don't think anyone wants to ponder the difference between JSX.Element React.ReactElement while reading our examples :)

PR #2981, opened 2020-12-26
Allow explicitly setting action to empty string

KATT:
If someone actually wants to do a standard http post to the same page.


johnrom (approved):
Makes sense to me.

Issue #2993, opened 2020-12-30
Top level field validation rule violations on a <FieldArray> return invalid error messages


elderfo:
## Bug report

Removing item from `FieldArray` when there is a top level validation rule (such as min/max), the error message is missing characters.

When I tested multiple versions, this appears to show up for the first time in 2.0.4 and still exists in the current version 2.2.6.

### Current Behavior

This can be recreated fairly simply by using a schema with an array using `min()` or `max()`

Here's sample code and (repro sandbox)[https://codesandbox.io/s/formik-codesandbox-template-forked-3rbjh?file=/index.js]:
```js
export const schema = yup.object({
  fieldIds: yup.array(yup.string()).required().min(5)
});

const FormikForm = () => {
  const [, setCtr] = useState(0);

  return (
    <Formik
      onSubmit={console.log}
      validationSchema={schema}
      validateOnMount
      initialValues={{
        fieldIds: []
      }}
    >
      {(formik) => (
        <FieldArray name="fieldIds">
          {({ push, remove }) => (
            <div>
              <h1>Formik</h1>
              <div>Error: {formik.errors.fieldIds}</div>
              <button
                onClick={() => {
                  setCtr((c) => {
                    const newCtr = c + 1;
                    push(newCtr);
                    return newCtr;
                  });
                }}
              >
                Add
              </button>
              {formik.values.fieldIds.map((f, idx) => (
                <div key={f}>
                  {f} <button onClick={() => remove(idx)}>Delete</button>
                </div>
              ))}
              <DisplayFormikState {...formik} />
            </div>
          )}
        </FieldArray>
      )}
    </Formik>
  );
};
```

To repro: 
1. Click Add 5 times
2. Click Delete multiple times, make note of the error message after each delete

_Note:_ In the same sandbox, I've included a straight yup version where validation works as expected, just to rule out the issue on it's end

### Expected behavior

Validation errors should be returned properly as a string for top level errors, or an array for nested value errors. The (documentation)[https://formik.org/docs/api/fieldarray#fieldarray-validation-gotchas] calls out this expected behavior

### Reproducible example

See Current Behavior

### Suggested solution(s)

Looking at the code, top level error messages are being converted to arrays and then having the deleted index removed. FieldArray errors that are strings should be kept as strings.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.2.6 |
| React            | 16.14.0 |
| TypeScript       | 4.1.2 |
[...]

PR #2994, opened 2020-12-31
fix: errors in FieldArray being returned as arrays

elderfo:
Fixes: #2993


ryangiglio-remesh:
Is there a reason this fix was never merged? I'm having the same issue and I see a few other reports of it as well.

jaredpalmer (dismissed):

PR #2997, opened 2021-01-06
Fix setError typing

JamieS1211:
Has been brought up before however has since become stale - https://github.com/formium/formik/pull/2215

### DIFF
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -299,7 +299,7 @@ export interface FieldHelperProps<Value> {
   /** Set the field's touched value */
   setTouched: (value: boolean, shouldValidate?: boolean) => void;
   /** Set the field's error value */
-  setError: (value: Value) => void;
+  setError: (value: string) => void;
 }
 
 /** Field input value, name, and event handlers */


packages/formik/src/types.tsx
```diff
@@ -299,7 +299,7 @@ export interface FieldHelperProps<Value> {
   /** Set the field's touched value */
   setTouched: (value: boolean, shouldValidate?: boolean) => void;
   /** Set the field's error value */
-  setError: (value: Value) => void;
+  setError: (value: string) => void;
```
johnrom, line 302: This only covers one case, since `values[name]` can be a nested value requiring an object FormikErrors signature, or an array if Value is an array. It'd need to be something like the following, with tests added to the unit tests for each case.

```ts
type FieldError<Value> = 
  | string
  | Value extends Record<string, any> ? 
    FormikErrors<Value> :
    Value extends any[] ?
      string[]
      : never;

setError: (value: FieldError<Value>) => void;
```

Tests:

```
Value is string && error is string
Value is number && error is string
Value is object && error is string
Value is AnyShape && error is FormikErrors<AnyShape>
Value is array && error is string
Value is array && error is string[]
```

Note that with TS template literals it might actually make more sense to make the type:

```ts
type SetErrorsFn<Values, Path> = (error: PathMatch<FormikErrors<Values>, Path>) => void;
```

Or something.

johnrom (changes requested):
Needs to cover more cases or maybe rethink implementation with Template Literals.

packages/formik/src/types.tsx
johnrom: Also FormikErrors might already have some of that functionality built in? So it could be as easy as `string | FormikErrors<Value>`.

PR #3015, opened 2021-01-27
Fix examples in docs

jaredpalmer:


jaredpalmer:
It's getting late here. This is slightly janky because of hydration, but it works and so do the examples and iframes. Can clean up in the morning.

PR #3030, opened 2021-01-31
Docs: add a small note about item keys in <FieldArray />

gejimayu:
Resolution to this request https://github.com/formium/formik/issues/3007

Screenshot 2021-01-31 at 7 33 52 PM


johnrom:
I think it would be a bit more clear to use `friend[index].id` in the Object example, and to explain in the paragraph above it why that is a better choice.

https://github.com/formium/formik/pull/3030/files#diff-c22f71f4f57b6834b24b2b43a21dd4c991a6c07de270233e26cacedf2e91387aR89

gejimayu:
okay i agree, done

Screenshot 2021-02-01 at 11 32 12 AM

PR #3039, opened 2021-02-09
Update Formik.tsx

mahmut-gundogdu:
when the value is not string and the value is not React ChangeEvent event. it throw error. it fixed.

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -599,7 +599,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       let parsed;
       // If the first argument is not a string though, it has to be a synthetic React Event (or a fake one),
       // so we handle like we would a normal HTML change event.
-      if (!isString(eventOrTextValue)) {
+      if (!isString(eventOrTextValue) && (eventOrTextValue.target || eventOrTextValue.currentTarget) ) {
         // If we can, persist the event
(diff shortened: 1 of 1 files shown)


johnrom:
If you want to update the value directly, use `handleChange('name')(value)`.

JulianKingman:
@johnrom as of v2 using `handleChange('name')(value)`, value can only be a string or event, which it seems like this PR is for?

mahmut-gundogdu:
I think  value could be string ,  date or number (even object) , but now 
`handleChange('name')(value)` 
only works string or event. This pr changed this. (actually it is a a suggestion)  (it might be like that)

johnrom:
Ah I see. This doesn't actually fix the underlying issue, though. `handleChange` does indeed only support string or a ChangeEvent, and other values like integer, objects, arrays, dates, etc, are meant to be parsed and used with setFieldValue: 

```
return <input 
  name="age" 
  // in v3 this is replaced with
  // parse={value => parseInt(value, 10)}
  onChange={event => {
    setFieldValue(event.target.name)(parseInt(event.target.value, 10))
  }}
/>
```

I think there would be value to adding a separate `handleValueChange` callback which only names a name or event, then accepts a pre-parsed value, circumventing the default parsing (`input type=number` is automatically cast to a number, etc). But I don't think this PR would work, for example you'd have the same exact issue as before if you passed `{ target: "This is my review of Target, the US store." }` as a value.

PR #3053, opened 2021-02-19
Make favicon href path absolute

dbernar1:
The way it is now does not work for the docs section, which I noticed at URLs like https://formik.org/docs/overview .

Please let me know if you need something different for this PR. Good package!


jaredpalmer (approved):

dbernar1:
🎉

PR #3063, opened 2021-02-24
Create snippet for the usage with useFormik with Yup

warifmust:
I thought it might be a good idea to have a separate snippet for the usage with `useFormik` and `Yup` validation, although I agree on the two different methods are pretty similar. However, it might be a helpful start for React Native newcomers.


docs/guides/react-native.md
```diff
@@ -45,3 +45,45 @@ DOM and React Native are:
...
+export const MyReactNativeForm = props => {
+    const formik = {
+        initialValues: {
+            email: ''
+        },
+        validationSchema: MySchema,
+        onSubmit: (values) => console.log(values)
+    }
+    
+    const {values, handleChange, handleBlur, handleSubmit} = formik;
```
seropaski, line 75: @warifmust Seems like you lost `useFormik` hook calling.

PR #3065, opened 2021-02-25
Fix setError type

dantman:
setError accepts an error not the Field's value type. The actual code implementing this interface is actually written `setError: (value: any) => setFieldError(name, value)`.


dantman:
Still relevant.

dantman:
Should still be relevant.

dantman:
Seems to be fixed by https://github.com/formium/formik/pull/3203 now.

Issue #2502, opened 2020-05-29
onValidationEnd handler


pro-nasa:
## 📖 Documentation
Could you give such an example in the documentation? I think it will save a lot of time for people who need to add a listener for validation (e.g. to change a button outside of Formik)

```
import { Formik, FormikErrors, yupToFormErrors } from 'formik'
import React from 'react'
import * as Yup from 'yup'

interface MyFormValues {
  name: string
  lastName: string
}

interface Props {
  initialValues: MyFormValues
  onSubmit: (values: MyFormValues) => any
  onValidationEnd: (isValid: boolean, errors: FormikErrors<MyFormValues>) => any
}

const validationSchema = Yup.object().shape<MyFormValues>({
  name: Yup.string().required(),
  lastName: Yup.string().required(),
})

export const MyForm: React.FC<Props> = ({ initialValues, onSubmit, onValidationEnd }: Props) => {
  const handleValidate = async (values: MyFormValues) => {
    try {
      await validationSchema.validate(values)
      onValidationEnd(true, {})
    } catch (err) {
      const errors = yupToFormErrors<MyFormValues>(err)
      onValidationEnd(false, errors)
      return errors
    }
  }

  return (
    <Formik
      initialValues={initialValues}
      onSubmit={onSubmit}
      onValidate={handleValidate}
      validateOnMount
      validateOnChange
    />
  )
}
```

Lukortech:
This looks pretty solid. 
It would be pretty handy to have it when I needed to pick a way to get forms to work in my project.

singleseeker:
really help, good job.

JGrenet:
Big big thanks !

PR #3072, opened 2021-03-01
Add onChange, onValidationErrror callback

ardyfeb:
- `onChange` will called every single value of field changed
- `onValidationError` will called when validation fails, (its not called if user provide custom validation function `props.validate`)

https://github.com/formium/formik/issues/1633
https://github.com/formium/formik/issues/1484

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -581,6 +581,13 @@ export function useFormik<Values extends FormikValues = FormikValues>({
           value,
         },
       });
+
+      if (props.onChange) {
+        props.onChange(
+          { ...state.values, [field]: value }, imperativeMethods
+        )
+      }
+      
       const willValidate =
         shouldValidate === undefined ? validateOnChange : shouldValidate;
       return willValidate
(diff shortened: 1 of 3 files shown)


johnrom:
There are a few issues with this. 

1. I'm not certain onX for things like validation handlers is the way I'd recommend going about things like validationError. Probably warrants more discussion from an API design level. Not totally against it, but it definitely isn't a change we'd make lightly.
2. The provided onChange event only occurs on a single type of change, doesn't detect setValues, or any other method of changing Formik's values (reset, reinitialize, etc). 
3. Finally, it actually emits a _new_ object instead of emitting the object created by setFieldValues (which only exist deep within the dispatch I think). It should return the same object.

Generally, onChange would occur via an effect instead of as an event when dealing with useReducer. The other option would be to abandon useReducer and keep the reducer logic outside of React, emitting these events during the processing of Formik's internal state which would be loosely coupled to React's state.

ardyfeb:
Hy, thanks for reviewing

2. reset event was handled by `onReset`
3. onChange should send new object, because because that's how it works

Listening changes with react effect user will lose previous value, so if we require to compare with old values will requiring extra hook to remember it

minimal Formik should expose what event dispatched internally, as you said https://github.com/formium/formik/issues/1633#issuecomment-545942765

johnrom:
Related #2502

ardyfeb:
Closing this, i'll keep my fork for anyone want this feature temporary

PR #3089, opened 2021-03-11
Use FormikApi + FormikState Subscriptions

johnrom:
This pivots Formik away from using Context for anything but the API itself, and exposes a `formik.useState()` implementation to subscribe to Formik State via a selector and comparer when you need to do specialized equality checks. Implemented with [use-subscription](https://www.npmjs.com/package/use-subscription).

### Upgraded Dev Environment

The new Development environment is awesome. Try it out. VS Code will now run your tests automatically with the Jest extension. Running `npm run start:app`, or just Launching from VS Code, will now run the `app/` folder aliased directly to Formik's source at localhost:3000, so you can test and develop Formik directly through /App. You can also launch Chrome against it to attach breakpoints to Formik.


packages/formik/src/hooks/useFormikComputedState.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+}
+
+const selectStateToCompute = (state: FormikState<any>) => ({
+  errors: state.errors,
+  values: state.values,
+  initialValues: state.initialValues,
+});
+
+export const useFormikApiComputedState = <Values = any>(formikApi: FormikApi<Values>) => {
```
jaredpalmer, line 30: consider removing? or replacing with just `useIsValid` `useIsDirty`

packages/formik/src/hooks/useFullFormikState.ts
```diff
@@ -0,0 +1,21 @@
+import { useMemo } from 'react';
+import { FormikApi, FormikComputedState, FormikState } from '../types';
+import { selectComputedState } from './useFormikComputedState';
+
+export const selectFullState = <T>(value: T) => value;
+
+export const useFullFormikState = <Values>(
```
jaredpalmer, line 7: full, computed, api, state, selectFull, need to reduce the surface area.

packages/formik/test/withFormik.test.tsx
```diff
@@ -103,6 +103,7 @@ describe('withFormik()', () => {
       setSubmitting: expect.any(Function),
       setTouched: expect.any(Function),
       setValues: expect.any(Function),
+      isFormValid: expect.any(Function),
```
jaredpalmer, line 106: isValid?

johnrom: isFormValid is a function passed by Formik which is needed to calculate whether the form is valid. Otherwise, we need to pass a bunch of config props from Formik through the API, and if a user doesn't memoize them we lose all of the performance gains.

packages/formik/src/hooks/useFormikComputedState.ts
johnrom: Agreed.

packages/formik/src/hooks/useFullFormikState.ts
johnrom: - `useFormikApi` is needed, as it gets the Context, which is now only the API
- `useFormikState` is an alias to `(selector, comparer) => useFormikApi().useState(selector, comparer)`, which seemed a little tedious to make users use. However, we have to expose that for any of this to work, so it's up to you whether we keep this hook or make users use `useFormikApi().useState()`
- `useFormikApiState(api, ...etc)` can probably go, since it is just `api.useState(selector, comparer)`;
- `selectFullState` is not something we should expose, it's just something we use internally instead of memoizing a new selector. I can go through and mark stuff like this @internal.
- `useFullFormikState` is used by backwards compatible `useFormikContext` which users expect to load the  legacy "formikBag". It's also used internally to subscribe to the full state in the case of Formik and Field `render` / `child fn`s which do not have access to hooks to get Formik's state.

jaredpalmer: Can useFormikState become useFormikStateSelector() or just useFormikSelector()?

I would also like to suggest changing useFormikApi() to useFormikActions() or useFormikMethods(). Mentally, easier to think about state or actions/methods versus state or api I think. api could be anything and it isn’t clear whether or it holds state.

jaredpalmer: Edit: I now see that these mean different things and my prior comment is wrong.

I still think API is a meaningless word and would like to see it changed. useFormikApiState vs. useFormikApi vs. useFormik vs. useFormikState isn’t intuitive for end users.

johnrom: Formik API is the only word here that makes sense to me.

- A getter is not an action
- A method is a function belonging to an object in OOP, which isn't really applicable to functional programming like Formik whereas it might apply more in `react-hook-form` since they use classes under the hood.
- The type returned by `useFormik` is `FormikApi` and the type exposed by Context is also `FormikApi`, so `useFormikApi` seems to be the best naming convention for the context since we can't use `useFormikContext`.

If you do not believe that `useFormik` returns an API, that is a whole different story. I don't think it returns Methods, or Actions. Maybe Helpers?? But we already draw a distinction in TypeScript between Helpers, Handlers, and Config.

Let's decide what the TypeScript name for the return value of `useFormik` is, and we'll change the name of `useFormikApi` to match. Whatever it is, the Context holds that value so the names should match.

johnrom: I think it'll be super clear that "API" is the right word for the API that Formik provides if we document it separately from the various components and hooks that we provide, and separately from the State.

A person who simply uses `useFormik()` and builds everything else from scratch is only using the Formik API, and we should document that separately. The current documentation uses API Reference to mean "all the things" whereas the API Formik provides is just what is returned by `useFormik` and every other component or hook we provide is a shortcut for using that API.

johnrom:
This currently passes the following props through the API which I think we should remove, because they are likely to cause re-renders if a user doesn't optimize them.

```ts
{
  validationSchema?: any | (() => any);
  validate?: (values: Values) => void | object | ValidateFn<Values>;
}
```

packages/formik/src/hooks/useFormikComputedState.ts
johnrom: I tried removing this, but then users lose the ability to calculate validity after useFormik as done right in the example app:

https://github.com/formium/formik/pull/3089/files#diff-13e2cca932df8f5229beec94cf2ea0c9f18dbfe117bcdd1c84f97041b216008fR33

In order to get computed state from the API both with or without context, we have to expose something that can calculate it from either place. useFormikApiX was the convention I was using for that purpose.

packages/formik/src/hooks/useFullFormikState.ts
johnrom: We can always break backwards compatibility and switch `useFormikContext` to just have the API, and then it'll be `useFormikContext` and `useFormikState`.

packages/formik/test/withFormik.test.tsx
johnrom: Removed isFormValid in favor of exposing formikApi.useComputedState().

packages/formik/src/hooks/useFullFormikState.ts
johnrom: I did the above, switched back to useFormikContext and useFormikState, abandoning backwards compatibility for clarity.

packages/formik/src/hooks/useFormikComputedState.ts
johnrom: I moved computed state to `api.useComputedState()` so it is accessible from anywhere, and only expose the hooks `useIsValid` and `useIsDirty`.

johnrom:
I found a small issue with this method. If we dispatch two values in a row, it's theoretically possible to skip a dispatch using `use-subscription` because it comes directly from `getState()` instead of `state`. This is fine for the render because it'll always become consistent with the latest info, but it's not OK if a dev needed two discrete effects to be triggered. This didn't happen in my original subscription implementation. I'll write a test for this scenario.

I'm going to start adding some of the `app/` code from #2846 and #2931 over the next week and do some more testing as well.

johnrom:
I figured out how to remove `useComputedState` entirely. Now `useState()` will return `FormikState`, which is `FormikReducerState + FormikComputedState`, simplifying the API dramatically.

johnrom:
@jaredpalmer this is ready for review and I believe I've resolved all the items above. I did the following:

- removed all State from FormikContext and switched back to it
- combined computed state into `api.useState` / `useFormikState`, removing all the confusion around computed vs reducer state
- removed isFormValid function as isValid is computed during getState()

You can test the branch locally by pulling the branch in VS Code and running the `run npm dev` task. It will boot up /app folder connected directly to /packages/formik/src.

packages/formik/src/Formik.tsx
```diff
@@ -5,29 +5,48 @@ import isPlainObject from 'lodash/isPlainObject';
...
 } from './utils';
 import { FormikProvider } from './FormikContext';
 import invariant from 'tiny-warning';
+import { useIsomorphicLayoutEffect } from './hooks/useIsomorphicLayoutEffect';
+import {
+  Comparer,
+  Selector,
+  useOptimizedSelector,
+} from './hooks/useOptimizedSelector';
+import { unstable_batchedUpdates } from 'react-dom';
```
jaredpalmer, line 42: how does this work with react native?

packages/formik/src/Formik.tsx
```diff
@@ -129,33 +148,45 @@ interface FieldRegistry {
...
+   */
   const isMounted = React.useRef<boolean>(false);
   const fieldRegistry = React.useRef<FieldRegistry>({});
+
   if (__DEV__) {
     // eslint-disable-next-line react-hooks/rules-of-hooks
     React.useEffect(() => {
       invariant(
-        typeof isInitialValid === 'undefined',
+        typeof props.isInitialValid === 'undefined',
```
jaredpalmer, line 189: Should we actually break the rules of hooks here? (i believe the rationale is that we only ever want to show this error message once) If so, let's add a comment about it or fix this.

packages/formik/src/Formik.tsx
```diff
@@ -170,18 +201,51 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     };
   }, []);
 
-  const [state, dispatch] = React.useReducer<
-    React.Reducer<FormikState<Values>, FormikMessage<Values>>
-  >(formikReducer, {
+  /**
+   * This is the true test of spacetime. Every method
+   * Formik uses must carefully consider whether it
+   * needs to use the ref or the render snapshot.
```
jaredpalmer, line 207: Elaborate more here in comments for outside contributors as to why we need this optimization. Include mental model for deciding whether to use ref and or snapshot.

packages/formik/src/Formik.tsx
```diff
@@ -334,53 +398,54 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     }
   );
 
-  React.useEffect(() => {
+  const performValidationOnMount = useEventCallback(() => {
```
jaredpalmer, line 401: `executeValidationOnMount` for consistency?

packages/formik/src/Formik.tsx
```diff
@@ -407,73 +472,123 @@ export function useFormik<Values extends FormikValues = FormikValues>({
...
+    (
+      nextInitialTouchedProp: FormikConfig<Values>['initialTouched'],
+      nextEnableReinitialize: FormikConfig<Values>['enableReinitialize']
+    ) => {
+      if (
+        nextEnableReinitialize &&
+        isMounted.current === true &&
+        !isEqual(state.initialTouched, nextInitialTouchedProp)
+      ) {
+        const touched = nextInitialTouchedProp || emptyTouched;
```
jaredpalmer, line 549:
```suggestion
        const touched = nextInitialTouchedProp ?? emptyTouched;
```

packages/formik/src/connect.tsx
```diff
@@ -31,7 +39,7 @@ export function connect<OuterProps, Values = {}>(
 
   // Assign Comp to C.WrappedComponent so we can access the inner component in tests
   // For example, <Field.WrappedComponent /> gets us <FieldInner/>
-  (C as React.FC<OuterProps> & {
+  (C as React.SFC<OuterProps> & {
```
jaredpalmer, line 42: I thought SFC was deprecated?

packages/formik/src/hooks/hooks.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useMemo } from 'react';
+import {
+  selectFieldMetaByName,
+} from '../helpers/field-helpers';
+import { FieldMetaProps, FormikState } from '../types';
+import { useFormikState } from './useFormikState';
+import { isShallowEqual } from '../utils';
+
+/**
+ * Returns @see FieldMetaProps<Value>
```
jaredpalmer, line 11: Improve comment. This will be shown in intellisense.

packages/formik/src/hooks/useFormikState.ts
```diff
@@ -0,0 +1,15 @@
+import { Selector, Comparer } from './useOptimizedSelector';
+import { useFormikContext } from '../FormikContext';
+import { FormikApi, FormikReducerState } from '../types';
+
+/**
+ * @see {@link FormikApi['useState']} for info on using Formik's State.
```
jaredpalmer, line 6: Improve comment

packages/formik/src/hooks/useOptimizedSelector.ts
```diff
@@ -0,0 +1,28 @@
+import { useMemo } from 'react';
+
+export type Selector<Value, Return> = (value: Value) => Return;
+export type Comparer<Return> = (prev: Return, next: Return) => boolean;
+
+const UNINITIALIZED_VALUE = Symbol();
+
```
jaredpalmer, line 7: Add comment

packages/formik/src/index.tsx
```diff
@@ -1,3 +1,5 @@
+export * from './hooks/useFormikState';
+export * from './hooks/hooks';
```
jaredpalmer, line 2: We are exporting `useOptimizedSelector`?

packages/formik/src/types.tsx
```diff
@@ -53,118 +55,281 @@ export interface FormikState<Values> {
...
+export type SetStatusFn = (status: any) => void;
+
+export type SetErrorsFn<Values extends FormikValues> = (
+  errors: FormikErrors<Values>
+) => void;
+
+export type SetSubmittingFn = (isSubmitting: boolean) => void;
+
+export type SetTouchedFn<Values extends FormikValues> = (
+  touched: import('./types').FormikTouched<Values>,
```
jaredpalmer, line 140: why is this `import()` here? is this importing itself? that works???

jaredpalmer (changes requested):
Almost there...

packages/formik/src/Formik.tsx
johnrom: Yeah, I didn't make any changes to this except to update the variable, but I believe you wrote it so that it would only print the warning once in development on the first use.

I would create a separate issue for this since it's unrelated to this PR.

packages/formik/src/types.tsx
johnrom: I think vs code autocompleted this... And somehow it works I guess?! Lol

packages/formik/src/Formik.tsx
johnrom: I'll have to test it out, but I think you're right this needs a different import for react native. I've never actually used React Native, but maybe we should have an expo version of app/ for testing?!

packages/formik/src/index.tsx
johnrom: In another PR I move useOptimizedSelector to a separate package. You should check that out before I modify this.

Also, I don't think I exported this but I'll double check.

packages/formik/src/Formik.tsx
johnrom: We don't need to manually batch updates here anymore as I moved this into an effect. I'll remove and test.

packages/formik/src/Formik.tsx
johnrom: I currently left the code functioning _how it does currently_ even though I created the ability to "fast-forward" via refs. But I need to go and revisit exactly what each bit of Formik uses, and whether it currently uses the right one. I'm pretty sure the answer is always _use getState()_ internally, except for code used inside hooks.

packages/formik/src/hooks/useOptimizedSelector.ts
johnrom: I've moved this package to its own unit tested npm package in another PR. see #3124.

packages/formik/src/index.tsx
johnrom: It is not exported.

packages/formik/src/Formik.tsx
johnrom: Created https://github.com/formium/formik/issues/3145 for this, as I wasn't involved in the __DEV__ checks.

johnrom:
@jaredpalmer I simplified the mental model here by enforcing a boundary between the `state` used under the hood to update subscriptions and thus used by hooks, and getState() which is used by imperative methods. I did this by moving the subscription portion back into its own hook.

I moved FieldHelpers to their own hooks as well, to enforce this boundary.

```
getFieldMeta -> useFieldMeta
getFieldHelpers -> useFieldHelpers
getFieldProps -> useFieldProps
```

Also fixed various other items and pushed back on items which can be addressed outside of this PR.

johnrom:
closed in favor of #3231

PR #3092, opened 2021-03-12
feat(type): add missing return type for useFormik

meeroslav:
**Actual:** 
`useFormik` has no defined return type, forcing the developer to guess the correct type in order to use the response object as a parameter

**Expected:** 
`useFormik` returns type `FormikProps<Value>`

**Note:** 
To make the return type complete, two properties - `isInitialValid` and `enableReinitialize` were added.


johnrom:
This will land in v3 #3089

PR #3105, opened 2021-03-18
Removing enclosed { } from validationSchema

Saliba:
it returns error "TypeError: schema[(intermediate value)(intermediate value)(intermediate value)] is not a function" if validationSchema is enclosed with { }

As seen on https://github.com/jquense/yup/issues/972

### DIFF
--- a/examples/SchemaValidation.js
+++ b/examples/SchemaValidation.js
@@ -28,7 +28,7 @@ const SignUp = () => (
         firstName: '',
         lastName: '',
       }}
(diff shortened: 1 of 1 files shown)


johnrom:
The Yup issue mentioned above was using `useFormik`, like: `useFormik({ validationSchema })`. Did you try these changes? I'm 99% certain it doesn't build, because `param=Value` isn't how prop syntax works. Props must either be strings, or wrapped with brackets.

PR #3124, opened 2021-03-25
Punt useOptimizedSelector to separate package

johnrom:
I am willing to take on the maintenance for this, since it could be useful for more than Formik. And it's very small!

https://www.npmjs.com/package/use-optimized-selector


johnrom:
closed in favor of #3231

PR #3125, opened 2021-03-25
Fix UseField type definition, add micro-hooks.

johnrom:
This adds the convenience hooks from the original v3.

It doesn't add useFieldX, useSetFieldX hooks, because users should simply use `useFieldMeta()` and the relevant API methods from `useFormikContext()`. It is not going to have a noticeable performance difference. `useFieldMeta` provides the same functionality as using all the `useFieldX()` combined.


dantman:
> It doesn't add useFieldX, useSetFieldX hooks, because users should simply use `useField()`. It is not going to have a noticeable performance difference. `useField` provides the same functionality as using all the `useFieldX()` combined.

Doesn't `useField` declare the field? I use `useFieldValue` and `useFieldSetValue` in various places other than the field itself. e.g. A Collapse component that depends on the value of a checkbox. An effect that resets the value of a field when conditions change in a way that will make the field become invalid. etc.

johnrom:
You can use `useFieldMeta` for this purpose which supplies the underlying state.

dantman:
Ok, fine. Though this does look a lot uglier.

```js
const { setFieldValue } = useFormikContext();
const { value: otherValue } = useFieldMeta(props.otherFieldName);

useEffect(() => {
  console.log(otherValue);
  setFieldValue(props.otherFieldName, 'foo');
}, [otherValue, setValue, props.otherFieldName]);
```

Than the previous v3 code:

```js
const [otherValue, setOtherValue] = useFieldValue(props.otherFieldName);

useEffect(() => {
  console.log(otherValue);
  setOtherValue('foo');
}, [otherValue, setOtherValue]);
```

johnrom:
I agree it could be more concise. The previous concise API had a reason - the smallest possible hooks were necessary to get optimized subscriptions.

This one doesn't need it, and in fact using smaller subscriptions may have a performance hit because each one is a subscription. I think a balance of this would be to return Formik API from useFieldMeta:

`const [fieldMeta, api] = useFieldMeta()`

johnrom:
closed in favor of #3231

PR #3126, opened 2021-03-29
added setFieldValue function overload for better editor support

Bezmehrabi:
most of the time, we probably want to `setFieldValue` one of the existing fields in initial values, so it's good to have some editor support for it. :)

Screenshot_20210329_122730

I expected something like this to work out but it doesn't:
```setFieldValue = (field: keyof Values | string, value: any, shouldValidate?: boolean) => void;‍‍‍‍‍‍```

### DIFF
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -92,7 +92,8 @@ export interface FormikHelpers<Values> {
     shouldValidate?: boolean
   ) => void;
   /** Set value of form field directly */
-  setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void;
+  setFieldValue(field: keyof Values, value: any, shouldValidate?: boolean): void;
+  setFieldValue(field: string, value: any, shouldValidate?: boolean): void;
   /** Set error message of a form field directly */
   setFieldError: (field: string, message: string | undefined) => void;
   /** Set whether field has been touched directly */


johnrom:
values can be nested so this doesn't work. in order to determine options for how one might strongly type the name, check out #1334

Bezmehrabi:
Ooo, ok I see, I'll try to work on that in my free time.

Bezmehrabi:
@johnrom shouldn't the type of `meta` be `meta?: FieldMetaProps<FieldValue>`(possibly undefined) because of the backward compatibility?

johnrom:
You can see how setFieldValue can be completely typed, [here](https://github.com/formium/formik/pull/3152/files/a9216a0295d46614097372b0de6e8ec2d04bc113..7d595921c829c3e8e43bb4f0a9cb2ea9802dd33b?file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-c1dc709f52d905c50678cf451af56657c12990daa96a7764bbb4e9066758d42bR158)

johnrom:
CodeSandbox here: https://codesandbox.io/s/serene-napier-4wjxs?file=/src/App.tsx:792-1010

Bezmehrabi:
nice!

PR #3152, opened 2021-04-18
Strongly Typed Formik, v3

johnrom:
Adds strong types to Formik by typing Fields with `Field<FormValues, Path>`. This is an extremely breaking change.

Todo: 

- FieldArray
- Type "AnyFieldName" with a flat list of possible field names. I couldn't figure out how to do this. Without it, we cannot provide intellisense.


johnrom:
I haven't been able to infer array types properly. Maybe someone (or future me) can find a workaround.

 https://github.com/microsoft/TypeScript/issues/43907

johnrom:
I was able to infer Array types by inferring all array numerics as "0", and then converting "0" to `${number}` template literal later in the PathMatchesValue type. It starts to push TypeScript to its limit...

jaredpalmer:
I am using this branch on Turborepo. Will report back with errors

nandorojo:
Really excited for this.

johnrom:
@nandorojo linked an npm package above you can use to play with these changes if you want to test them out. I can't promise it won't break anything!

johnrom:
Added `PathLikeValue` alongside `PathMatchingValue` for functions like setFieldValue where the value might be a narrower type and the path could match a wider type, like `setFieldValue("numberOrString", numValue);`. There's still an issue specifically with constant strings, like `number | ""` that will require casting due to https://github.com/microsoft/TypeScript/issues/30808

PR #3157, opened 2021-04-21
fix: submitForm promise rejects if validation fails

### LINKED ISSUE #2958: submitForm returned promise not rejected if form is invalid
## Bug report

Promise is resolved both when form is valid and invalid.

### Expected behavior

According to the [docs](https://formik.org/docs/api/formik#submitform---promise) the promise should be rejected if the form is invalid.

### Reproducible example

https://codesandbox.io/s/formik-example-forked-ru3i1

Submit the form and check the console

### Suggested solution(s)

Reject the promise as specified in the docs

### Your environment

Check codesnabox example

Averethel:
Fixes #2958

### DIFF
--- a/.changeset/heavy-islands-give.md
+++ b/.changeset/heavy-islands-give.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
+
+submitForm promise rejects if validation fails
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -783,9 +783,10 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         } else if (!!isMounted.current) {
           // ^^^ Make sure Formik is still mounted before updating state
           dispatch({ type: 'SUBMIT_FAILURE' });
-          // throw combinedErrors;
           if (isInstanceOfError) {
             throw combinedErrors;
+          } else {
+            throw new Error('Submit failure');
           }
         }
         return;
--- a/packages/formik/test/Formik.test.tsx
+++ b/packages/formik/test/Formik.test.tsx
@@ -528,6 +528,18 @@ describe('<Formik>', () => {
         });
       });
 
+      it('should reject the promise if the form is invalid', async () => {
+        const onSubmit = jest.fn();
+        const validate = jest.fn(() => ({ name: 'Error!' }));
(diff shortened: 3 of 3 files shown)


Averethel:
@jaredpalmer @johnrom anything I can do to get guidance here?

johnrom:
I don't think throwing errors (usually known as exceptions in other languages) makes sense here. Exceptions are meant for exceptional behavior, and a form being invalid is not exceptional. It is not a state from which the application needs to recover.

I think it makes more sense to return a result which can indicate whether the submit actually went through.

It only really makes sense to me to throw if someone throws an error themselves within onSubmit, because the developer is basically throwing up their hands and saying they can't proceed with the code. Formik doesn't have that issue, it can simply not run onSubmit and the application will continue functioning normally.

Averethel:
@johnrom Thank you for your answer. 
I stumbled upon this when I tried submitting the form programmatically, when I checked the docs the `submitForm` is documented as follow:

> Trigger a form submission. The promise will be rejected if form is invalid.

This is not the case when I have a `validation` function that returns an object containing errors, hence my change.

Let me try to explain my use-case and why I need to submit the form programmatically in the first place, maybe this helps to understand my motivation, or maybe there is another way to achieve what I want.
I have a very complex multi-step, branching, wizard style form. Depending on which submit action is used different things[1] could happen after submitting the form, but only if the submission is successful.
My previous approach was to use different `value` for submit buttons and handle success in the submission handler but with every new action the submission handler grew and was hard to maintain. Plus finding corresponding behaviours simply become a nightmare.

So I reached to `submitForm` and wanted to add the specific post submission handling in the event handlers on the buttons themselves. But, in the current implementation, there is no way of knowing if the submission did not happen at all because of failed validations. Maybe rethinking the approach is in order? What if `submitForm` was a promise returning a boolean instead of rejection if submission failed?


-----
[1] this could include navigating to a different step, revalidating the form with a different set of constraints, creating additional nested item, etc.

johnrom:
I think it could return something. Probably not a boolean, because it's possible there could be more states in the future.

validation_failed, submit_failed, submit_success, submit_cancelled, etc

Averethel:
@johnrom That makes sense, I like it. I take another look over the weekend and see how I can make that work

PR #3177, opened 2021-04-30
Require `ExtraProps` to be explicit

### LINKED ISSUE #3170: <Formik /> has loose prop types
### Current Behavior
`<Formik />` accepts any prop, this make it impossible for TS compiler to find typos on props name.

### Expected behavior
IMHO, unexpected props should not be acceptable by default, but rather require explicit type
### Reproducible example

```ts
return (<Formik
  initialValues={myInitialValue}
  validationScheme={myValidationSchema} // fails to spot the typo.
  onSubmit={onSubmit}
>)
```

### Suggested solution(s)
Currently, passing `Formik` explicit type parameters works around this behavior.

```ts
return (<Formik<MyValueType>
  initialValues={myInitialValue}
  validationScheme={myValidationSchema} // Object literal may only specify known properties, but 'validationScheme' does not exist in type 'FormikConfig<MyValueType>'. Did you mean to write 'validationSchema'?
  onSubmit={onSubmit}
>)
```
This works because it bypass type inference for type parameters and the compiler was inferring the `ExtraProps` with `validationScheme` on it.
To solve this it's needed to somehow disable inference for `ExtraProps`.
The way i find to work is to change:
```ts
export declare function Formik<Values extends FormikValues = FormikValues, ExtraProps = {}>(props: FormikConfig<Values> & ExtraProps): JSX.Element;
```
to
```ts
type NoInfer<T> = [T][T extends any ? 0 : never];
export declare function Formik<Values extends FormikValues = FormikValues, ExtraProps = {}>(props: FormikConfig<Values> & NoInfer<ExtraProps>): JSX.Element;
```
### Your environment

| Formik           | 2.1.5
| React            | 16.13.0
| TypeScript       | 4.2.3
| npm/Yarn         | 1.22.10

munizart:
As explained in #3170, due to type inference TS is unable to report typos when passing unknown properties to `<Formik />`

This proposal uses a conditional type `NoInfer<T>` to bypass type inference for the `ExtraProps` type parameter, while still enabling users to pass it explicitly.

Please note this is a opinionated proposal, therefor further discussion may be needed.
closes #3170

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -987,10 +987,11 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   return ctx;
 }
 
+type NoInfer<A> = [A, never][A extends any ? 0 : 1];
 export function Formik<
   Values extends FormikValues = FormikValues,
   ExtraProps = {}
->(props: FormikConfig<Values> & ExtraProps) {
+>(props: FormikConfig<Values> & NoInfer<ExtraProps>) {
   const formikbag = useFormik<Values>(props);
   const { component, children, render, innerRef } = props;
 


johnrom:
This is pretty cool, but I'd be concerned about a future version of TypeScript removing this optimization and being able to infer the underlying value itself. If that happened, I think it would be fine as it would simply "loosen the type" and not break any existing implementations, but I'm not 100% able to evaluate what would happen in this circumstance.

PR #3188, opened 2021-05-10
Add note about using React.Fragment within FieldArrray

DysphoricUnicorn:
I generally try to avoid using `<div>`s where they are not necessary to avoid building huge pyramids in the dom.  
In many instances, replacing `<div>` tags with `<React.Fragment>` works without issue.  
Sadly, here it causes the render to fail with a somewhat cryptic error message that does not show where the issue lies and is difficult to debug.

While creating an issue for this would be possible, I am unsure if it is "wanted" behavior and therefore I thought it would be best to just propose a change to the docs, warning about this.

### DIFF
--- a/docs/examples/field-arrays.md
+++ b/docs/examples/field-arrays.md
@@ -13,3 +13,6 @@ This example demonstrates how to work with array fields in Formik.
   sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
   ></iframe>
 </div>
+
+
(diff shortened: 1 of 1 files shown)


johnrom:
@DysphoricUnicorn can you share the error that you're seeing? I don't know of any specific limitations surrounding Fragment here, so I would consider this a bug.

DysphoricUnicorn:
Yeah, sorry for any work this PR caused. Apparently the bug I had encountered was not caused by the Fragment even though replacing it fixed it.
Today I went back to put in a Fragment again and it didn't happen.

Anyways, thank you for the quick response :)

PR #3200, opened 2021-05-18
Docs: @types/yup no longer required

espipj:
Seems that recently yup moved to have built-in typscript support https://github.com/jquense/yup/blob/HEAD/docs/typescript.md
https://www.npmjs.com/package/yup shows as typescript package:
image


jaredpalmer (approved):

jaredpalmer:
Thanks!

PR #3203, opened 2021-05-20
fix:  value  is same as message in  setFieldError

hixus:
setError is setFieldError bound to specific field. Therefore the setError value should have same type as setFieldError message


jaredpalmer:
Yup

jaredpalmer (approved):

PR #3214, opened 2021-05-25
add useFormik return type

alanszp:
Based on https://github.com/formium/formik/issues/2924

Create return type for useFormik.
Based on @johnrom 's  v3 commit for some naming: https://github.com/formium/formik/pull/3121/files#diff-2421f0a25f33b0ee81ca1de9dee2a1743711e08f273a5733ed03493de527de21R135-R137

But if you have a better name for the return type, please tell me and I'll refactor it.

------------------

Will explain step by step how I did it so you can have a quick look on my steps and maybe detect a bug without checking the code:

- Used the return type definition exported in dist.
- Removed 1 by 1 the fields when I found a good type which encapsulated all the props with the same firm.
- Had to divide the current type: FormikSharedConfig. Extracted from there FormikValidationConfig.
- Profit

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -14,6 +14,7 @@ import {
   FieldInputProps,
   FormikHelpers,
   FormikHandlers,
+  FormikApi,
 } from './types';
 import {
   isFunction,
@@ -137,7 +138,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   enableReinitialize = false,
   onSubmit,
   ...rest
-}: FormikConfig<Values>) {
+}: FormikConfig<Values>): FormikApi<Values> {
   const props = {
     validateOnChange,
     validateOnBlur,
@@ -947,7 +948,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
     [isInitialValid, dirty, state.errors, props]
   );
 
-  const ctx = {
+  const ctx: FormikApi<Values> = {
     ...state,
     initialValues: initialValues.current,
     initialErrors: initialErrors.current,
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -150,16 +150,19 @@ export interface FormikHandlers {
   getFieldHelpers: <Value = any>(name: string) => FieldHelperProps<Value>;
 }
 
-/**
- * Base formik configuration/props shared between the HoC and Component.
- */
-export interface FormikSharedConfig<Props = {}> {
+export interface FormikValidationConfig {
   /** Tells Formik to validate the form on each input's onChange event */
   validateOnChange?: boolean;
   /** Tells Formik to validate the form on each input's onBlur event */
   validateOnBlur?: boolean;
   /** Tells Formik to validate upon mount */
   validateOnMount?: boolean;
+}
+
+/**
+ * Base formik configuration/props shared between the HoC and Component.
+ */
+export interface FormikSharedConfig<Props = {}> extends FormikValidationConfig {
   /** Tell Formik if initial form values are valid or not on first render */
   isInitialValid?: boolean | ((props: Props) => boolean);
   /** Should Formik reset the form when new initialValues change */
@@ -247,6 +250,14 @@ export interface FormikRegistration {
   registerField: (name: string, fns: { validate?: FieldValidator }) => void;
   unregisterField: (name: string) => void;
 }
+
+export type FormikApi<Values extends FormikValues = FormikValues> =
+  & FormikHelpers<Values>
+  & FormikHandlers
+  & FormikRegistration
+  & FormikComputedProps<Values>
+  & FormikState<Values>
+  & FormikValidationConfig;
 
 /**
  * State, handlers, and helpers made available to Formik's primitive components through context.
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -251,13 +251,14 @@ export interface FormikRegistration {
   unregisterField: (name: string) => void;
 }
 
-export type FormikApi<Values extends FormikValues = FormikValues> =
-  & FormikHelpers<Values>
-  & FormikHandlers
-  & FormikRegistration
-  & FormikComputedProps<Values>
-  & FormikState<Values>
-  & FormikValidationConfig;
+export type FormikApi<
+  Values extends FormikValues = FormikValues
+> = FormikHelpers<Values> &
+  FormikHandlers &
+  FormikRegistration &
+  FormikComputedProps<Values> &
+  FormikState<Values> &
+  FormikValidationConfig;
 
 /**
  * State, handlers, and helpers made available to Formik's primitive components through context.


johnrom:
These types aren't actually compatible -- the FormikHelpers aren't all correct (they already were not correct). I fixed them in my PR, but for example if you remove the return type from useFormik and add the following line somewhere, you'll get an error, with a number of underlying differences:

```tsx
// try assigning FormikApi to the return type of useFormik
const test: Partial<ReturnType<typeof useFormik>> = {} as any as FormikApi<{}>;
```

A number of types are incompatible, including:

- setFieldValue return type
- setFieldTouched return type
- setTouched return type
- setValues return type
- validateField return type

I think once those are fixed, this looks good to me.

The problem is the API has a lot of setX or validateX as returning `void` even though they actually return Promise<void> | Promise<FormikErrors>. Those functions are technically assignable to a function returning void even though they return something, but it would hide their return values for devs in typescript.

alanszp:
Nice catch @johnrom !

Making a summary of the breaking types so you can analyze faster:


| Field | Original definition | Real output inferred by TS 
| ----- | ------ | ------ 
| setTouched | `( ... ) => void`  |  `( ... ) => Promise<FormikErrors<Values>> \| Promise<void>` 
| setValues | `( ... ) => void`  |  `( ... ) => Promise<FormikErrors<Values>> \| Promise<void>` 
| setFieldValue | `( ... ) => void`  |  `( ... ) => Promise<FormikErrors<Values>> \| Promise<void>` 
| setFieldTouched | `( ... ) => void`  |  `( ... ) => Promise<FormikErrors<Values>> \| Promise<void>` 
| validateField | `( ... ) => void`  |  `( ... ) => Promise<void> \| Promise<string \| undefined>` 
| setFormikState | `(firstArg, cb?: () => void) => ...` | `(firstArg) => ...`
| registerField | `(firstArg, fns: { validate?: FieldValidator }) => ...` | `(firstArg) => ...`

--------------


**Made a check on every prop**

Making another branch with commit enforcing the types all around the code: https://github.com/alanszp/formik/pull/1/files

| Field | Check result | Commit
| ----- | ------  | ------
| setTouched | Added the inferred values that were right. In the same commit did it for setFieldTouched. Added typings to getFieldHelpers too to check everything works ok | https://github.com/formium/formik/commit/83806f8d267d9e330fd33a59e0fef3eeee2a835f
| setValues | Added the inferred values that were right.  | https://github.com/formium/formik/commit/8a4001dbf942d5c74a361e186f9dbea1ff2006e1
| setFieldValue |  Also added the same inferred value to setValue from FieldHelperProps since it was returning the result of setValues  | https://github.com/formium/formik/commit/2e8778d22da01abf91974ae41ef1266bfb14ae7d
| setFieldTouched | Same as setTouched | https://github.com/formium/formik/commit/83806f8d267d9e330fd33a59e0fef3eeee2a835f
| validateField | Added the inferred values that were right. Also removed `.then((x: any) => x)` that was unnecessary (add it back if I'm correct. Here I found a potential bug explained below *1 | https://github.com/formium/formik/commit/a5135b5d3b1620159fa13257300fbc03fd679b10
| setFormikState | Removed second argument callback. Was never used in code | https://github.com/formium/formik/commit/359284f60ef7126fda1515aab86498f42567c6e3
| registerField | Made a check weather validate is present and enforced types around. Reused FieldValidator type | https://github.com/formium/formik/commit/c0794643b6b7ae4666d5f9a78d351acb01720ec1

--------------

**Warnings with types:** 

*1

In `validateField` we now have this code: 

```ts
return runValidationSchema(state.values, name)
        .then((error: any) => {
          dispatch({
            type: 'SET_FIELD_ERROR',
            payload: { field: name, value: error[name] },
          });
          dispatch({ type: 'SET_ISVALIDATING', payload: false });
        });
```

The potential problems (which I detected with types and not with the proper knowledge of what it does/how it works) is:
`runValidationSchema` returns: `FormikErrors` which is an Object like [key = fieldName]: string OR **string[]** (and possible more combinations)

Now when we type `error` as `any`, we are hiding the **string[]**. The SET_FIELD_ERROR only supports the `payload.value` to be a string.

So now we have one of two problems:
- SET_FIELD_ERROR's  `payload.value` mus be a string | string[] ...  (which I think it's correct)
- There might be a bug when runValidationSchema returns an array of errors for a field.


*2 

I couldn't fix, but useEventCallback as it's defined is hiding some type errors, mainly on the return side. And the <T> generic is not working well to type/enforce the function passed as arg.

*3

FormikHandlers #setValue & #setTouched may be set with better typings so can inherit the Value type from the context. Didn't handle this in this PR.

johnrom:
@alanszp when you approve and merge your own PR to your own branch I can take a look at the final version to review this PR.

in terms of the types being _correct_, it's not super necessary to fix every type that was previously incorrect (there are a lot, and I think I fixed them all in my V3 PR), but we do need these types to be backwards compatible with the previous useFormik types so devs already using these definitions don't experience any breakages. does that make sense?

alanszp:
@johnrom merged. I didn't before so you and @jaredpalmer could check the main change without all the fixes.
But I checked all returns in functions used all around the code and the types are those stated, so I think we can merge this.

About typefixes I used all types inferred by typescript, so I don't add breaking changes. (used the table [in this comment](#issuecomment-849257934))

I think that it will not break but I will take another look today to doublecheck

packages/formik/src/types.tsx
```diff
@@ -113,8 +117,7 @@ export interface FormikHelpers<Values> {
   setFormikState: (
     f:
       | FormikState<Values>
-      | ((prevState: FormikState<Values>) => FormikState<Values>),
-    cb?: () => void
```
alanszp, line 117: this may be the only breaking change in the API (but it's not used anymore). Maybe we can mark it as `deprecated`?

packages/formik/src/Formik.tsx
```diff
@@ -504,35 +510,37 @@ export function useFormik<Values extends FormikValues = FormikValues>({
             value: maybePromise as string | undefined,
           },
         });
-        return Promise.resolve(maybePromise as string | undefined);
+        return Promise.resolve(maybePromise);
       }
     } else if (props.validationSchema) {
       dispatch({ type: 'SET_ISVALIDATING', payload: true });
-      return runValidationSchema(state.values, name)
-        .then((x: any) => x)
```
alanszp, line 512: this is not a type change, but removed as I thought it was left behind in a refactor. 
Maybe it's to avoid a race-condition?

johnrom: let's keep this a typescript-only change

packages/formik/src/Formik.tsx
```diff
@@ -504,35 +510,37 @@ export function useFormik<Values extends FormikValues = FormikValues>({
...
 
     return Promise.resolve();
   });
 
-  const registerField = React.useCallback((name: string, { validate }: any) => {
-    fieldRegistry.current[name] = {
-      validate,
-    };
-  }, []);
+  const registerField = React.useCallback<FormikRegistration['registerField']>(
```
johnrom, line 529: please keep this a typescript only change. this change is actually breaking. not every field has a validate fn

packages/formik/src/types.tsx
johnrom: I think you're right, seems old. We can add `@deprecated` here.

packages/formik/src/types.tsx
```diff
@@ -318,6 +339,6 @@ export interface FieldInputProps<Value> {
   onBlur: FormikHandlers['handleBlur'];
 }
 
-export type FieldValidator = (
-  value: any
-) => string | void | Promise<string | void>;
+export type FieldValidatorResponse = string | void | Promise<string | void>;
```
johnrom, line 342: Response to me implies there is a Request, like ajax or XHR. I'd go with `FieldValidationResult` as it doesn't imply any kind of Request.

johnrom (changes requested):
Some small changes requested, afterwards I think it looks good.

packages/formik/src/Formik.tsx
alanszp: this was added bc validate could be undefined, and fieldRegistry only let me add string. If you prefer we can cast it as string and continue with this behavior

packages/formik/src/types.tsx
alanszp: Perfect! It seems that JSDocs can't annotate a param as deprecated, only values.

I changed the name from cb to cbDeprecated to be blunt about it and add an inline deprecation that I read that typescript interprets (but it's not working on vscode).

If you have a better alternative, please tell me :)

packages/formik/src/Formik.tsx
alanszp: rollback the any in the param so we dont have a logic change

alanszp:
@johnrom did all the changes. I think it's ready to go

johnrom:
@alanszp hope your weekend is going well. added a PR on top of your PR with some changes to fix up the type a bit.

alanszp:
@johnrom hope yours too!

Great, I just realized what you meant in [your comment](https://github.com/formium/formik/pull/3214#discussion_r640912294). I didn't saw the isFunction check where it's used.

I just merged it, but was thinking for a future PR: isn't better to make validate field as a FN instead of FN | undefined?
you can check only once (in the setter) and then you don't have to worry if it's a function or not.  Really don't know very well the internals, so maybe this is not a possibility.

johnrom (approved):

johnrom:
Not really, because validate can be undefined if there is no field validation on a field. I think it would be odd to add a default validation function, especially because it might be useful to change behavior depending on whether a field has a validation function or not down the road.

It's possible to remove the optional checks by assigning the variable earlier (we already check whether the function exists, the optional check is just to make typescript happy. (TypeScript sees that theoretically `fieldRegistry.current[name]` could change between checks in a multi-threaded environment, which is correct, but I don't think it's necessary to fix that issue at this time. Famous last words)

@jaredpalmer lgtm

alanszp:
Thanks for the explanation about fn check! 

I didn't imagine a default validation function. Will elaborate a little bit better just for you to understand, but I'm ok with the state of the art.

My comment on the topic was: Just to fill the validation[field] when there is a real validation fn, so you either have a fn or directly don't have a property for that field in FieldRegistry.

Right now, we have 3 states in FieldRegistry for each field:
1. field does not exist
2. field is undefined
3. field is fn

Bc 1 & 2 are virtually the same, we could restrict not setting undefined in the field.

But anyway, this is another PR/issue talk. Let's go with the new typings!

Issue #3181, opened 2021-05-04
Unexpected value of meta.initialValue


tadeas22:
### Current Behavior

Value of `meta.initialValue` is null in case if path is `path.to.property` for object 

```json
{
"path": {
  "to": null
 }
}
```

```javascript
const [_, meta] = useField("path.to.property")
meta.initialValue === null
```

### Expected behavior

Value of `meta.initialValue` is undefined

```javascript
const [_, meta] = useField("path.to.property")
meta.initialValue === undefined
```

### Suggested solution(s)

```javascript
export function getIn(
  obj: any,
  key: string | string[],
  def?: any,
  p: number = 0
) {
  const path = toPath(key);
  while (obj && p < path.length) {
    obj = obj[path[p++]];
  }

  // check if path is not in the end
  if(p !== path.length && obj === null)
    return undefined

  return obj === undefined ? def : obj;
}
```

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           |2.2.5|
| React            |16.14.0|
| Operating System |linux|

pieplu:
We are experiencing the same problem.
However, this does not only apply to nulls.
But for all the falsy:

https://codesandbox.io/s/throbbing-resonance-2tm55?file=/src/index.ts

![image](https://user-images.githubusercontent.com/3260285/120690667-8107fe80-c473-11eb-8431-60b093ee3f3b.png)
![image](https://user-images.githubusercontent.com/3260285/120690689-86fddf80-c473-11eb-84ec-6590605df6f7.png)

tadeas22:
PR is still open

PR #3237, opened 2021-06-11
Fix getIn unexpected value

pieplu:
As described in #3181

It scares me a bit, touching getIn seems a big change.
But this will correct a bug, rare, but a bug in my opinion


johnrom:
This seems like a good change to me, but we'll definitely want thorough testing when modifying this particular bit of code.

packages/formik/src/utils.ts
```diff
@@ -66,18 +66,24 @@ export function getActiveElement(doc?: Document): Element | null {
...
-export function getIn(
-  obj: any,
-  key: string | string[],
-  def?: any,
-  p: number = 0
-) {
-  const path = toPath(key);
-  while (obj && p < path.length) {
-    obj = obj[path[p++]];
+ export function getIn(
```
johnrom, line 69: Looks like the formatting got messed up here. Is this indented correctly?

pieplu: You're right, prettier doesn't seem to run automatically. I just fixed it, and added more tests.

pieplu:
up

pieplu:
We have been using this patch in our application, on the test environment, for a few weeks, no regression found yet.
We have about 100 different components using Formik.

I totally understand that there is a fear with a change in such an important part of Formik, and that it takes time. I will keep you posted when we have deployed this to our production environment.

pieplu:
We have been using this patch in our production environment for almost a month.
We have at least 40 users, who use our form components to create and update content. All day, every day.
No regression has been seen, and the related bug no longer occurs.

@jaredpalmer

jaredpalmer (approved):

PR #3286, opened 2021-07-26
set onBlur to TextFields in material-ui example

### LINKED ISSUE #3183: Docs - Fomik React Material-UI example missing onBlur for touched
### Current Behavior

On the docs https://formik.org/docs/examples/with-material-ui 

the example is missing a very important prop to be passed into the TextField component
`onBlur={formik.handleBlur}`

This is needed to show errors when the user has click away from the textfield. otherwise the errors wont show until you click submit. But in a lot of use cases, you want the submit button to be disabled until the form is valid.
### Expected behavior

Add `onBlur={formik.handleBlur}`

I think this is straight-forward. let me know if you need actual examples to try this. i figure you can just add this prop to the editor at https://formik.org/docs/examples/with-material-ui  and try it there.

tsekityam:
set `onBlur` to TextFields based on the suggestion from @jeffbski

### DIFF
--- a/examples/with-material-ui/index.js
+++ b/examples/with-material-ui/index.js
@@ -38,6 +38,7 @@ const WithMaterialUI = () => {
           label="Email"
           value={formik.values.email}
           onChange={formik.handleChange}
+          obBlur={formik.handleBlur}
           error={formik.touched.email && Boolean(formik.errors.email)}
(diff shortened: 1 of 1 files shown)


tsekityam:
Closing stale PR opened before Aug 2026.

PR #3359, opened 2021-09-29
Fix error in FastField docs - Field render

jcomo:
The example cost in the FastField docs has a component that does not unpack the render props. This PR simply adds in the props so that the example can be copy/pasted without error.


jcomo:
No changeset needed

PR #3362, opened 2021-09-30
Make formik helpers types stricter

khevamann:
This PR updates the types of the helpers such as `setFieldValues` to make sure that only field names that exist in the FormValues type are allowed. Provides better completion as well as error detection.

### DIFF
--- a/packages/formik/src/types.tsx
+++ b/packages/formik/src/types.tsx
@@ -92,19 +92,19 @@ export interface FormikHelpers<Values> {
     shouldValidate?: boolean
   ) => void;
   /** Set value of form field directly */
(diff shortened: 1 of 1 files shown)


khevamann:
@johnrom Ok so should I just close this in favor of that PR or make a change to support the nested fields?

johnrom:
I would probably recommend checking out that PR as it has tons and tons of type fixes. You can branch from there and continue trying to get the TS even tighter (I haven't kept up with some of the more recent releases which could have even more powerful methods of typing), or you can test it in development by using [this fork](https://www.npmjs.com/package/@johnrom/formik-v3/v/3.1.0-types12)

PR #3406, opened 2021-11-09
cleanup arrays if the last value, which is not undefined, gets removed from it

giggo1604:
This fixes an error where you can get to an invalid error state if you remove an array field which has a corresponding error.

Here is a basic example of the original problem: https://codesandbox.io/s/formik-example-forked-gi0kc?file=/index.js

If the empty field gets removed the form will not turn valid because the array is not "empty" and therefore gets not removed from the parent object.

An other way to to fix this might be to rewrite the `isEmptyArray` assertion to also consider arrays with only `undefined` values as empty.

### DIFF
--- a/.changeset/fair-students-fix.md
+++ b/.changeset/fair-students-fix.md
@@ -1,0 +1,6 @@
+---
+'formik': patch
+'formik-native': patch
+---
+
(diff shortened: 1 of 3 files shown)


giggo1604:
@jaredpalmer @johnrom any chance to get this process going?

PR #3439, opened 2021-12-16
[Formik Website] Remove typescript badge image to prevent broken links on the page

aflashyrhetoric:
## Screenshot of the issue
Screen Shot 2021-12-15 at 10 08 23 PM

It seems that the badge at the top of [this page](https://formik.org/docs/guides/typescript) is getting rendered as a page-wide link that covers the rest of the links. 

## Proposed Fix
- I'm not sure if it's necessary, but remove the badge (which currently links to Formik's NPM page) seemed to work. 

If the badge needs to remain, please let me know - I can try to dig into it from a different angle. Cheers to the Formik team!

### DIFF
--- a/docs/guides/typescript.md
+++ b/docs/guides/typescript.md
@@ -3,8 +3,6 @@ id: typescript
 title: TypeScript
 ---
 
-[![TypeScript Types](https://img.shields.io/npm/types/formik.svg)](https://npm.im/formik)
-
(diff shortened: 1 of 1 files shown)


franky47:
The problem seems to come from the Next.js `<Image>` component being used to render the SVG badge from Shields.io. 

Vercel/Next fails to optimise the SVG with a 400 Error: INVALID_IMAGE_OPTIMIZE_REQUEST:
image

Issue #3455, opened 2022-01-11
FormikTouched is typed incorrectly for primitive array values


kmurgic:
### Current Behavior
For values `{ x: string [] }`, touched is being typed as `{ x?: boolean }`

### Expected behavior

For values `{ x: string [] }`, touched should be typed as `{ x?: boolean[] }`

### Reproducible example

https://codesandbox.io/s/infallible-shape-nhd9i?file=/index.tsx:537-685
Edit the names of one of the friends. Note that in the console touched.friends is an array of booleans.

### Suggested solution(s)

Replace `boolean` with `boolean[]` [here](https://github.com/jaredpalmer/formik/blob/master/packages/formik/src/types.tsx#L30)
I plan on making a small PR soon.

kmurgic:
Opened a [PR](https://github.com/jaredpalmer/formik/pull/3456)

quantizor:
I think this is intended behavior, `touched` is describing the field not the contents of the field.

PR #3456, opened 2022-01-11
Change touched type for primitive array value from boolean to boolean[]

kmurgic:
Fixes https://github.com/jaredpalmer/formik/issues/3455


kmurgic:
Not sure why this is failing CI

kmurgic:
@johnrom @jaredpalmer 

I'm sure there are much higher priority issues than a small TS error like this one, but even though this for some inexplicable (to me) reason this is failing CI, I'm pretty confident this PR is correct and it is an easy one liner. Any chance you can take a few minutes to review and (hopefully) merge this?

johnrom:
This project appears to be dead, so I would recommend switching to another popular form provider.

Issue #2332, opened 2020-03-01
setFieldValue deletes the value from values and prevents errors from showing on submit


amir-virta:
### Current Behavior

When setFieldValue() sets a field to undefined it deletes it from `values`. The consequence of this is that Formik no longer marks the field as touched on submit and therefore error messages for the field no longer show (using `<ErrorMessage>`)

Steps to repro:
1) Formik state after submit is clicked and input field is not touched. As you can see the error message is visible.

![image](https://user-images.githubusercontent.com/38225887/75621864-cf32a180-5b4e-11ea-9c29-33c45723f6d7.png)

2) Formik state after (a) input field is set to a value, then (b) the input field is cleared, and then (c) submit is clicked again. The nuance here is that when the field is cleared, `setFieldValue` is called to set the form field to undefined. You can see the error exists, but the field name is no longer in values and touched does not have the field name either.

![image](https://user-images.githubusercontent.com/38225887/75621868-d8237300-5b4e-11ea-88ad-ea418632c222.png)

### Expected behavior

On submit, Formik should continue to mark the fields set to undefined as touched. 

### Reproducible example

https://codesandbox.io/s/formik-setfieldvalue-deletes-the-value-from-values-and-prevents-errors-from-showing-on-submit-ysxdh?fontsize=14&hidenavigation=1&theme=dark

### Suggested solution(s)

Two ways of this happening (a) set the field value to undefined instead of deleting it from values when `setFieldValue` is called with the value of undefined OR (b) clone initialValues and use that to determine which fields to touch on submit instead of the current set of keys in `values`.

### Additional context

None

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           |  2.1.1 (and 1.1.1) |
| React            | 16.12.0 |
| TypeScript       |  n/a |
| Browser          | Chrome Version 80.0.3987.122 (Official Build) (64-bit) |
| npm/Yarn         | Yarn |
| Operating System | Mac OS X |

justinbhopper:
I agree, formik should be setting the value to undefined, not deleting the property from the values object.  Many operations in Formik rely on the existence of the entry in the values object in order to know what is errored/touched, and by deleting, it is wiping that data.

These lines seem to be the culprits, although there may be others:
https://github.com/jaredpalmer/formik/blob/master/packages/formik/src/utils.ts#L131
https://github.com/jaredpalmer/formik/blob/master/packages/formik/src/utils.ts#L139

[...]

PR #3491, opened 2022-02-13
Remove the deletion of a property in `setIn` with undefined value #3474

### LINKED ISSUE #3474: setFieldValue will delete property with undefined value
### Current Behavior

After `setFieldValue('name', undefined)` values will be `{}`. 

### Expected behavior

After `setFieldValue('name', undefined)` values will be `{ name: undefined }`. 

### Reproducible example

I have 2 dependent fields. The value of one removes the value of the second.

https://codesandbox.io/s/formik-touched-issue-ov3no?file=/src/App.tsx:728-741

steps:
1. change organization_id
2. change person_list_id
3. change organization_id

values will be `{"organization_id": ""}` without person_list_id. 
And after submitting the form `person_list_id` will be not touched.

### Suggested solution(s)

Delete a condition https://github.com/jaredpalmer/formik/blob/master/packages/formik/src/utils.ts#L131

### Additional context

I know I can set null, but I think it's not obvious. I spent a lot of time to figure it out.

### Your environment

| Formik           |      2.2.9
| React            |       17.0.20
| TypeScript       |    4.4.2
| Browser          |
| npm/Yarn         |
| Operating System |

fenyagg:
Remove the deletion of a property in `setIn` with `undefined` value

closes #3474

### DIFF
--- a/packages/formik/src/utils.ts
+++ b/packages/formik/src/utils.ts
@@ -127,11 +127,7 @@ export function setIn(obj: any, path: string, value: any): any {
     return obj;
   }
 
-  if (value === undefined) {
-    delete resVal[pathArray[i]];
-  } else {
-    resVal[pathArray[i]] = value;
-  }
+  resVal[pathArray[i]] = value;
 
   // If the path array has a single element, the loop did not run.
   // Deleting on `resVal` had no effect in this scenario, so we delete on the result instead.


iamn1ck:
Would love to see an update on this bug. Its been a known issue for 2 years now https://github.com/jaredpalmer/formik/issues/2332

avasuro:
It will not work just to delete this lines, because `setIn` function is used to change `formik.errors` and `formik.touched` states (beside `formik.values` state), and it is required to actually remove undefined values from this states (e.g. see [this bug](https://github.com/jaredpalmer/formik/issues/727) for more info)

Need to somehow distinguish when undefined values should be removed, and when should not.

fenyagg:
I am deleting pr because it is not complete

PR #3500, opened 2022-02-26
feat: validateAfterSubmit prop

### LINKED ISSUE #1251: validateOnChange after submit
### Current Behavior

`validateOnChange` is either true or false

### Desired Behavior

`validateOnChange` value can be changed, is not just an initialization config.

### Suggested Solution

`validateOnChange` is a function that has access to the form. Once it's been submitted it validates on change, but not before. (This is what angular does and is quite nice)

All users

Tried modifying the value `onSubmit` but there's no way to access it. Also thought of hiding the errors until submission but didn't find a way to do so.

### LINKED ISSUE #3375: validateOnChange, validateOnBlur allow more complex interactions
### Current Behavior

Currently you define this properties when creating the form, either with the Component or with the hook. Afterwards you have no way of changing this options.

### Desired Behavior

In a real case scenario, I don't care to validate **the whole** form as the user types in the first input. But after the user attempted to submit the form and everything lighted up with errors, if he goes back to an input with an error I do care to validate as he types.

### Suggested Solution

Instead of just allowing for boolean in those two options, allow us to provide a function. This function get's called passing all of the formik's props (similarly to the submit prop that gets called with `(values, formikProps)`

```
  validateOnChange: (formikProps) => formikProps.submitCount > 0,
  validateOnBlur: (formikProps) => formikProps.submitCount > 0 && !formikProps.isValid
```

I think everybody as I believe what I described is the validation behavior you see in most of the internet

I did hacks to achieve this, overwriting the formik functions `handleChange` `handleBlur` but this is quite nasty.

MikeRomaa:
Fixes #1251
Fixes #3375

Adds the `validateAfterSubmit` prop to `<Formik />`. This will prevent form validation until after the first submission attempt, where it will then follow the logic defined by `validateOnBlur` and `validateOnChange`.

### DIFF
--- a/.changeset/pink-falcons-shake.md
+++ b/.changeset/pink-falcons-shake.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
+
+Added `validateAfterSubmit` prop to prevent form validation until the first submission attempt.
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -451,6 +451,11 @@ events and `change`-related methods. More specifically, when either
 Default is `false`. Use this option to tell Formik to run validations when the `<Formik />` component mounts
 and/or `initialValues` change.
 
+### `validateAfterSubmit?: boolean`
+
+Default is `false`. Use this option to tell Formik to not run validations before the first attempt to submit the `<Form />`.
+After submission, validation logic will follow the rules set by `validateOnBlur` and `validateOnChange`.
+
 ### `validationSchema?: Schema | (() => Schema)`
 
 [A Yup schema](https://github.com/jquense/yup) or a function that returns a Yup
--- a/docs/guides/validation.md
+++ b/docs/guides/validation.md
@@ -252,7 +252,7 @@ export const FieldLevelValidationExample = () => (
 
 ## When Does Validation Run?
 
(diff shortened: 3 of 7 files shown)


MikeRomaa:
bump @jaredpalmer

stee-ren:
@jaredpalmer I'd really love to see this merged

dmmarmol:
> @jaredpalmer I'd really love to see this merged

Had to upvote and quote this because it's 100% what I need right now

JaviCeRodriguez:
I need this! The `validateAfterSubmit` prop replace all solutions mentioned by the users in #1251. In particular, I have this case:

```tsx
const [isSubmit, setIsSubmit] = useState(false);

return (
  <Formik
    validateOnChange={isSubmit}
    // rest of Formik props
  >
    {({ handleSubmit, isSubmitting }) => {
      if (isSubmitting && !isSubmit) setIsSubmit(true);

      // rest of form
  </Formik>
)
```

But this raise a warning on console:

> Warning: Cannot update a component (`ContactForm`) while rendering a different component (`Formik`). To locate the bad setState() call inside `Formik`, follow the stack trace as described in [...]

This PR resolve this, and we should not force the value of `validateOnChange` anymore.

d4rky-pl:
bump

vbarzola:
Bumping this for visibility.

PR #3555, opened 2022-04-27
Trigger validation on Pop of array-of-objects

fenech:
Make a copy of the original array, so that validation happens the same
as when calling Remove on the last element.

Test case: https://codesandbox.io/s/formik-fieldarray-change-validation-bug-forked-fddidy?file=/index.js
note that after adding an element, clicking on "pop" doesn't clear the error, whereas clicking on its "x" button does.

This issue affects arrays of objects in the schema. Returning a copy of the old array, instead of the original, means that validation is triggered and the errors object is correctly updated.


probablyup:
Fixed in recent releases

PR #3698, opened 2022-12-09
improved typescript support for ArrayHelpers

### LINKED ISSUE #3641: TypeScript Support for `ArrayField`
## Feature request
Hello everyone, 

### Current Behavior
As far as I looked, `ArrayField` component doesn't have TypeScript support for its array utils.

image

### Desired Behavior
Have type safety on array utils

### Suggested Solution

We can supply type safety by passing type of the array into `ArrayField` like:

```tsx
<FieldArray<Fact> name="facts">
  {children}
</FieldArray>
```

It will be the users who use TypeScript. It will help us to sure all data that we use is type-safed.

I have no alternatives. Open to discuss. 

### Additional context
I don't know Formik codebase deeply therefore I don't have enough information about thats feature can be done or not. But, I think that it will be good feature for Formik community. I would like to handle it as possible as I can

amagana8:
Added typescript generics to `ArrayHelpers` interface and its methods so that users who use TypeScript can set the type for their arrays and have type safety on array utils. I have also gone ahead and made supplying a type for the generic optional for the sake of backwards compatibility so any existing TS code that does not give a type for the FieldArray will continue to work as they always have.

resolves #3641

### DIFF
--- a/packages/formik/src/FieldArray.tsx
+++ b/packages/formik/src/FieldArray.tsx
@@ -27,11 +27,11 @@ export type FieldArrayConfig = {
   /** Should field array validate the form AFTER array updates/changes? */
   validateOnChange?: boolean;
 } & SharedRenderProps<FieldArrayRenderProps>;
-export interface ArrayHelpers {
+export interface ArrayHelpers<T = void> {
   /** Imperatively add a value to the end of an array */
-  push: (obj: any) => void;
+  push: (obj: T) => void;
   /** Curried fn to add a value to the end of an array */
-  handlePush: (obj: any) => () => void;
+  handlePush: (obj: T) => () => void;
   /** Imperatively swap two values in an array */
   swap: (indexA: number, indexB: number) => void;
   /** Curried fn to swap two values in an array */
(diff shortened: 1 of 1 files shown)


packages/formik/src/FieldArray.tsx
```diff
@@ -27,11 +27,11 @@ export type FieldArrayConfig = {
   /** Should field array validate the form AFTER array updates/changes? */
   validateOnChange?: boolean;
 } & SharedRenderProps<FieldArrayRenderProps>;
-export interface ArrayHelpers {
+export interface ArrayHelpers<T = void> {
```
alperen, line 30: wouldn't it be more type-safe, if it had been `unknown[]`?

alperen (commented):
Hello, thanks for your collaboration, I've seen your PR when I was looking for my issues on Github. That's going to make Formik codebase more type-safe and (IMO) it is a needed feature.

But I have an opinion what do you think?

packages/formik/src/FieldArray.tsx
alperen:
```suggestion
export interface ArrayHelpers<T = unknown[]> {
```

amagana8: Yeah that's a great point, I'll update my PR. Thanks!

PR #3752, opened 2023-04-10
types: improve handlers type

chernodub:
Hi, I noticed that when using `getFieldMeta` and `getFieldHelpers`, IDE does not provide the list of the names of the form fields. So I made a **non-breaking** update to `FormikHandlers` interface.

This way we are able to leverage typed argument in `getFieldMeta` and `getFieldHelpers` to (1) avoid possible bugs after refactoring and (2) get a better DX when working with formik.

Example:
Screen Shot 2023-04-10 at 10 12 05

Please let me know if more updates or clarifications needed. Thanks!

### DIFF
--- a/examples/basic-typescript-2/index.html
+++ b/examples/basic-typescript-2/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Playground</title>
+  </head>
+
+  <body>
+    <div id="root"></div>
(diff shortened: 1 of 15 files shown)


probablyup:
Good start! These APIs allow for `deep.path.key` notation, so we'll need a utility type to extract all the deep object keys versus just the surface-level ones. Something like this could work: https://dev.to/pffigueiredo/typescript-utility-keyof-nested-object-2pa3

Pretty much everywhere we take a field `name` would benefit from this.

chernodub:
Hi @probablyup! I was finally able to find some time to work on this and I improved the overall implementation. Now all of the fields are strictly-typed. I also added some tests for this. Feel free to play around in a new example project I created.

Screenshot 2023-07-09 at 15 30 22

chernodub:
Seems that the repo is inactive, I'm closing the PR

Issue #3742, opened 2023-03-16
TypeScript type shorthands for setFieldValue and setFieldTouched


mathieucaroff:
### Current Behavior

Currently, the types of `setFieldValue`, `setFieldError` and `setFieldTouched` are the following:

```ts
export interface FormikHelpers<Values> {
  setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void
  setFieldError: (field: string, message: string | undefined) => void
  setFieldTouched: (field: string, isTouched?: boolean, shouldValidate?: boolean) => void
  // ...
}
```

### Desired Behavior

In TypeScript, I'd like to be able to call pass the `setFieldValue` and `setFieldTouched` functions as parameter of other functions without having to write such lengthy types.

### Suggested Solution

Introduce a dedicated type for these three functions. I'm not sure what naming would be best, but here's my best attempt:

```ts
export type SetFieldValue = (field: string, value: any, shouldValidate?: boolean) => void
export type SetFieldError = (field: string, message: string | undefined) => void
export type SetFieldTouched = (field: string, isTouched?: boolean, shouldValidate?: boolean) => void
```

### Who does this impact? Who is this for?

This will make typing easier for TypeScript users.

### Alternatives

Until this feature is adopted, Formik TypeScript users can have their own type definition somewhere in their codebase for these three functions.

pachuka:
Another alternative is to just reference the internal type:

```tsx
FormikHelpers<REPLACEME>['setFieldError'];
```

This results in: 
![image](https://user-images.githubusercontent.com/3280602/232829445-e4003322-cc67-46d2-89c6-d15854bf3cd2.png)

You can then use that directly rather than creating any custom types.

samirant15:
I implemented a small function as a solution for this issue, take a look here: https://github.com/jaredpalmer/formik/issues/1388#issuecomment-1969302106

PR #3755, opened 2023-04-13
Add SetField* type shorthands

mathieucaroff:
This MR addresses #3742

It adds type alias for the three functions of the FormikHelpers interface which do not depend on the <Value> type parameter and which accept at least two arguments. These alias are intended to ease user-side typing.

The modified file has been formatted by prettier.


probablyup:
@mathieucaroff can you add a changeset for this?

Issue #3546, opened 2022-04-15
Namespace 'React' has no exported member 'StatelessComponent'.


alaanescobedo:
## Bug report

The StatelessComponent type was removed in react update 18 - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210

### Current Behavior
Error when building app :
`node_modules/formik/dist/withFormik.d.ts(58,77): error TS2694: Namespace 'React' has no exported member 'StatelessComponent'.`

### Expected behavior
Build without errors.

### Reproducible example
Following the formik documentation - https://formik.org/docs/guides/typescript
https://codesandbox.io/s/hardcore-bhabha-kobk3m?file=/src/MyForm.tsx
Deploy sandbox to preview

### Suggested solution(s)
I don't know how to replicate it in codesandbox, but in a local environment in 
`node_modules/formik/dist/withFormik.d.ts` on line 58 replace` React.StatelessComponent<P>` by `React.FunctionComponent<P>`

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.2.9
| React            | 18.0.0
| TypeScript       | 4.6.3
| Browser          |  Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36
| npm/Yarn         | yarn
| Operating System | Windows

maximeg:
In a fixes.d.ts I added

```ts
/*
 * Namespace 'React' has no exported member 'StatelessComponent'
 * in formik, react-mapbox-gl
 */
declare namespace React {
  type StatelessComponent<P> = React.FunctionComponent<P>;
}

```

timmaloshtan:
@maximeg Where did you put the file `fixes.d.ts`?

johnrom:
@timmaloshtan you can declare your own fixes anywhere, but these patches are often placed outside of `src`, like:

```
/your-project
  /src
    /your-form.tsx
  /types
    /fixes.d.ts
```

Then, you would modify your `tsconfig.json`:

```json
{
  "include": ["src", "types"],
}
```

This helps you keep the patches outside of your source codebase so you can examine them often.

bennycode:
In my app I created the file `src/@types/index.d.ts` with the following content to have a workaround (thanks to @maximeg):

```ts
// Fixes TS2694
declare global {
  namespace React {
[...]

PR #3772, opened 2023-05-06
Update withFormik.tsx

ravshansbox:
React 18 does not contain StatelessComponent type anymore.

### DIFF
--- a/packages/formik/src/withFormik.tsx
+++ b/packages/formik/src/withFormik.tsx
@@ -81,7 +81,7 @@ export interface WithFormikConfig<
 
 export type CompositeComponent<P> =
   | React.ComponentClass<P>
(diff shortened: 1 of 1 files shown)


Philipp91:
While I can't comment on these CI failures, I'm generally in favor of this change. Currently, whenever I `npm install`, I need to manually open up `withFormik.d.ts` and make this change to get things to work. The workarounds in #3546 mess things up for me (making TypeScript believe that `React` now _only_ contains `StatelessComponent` and nothing else, leading to millions of errors), so I'd appreciate this proper fix here being released.

ravshansbox:
@Philipp91 you can work around this issue adding `skipLibCheck: true` to `tsconfig.json`

PR #3784, opened 2023-05-26
perf(FieldArray): add shouldComponentUpdate to cut down on unnecessary renders

probablyup:
Closes #3295

### DIFF
--- a/.changeset/afraid-kangaroos-cheer.md
+++ b/.changeset/afraid-kangaroos-cheer.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
+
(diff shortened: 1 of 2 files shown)


jaredpalmer (approved):

JannesMeyer:
@jaredpalmer @probablyup maybe this should've been a breaking change?

It broke our code because we have local state (a boolean popup toggle) in the `FieldArray` children.

probablyup:
> @jaredpalmer @probablyup maybe this should've been a breaking change?
> 
> It broke our code because we have local state (a boolean popup toggle) that affects the `FieldArray` children.

If you could provide a failing test/reproduction to work from in a new issue and tag me that would be appreciated

xconverge:
I also experienced breakage. 

Adding this for now restored behavior.
``` shouldUpdate={() => true}```

I will look at the logic a bit more in this PR to see if I can find a clue

PR #3796, opened 2023-05-26
feat(validationSchema): support yup transforms in validate & submit

### LINKED ISSUE #608: Feature & Question: Hook for manual transformations

probablyup:
Closes #728 
Closes #608

Add support for Yup ["transforms"](https://github.com/jquense/yup#parsing-transforms).

I do not believe this is a semver-major change as it's an optional Yup feature; it's actually probably confusing that it _hasn't_ worked to date.

Original credit for this work belongs to @sshmyg, I just modernized it and added tests.

### DIFF
--- a/.changeset/fluffy-rivers-rest.md
+++ b/.changeset/fluffy-rivers-rest.md
@@ -1,0 +1,5 @@
+---
+'formik': minor
+---
+
+Add support for Yup ["transforms"](https://github.com/jquense/yup#parsing-transforms).
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -218,10 +218,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
    */
   const runValidationSchema = React.useCallback(
     (values: Values, field?: string): Promise<FormikErrors<Values>> => {
-      const validationSchema = props.validationSchema;
-      const schema = isFunction(validationSchema)
-        ? validationSchema(field)
-        : validationSchema;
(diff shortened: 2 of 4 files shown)


jaredpalmer (approved):

rdenman:
FYI this probably would have been a good major release. We had breaking changes after bumping to the latest version of formik over the weekend (we use Renovate to auto-update).

We had a yup validation schema that called `lowercase` before checking for an existing name to prevent duplicates. Previously, the api call wouldn't send that transformed version in the body, but with latest, all our requests sent the field as lowercase. Not a huge issue, but I'd consider it breaking

Issue #3807, opened 2023-05-29
The type of `setFieldValue` seems incorrect.


NagaiKoki:
## Bug report
The following Eslint error appeared in the `setFieldValue` function when updating to [version 2.3.2](https://github.com/jaredpalmer/formik/releases/tag/formik%402.3.2) of formik.

````
Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead.eslint@typescript-eslint/ unbound-method
````

When I look at the `FormikHelpers` types, though, they were defined as follows,
```ts
setFieldValue(field: string, value: any, shouldValidate?: boolean): Promise<void | FormikErrors<Values>>;.
```

However, looking at the implementation, I think the correct type is

```ts
setFieldValue: (field: string, value: any, shouldValidate?: boolean) => Promise<void | FormikErrors<Values>>;
````

Is this type correct?

### Current Behavior
![スクリーンショット 2023-05-29 14 32 11](https://github.com/jaredpalmer/formik/assets/50698194/731f7f04-0282-457c-855b-ae2b4f54a093)

```ts
export interface FormikHelpers<Values> {
    // ...
    setFieldValue(field: string, value: any, shouldValidate?: boolean): Promise<void | FormikErrors<Values>>;
     // ...
}
```

### Expected behavior

```ts
export interface FormikHelpers<Values> {
    // ...
    setFieldValue: (field: string, value: any, shouldValidate?: boolean) => Promise<void | FormikErrors<Values>>;
  // ...
}
```

### Reproducible example
* access `formik.setFieldError` and eslint [@typescript-eslint/unbound-method](https://typescript-eslint.io/rules/unbound-method) get enabled

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.4.0
| React            | 18.2.0
| TypeScript       | 5.0.0
| Browser          |
| npm/Yarn         | yarn
| Operating System |

probablyup:
We can fix this, but it's probably also better to not check types that aren't yours with https://www.typescriptlang.org/tsconfig#skipLibCheck

klarstrup:
I face this issue with `skipLibCheck` enabled.

probablyup:
Fixed in https://github.com/jaredpalmer/formik/releases/tag/formik%402.4.1, thanks for contributing!

NagaiKoki:
@probablyup Thank you!

PR #3808, opened 2023-05-29
fix type of `setFieldValue `

NagaiKoki:
issue is https://github.com/jaredpalmer/formik/issues/3807

I fixed `setFieldValue` type because it seems to be incorrect.

### DIFF
--- a/.changeset/lemon-zoos-attend.md
+++ b/.changeset/lemon-zoos-attend.md
@@ -1,0 +1,5 @@
+---
+'formik': patch
+---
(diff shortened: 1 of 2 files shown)


probablyup (approved):

probablyup:
Can you add a changeset please for the fix?

jaredpalmer:
Should probably be a patch since its a bugfix

probablyup (approved):
Changes LGTM

PR #3813, opened 2023-05-31
followup from prior releases

probablyup:


jaredpalmer (approved):

packages/formik/src/FieldArray.tsx
```diff
@@ -155,26 +153,6 @@ class FieldArrayInner<Values = {}> extends React.Component<
     this.pop = this.pop.bind(this) as any;
   }
 
-  shouldComponentUpdate(props: any) {
```
hanayashiki, line 158: Did you intend to delete this? (although this is actually breaking my code when a component returned by the `render` function depends on local state.)

probablyup: I intended to revert the change yes.

PR #3817, opened 2023-05-31
refactor: react 18 internal upgrade

probablyup:
`useReducer().dispatch` appears to be async now where it used to be synchronous, leading to many broken tests. I have replaced usage of it with equivalent synchronous code.


jaredpalmer (approved):

litinskii:
image
start got ts error after this update
on 2.4.1 this not happend

PR #3825, opened 2023-06-12
FEATURE: Add optional Id prop on ErrorMessageProps type

avrlsngh:
- Additional optional `id` prop added to `ErrorMessageProps` to make it easily testable in e2e tests. 
- Updated the docs of the section ErrorMessage and added an example to demonstrate this update.
- Since it doesn't change the functionality, it didn't need to be covered under any test case.

**NOTE:** This change won't affect any functionalities and since it is an optional param, it can be omitted if not needed.

**This is my first contribution to this project, please let me know if I missed something..**

### DIFF
--- a/packages/formik/src/ErrorMessage.tsx
+++ b/packages/formik/src/ErrorMessage.tsx
@@ -4,6 +4,7 @@ import { getIn, isFunction } from './utils';
 import { connect } from './connect';
 
 export interface ErrorMessageProps {
+  id?: string;
   name: string;
   className?: string;
   component?: string | React.ComponentType;
--- a/docs/api/errormessage.md
+++ b/docs/api/errormessage.md
@@ -97,6 +97,17 @@ Either a React component or the name of an HTML element to render. If not specif
 // --> {touched.email && error.email ? error.email : null}
 ```
 
+### `id`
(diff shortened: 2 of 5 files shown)


packages/formik/src/ErrorMessage.tsx
```diff
@@ -4,6 +4,7 @@ import { getIn, isFunction } from './utils';
 import { connect } from './connect';
 
 export interface ErrorMessageProps {
+  id?: string;
```
probablyup, line 7: Should we just allow all typical `div` attributes and pull off our specific ones during render?

probablyup: @jaredpalmer

probablyup:
```ts
export interface ErrorMessageProps extends React.HTMLAttributes<HTMLDivElement> {
  component?: string | React.ComponentType;
  children?: (errorMessage: string) => React.ReactNode;
  name: string;
  render?: (errorMessage: string) => React.ReactNode;
}
```

Not sure if there's a native-specific version of this?

probablyup (approved):
Thank you!

Issue #3827, opened 2023-06-13
broken website links


adjenks:
These links are broken in the docs:

Page: All pages
Link Text: Edit this page on GitHub
href: https://github.com/formik/formik/edit/master/docs/api/field.md

Page: https://formik.org/docs/api/field
Link Text: FieldInputProps
href: https://formik.org/docs/api/field#fieldinputprops

Page: https://formik.org/docs/api/field
Link Text: FieldMetaProps
href: https://formik.org/docs/api/field#fieldmetaprops

probablyup:
Care to submit a fix PR? The first one is probably because the primary branch is `main` now

Second one should now point to → https://formik.org/docs/api/useField#fieldinputpropsvalue
Third one should now point to → https://formik.org/docs/api/useField#fieldmetapropsvalue

shouldnt:
can I submit a PR for this?

probablyup:
Please do

PR #3830, opened 2023-06-18
fix(docs): #3827 link ref to other doc pages is incorrect

shouldnt:


docs/api/formik.md
```diff
@@ -93,11 +93,11 @@ to `<button onClick={handleReset}>...</button>`
 
 #### `handleSubmit: (e: React.FormEvent<HTMLFormElement>) => void`
 
-Submit handler. This should be passed to `<form onSubmit={props.handleSubmit}>...</form>`. To learn more about the submission process, see [Form Submission](../docs/guides/form-submission.md).
```
probablyup, line 96: I don't think these changes are correct, they work properly on the current live site

shouldnt: Yes, I checked the live site is correct, but when navigating around directly in the md file in the GitHub repo will be incorrect
this change makes both md file and the website work properly

docs/api/fieldarray.md
```diff
@@ -67,7 +67,7 @@ export const FriendList = () => (
 
 ### `name: string`
 
-The name or path to the relevant key in [`values`](./formik.md#values--field-string-any-).
+The name or path to the relevant key in [`values`](./formik#values--field-string-any-).
```
shouldnt, line 70: I didn't remote .md extension because, in the website src code, the extension will be remote before transforming to a link
and keep .md extension will help navigate around directly in the md file correctly.

probablyup (dismissed):
LGTM thank you

probablyup (approved):

PR #3832, opened 2023-06-20
chore: yarn 3

probablyup:


website/package.json
```diff
@@ -45,7 +45,6 @@
     "rehype-img-size": "^0.0.1",
     "rehype-raw": "^5.0.0",
     "rehype-sanitize": "^4.0.0",
-    "rehype-shiki": "^0.0.9",
```
probablyup, line 48: Not used and was causing issues

website/src/index.d.ts
```diff
@@ -1,11 +1,11 @@
+declare module 'gray-matter';
```
probablyup, line 1: Missing stub

jaredpalmer:
We should move to pnpm but this is fine for short term

probablyup:
> We should move to pnpm but this is fine for short term

I'd actually say bun if we intended to move off yarn

PR #3852, opened 2023-08-01
Feature: Pass Yup context to validationSchema using new prop validationSchemaContext in <Formik>

john-raymon:


mikegreiling:
What can be done to move this PR forward? I could really use this feature and I'd be happy to contribute, but it looks like this has gone 8 months with no feedback from code owners :(

Cbridger-mywellabee:
I just ripped out RHF and replaced it with Formik because RHF doesn't support errors being raised on fields that weren't touched (although it supported contexts!), only to find out Formik doesn't currently support Yup Contexts 😅, I would love it if this PR could gain some momentum!

yogiduzit:
+1, would appreciate this PR's prioritization.

rgallison:
Chiming in as this is an issue my team faces regularly

higgsboz-beam:
+1 on this being a useful feature to me and my team

eliasthompson:
+1, Would also love this to be merged for all reasons discussed.

Issue #3861, opened 2023-08-09
Missing `props.onReset` dependency in useCallback of resetForm function.


rajpatelbot:
### Current Behavior
If we have alternative thing to reset like search and filter, so suppose i have a object which have search and filter key and at first the i searched something and then i do filter and now after filter i reset the search value and then i reset the filter then the automatically the previous searched value comes in search and again i have to press on reset search button to clear it.

### Expected behavior

I am expecting that after clicking the resetFilter button the search should not take the previously searched value.

### Suggested solution(s)
In the resetForm function the useCallback should have dependency `props.onReset`.

### Additional context

Reference line: https://github.com/jaredpalmer/formik/blob/main/packages/formik/src/Formik.tsx#L420

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.4.2
| React            | 18.2.0
| TypeScript       | 5.0.2
| Browser          | Chrome
| npm/Yarn         | npm
| Operating System | Windows 10

rajpatelbot:
This issue is not fixed still same behaviour happens.

PR #3872, opened 2023-08-31
FIX: Fixed resetForm function dependency issue

rajpatelbot:
Closes: #3861

As i mentioned in the issue section that the unexpected behaviour is comming while using resetForm function due to missing dependency. Here is the fix.


probablyup (approved):
Makes sense ty

Issue #3874, opened 2023-09-05
passing all values as the second parameter when calling Field's validate


myNameIsDu:
## Feature request

passing all values as the second parameter when calling Field's validate

### Current Behavior
just passes the current value as the first parameter

### Desired Behavior

passing all values as the second parameter when calling Field's validate

### Suggested Solution

passing all values as the second parameter when calling Field's validate

### Describe alternatives you've considered

I have tried many alternative solutions, but I haven't found a suitable one.

### Additional context

In the context of my application, there's a dynamic form, For example, it's a radio group. When I select the 'A' option, it triggers the rendering of an additional input or other components. Once rendered, this component is automatically registered and set as 'required'. On the other hand, if I opt for a selection other than 'A', different fields and components get registered. So that's the reason why I'm using Field's validate, I don't have to operate the huge schema, The Field'validae comes with automatic registration and logout

Here's the smallest reproducible example: https://stackblitz.com/edit/stackblitz-starters-pxyrgo?file=src%2FApp.tsx. 
To reproduce the issue, follow these steps:

1. Click the 'Add' button.
2. Enter "1" into the second input.
3. Check the console.

You'll notice that the values in the validate function are stale.

Finally, I'd like to thank you for your outstanding work in creating this invaluable form library. Regarding this issue, if you agree to add this feature, I'm more than willing to implement it

PR #3882, opened 2023-09-12
feat: passing all values as the second parameter when calling Field's validate

myNameIsDu:
ref: #3874


probablyup:
Generally in favor, could you adjust the documentation and add a changeset? Thanks

myNameIsDu:
I've already adjusted the Field API document, but I'm not sure if there are other places I need to adjust. If there are, please let me know.

Issue #3883, opened 2023-09-12
2.4.4 is a Breaking Change -  passing "className" this new way overwrites existing "className" in composed components


shaunaas:
### Current Behavior

2.4.4 unexpectedly broke all of my fields.
The way the code is implemented, className becomes required and overwrites all my classNames that exist on my composed components being passed into <Field component={<MyCustomComponent />} />

### Expected behavior

className does not become a required prop

### Reproducible example

`
const CustomInput = (props) => {
  return (
    <div>
      <input type="text" className="composed-component-classname" {...props} />
    </div>
  );
};

<Field
  name="name"
  component={<CustomInput/>}
/>

`

### Suggested solution(s)

1. Revert this change. Passing "className" was already supported via ...rest props and a component that either spreads/or specifically requires any specific props to be passed.

### Additional context

- passing className after the ...rest in the formik lib will overwrite any existing className prop already on the <input/>

[](https://github.com/jaredpalmer/formik/commit/41720c2f69407e41c27b325923bce63436b07f45)

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.4.4
| React            | 18.2.0
| TypeScript       | N.A.
| Browser          | Any
| Yarn         | 1.22.19
| Operating System |  macOS 13.5.1 (Ventura)

myNameIsDu:
This #3862  caused this problem. 

But I think you can make your component safer and clear. like this

```tsx
const CustomInput = ({propsA, propsB}) => { return ( <div> <input type="text" className="composed-component-classname"   propsA={propsA}  propsB={propsB}/> </div> ); };
```

Probably you have written `props type` to make your code look clearer, but either way {...props} is not a safe style of writing,  because if Formik add a `type` props one day, it will break your code

shaunaas:
The above is an example for simplicity.. not actual code.

1 - A breaking change should not be put into a minor release.
2 - Spread props are explicitly intended to work exactly like the above example. The lib should have a clear API - as consumers of the library may have many different use cases. There is no reason to add "type" or "className" as they were already covered by the spread ...rest props.

myNameIsDu:
> The above is an example for simplicity.. not actual code.
> 
> 1 - A breaking change should not be put into a minor release. 2 - Spread props are explicitly intended to work exactly like the above example. The lib should have a clear API - as consumers of the library may have many different use cases. There is no reason to add "type" or "className" as they were already covered by the spread ...rest props.

[...]

Issue #3858, opened 2023-08-03
className Issues


CEN-Smart:
## Bug report

Please, the `<Field/>` component should have `className` extented as props to it, this will enable it have a custom styling for developers.

yazaldefilimonepinto:
hi, I resolve that in #3862

yazaldefilimonepinto:
@CEN-Smart we can close this issue?

PR #3884, opened 2023-09-14
fix: fix Field allways passes a undefined className to custom component

myNameIsDu:
Ref https://github.com/jaredpalmer/formik/issues/3883

This issue is caused by this [PR](https://github.com/jaredpalmer/formik/pull/3862), The original issue can be found [here](https://github.com/jaredpalmer/formik/issues/3858)

There are some explanations

1. [PR](https://github.com/jaredpalmer/formik/pull/3862) is a little weird, because the `className` prop was already included by `props`, and this style of writing (extracting it from `props` and then passing it into children's props alone) will cause there's always a className in children's props
2. The [original issue](https://github.com/jaredpalmer/formik/issues/3858) is confusing, because the className is already passed, except that there's no declaration.

Then there are three revises in this pr

1. Delete extracted className, but keep the type declaration
2. Add test cases
3. Fix the wrong usage of `cases` function in the test cases before by the way


colin-oos:
Any update on this?

colin-oos:
Bumping this again. This really needs to get merged. Currently you are not able to style custom components because className on the custom component is always undefined.

PR #3896, opened 2023-10-04
Fix-field-typing

qwezey:


qwezey:
@jaredpalmer @probablyup is this something we can get out? this fix makes working with the `Field` component in typescript much easier

Zikoat:
There are some issues with this, namely that custom props which are required by a custom component cannot be passed to Field because the props type to field doesnt include the props from the custom component. We will need another generic for this or change the structure. I will take a crack at this, but I have to discuss it in a new issue so Jared gives a thumbs up before I spend a lot of time on it.

PR #3919, opened 2023-11-23
Update useField.md

shuryhin-oleksandr:
Unused `helpers` value. It should be either deleted, either example demonstrating how to use it should be provided.


rushmata (approved):
GTG!

quantizor (approved):
Fair enough, ty

PR #3930, opened 2023-12-18
Included Techjobs to user list

CharlesDM:
We're currently using Formik for some of our internal tools. Works well, and we are happy users. :)


quantizor (approved):
Glad to hear it :)

PR #3231, opened 2021-06-04
Formik v3 Umbrella PR.

johnrom:
Todo:

- [ ] Merge main
- [ ] Documentation

PR #3931, opened 2023-12-22
feat: Added ability to provide patching function to setStatus method

dan-irf:
Problem
When setting status, sometimes user needs to update the existing value. When used inside a useCallback hook for instance, it introduces an unnecessary variable to the dependency list.
Solution
setStatus now acepts a patching function

### DIFF
--- a/docs/api/formik.md
+++ b/docs/api/formik.md
@@ -200,7 +200,7 @@ Set the value of a field imperatively. `field` should match the key of
 
 If `validateOnChange` is set to `true` and there are errors, they will be resolved in the returned `Promise`.
 
-#### `setStatus: (status?: any) => void`
+#### `setStatus: (status?: React.SetStateAction<any>) => void`
 
 Set a top-level `status` to anything you want imperatively. Useful for
 controlling arbitrary top-level state related to your form. For example, you can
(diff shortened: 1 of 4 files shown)


dan-irf:
@quantizor  @johnrom  @jaredpalmer just wanted to ask you for PR review explicitly

quantizor (approved):
Can you add a patch changeset please?

johnrom:
I'm not sure what the purpose of this change is. Are you creating a new status based on the last status? It's not really clear by default where this status comes from (render or state), so I would expect Formik (if it is being brought back from the dead) to move in a direction where the source of this value can be determined explicitly, either within a render using `useFormikContext().status`, or the absolute current value which may be "ahead" of the render, using a `getState()`-like API, similar to redux.

```tsx
const { setStatus, getState, status } = useFormikContext();

useEffect(() => { setStatus( getState().status === "who" ? "what" : "where" }, [setStatus, getState]);
```

You can see what that API could look like here: [#3231](https://github.com/jaredpalmer/formik/pull/3231/files#diff-2421f0a25f33b0ee81ca1de9dee2a1743711e08f273a5733ed03493de527de21R183-R203)

PR #3937, opened 2024-01-10
feat: add submission failed handler

amazingdudu:
Provide a handler when form validation fails。


quantizor (commented):
Nice addition, can you add at least one unit test and a minor changeset for this?

Issue #2083, opened 2019-12-02
[v2] Validation runs on old values after setFieldTouched


bkiac:
### Current Behavior

Calling `setFieldTouched` runs validation on old values.

### Expected behavior

Validation should be called on new values.

### Reproducible example

https://codesandbox.io/s/formik-codesandbox-template-yqrgc

I can get around this issue with replacing
```javascript
props.setFieldValue("name", "John");
props.setFieldTouched("name", true);
```
with
```javascript
props.setFieldValue("name", "John");
// Set `shouldValidate` to `false` to prevent validation
props.setFieldTouched("name", true, false); 
// Call validation with the new values
props.validateField(name);
```

### Additional context

Related problems I could find: #1977, #2025 

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik           | 2.0.6 |
| React            | 16.12.0 |
| Browser          | Google Chrome 78 |
| npm/Yarn         | Yarn 1.19.0 |
| Operating System | Fedora 31 |

kirjai:
Not sure if the #2116 should have closed this issue or not, but i believe this is still a problem in `2.1.1`

> Reproducible example
https://codesandbox.io/s/formik-codesandbox-template-yqrgc

Upgrading `formik` to `2.1.1` in the reproducible example in the OP still shows the same behaviour

jaredpalmer:
Both setXXX will are called synchronously in your example. setFieldTouched thus doesn’t wait for setFieldValue. Because hooks are annoying af, there is no way for us to provide a promise or callback after the commit AFAIK. Thus my suggestion is then to call setFieldValue first and let it run validation as side effect and then call setFieldTouched but abort validation.

jaredpalmer:
My point above is that this _isn’t_ a bug, it’s how React works.

bkiac:
> Both setXXX will are called synchronously in your example. setFieldTouched thus doesn’t wait for setFieldValue. Because hooks are annoying af, there is no way for us to provide a promise or callback after the commit AFAIK. Thus my suggestion is then to call setFieldValue first and let it run validation as side effect and then call setFieldTouched but abort validation.

Thanks for the explanation!

straiki:
> My point above is that this _isn’t_ a bug, it’s how React works.

I understand that, but isn't there way, how to avoid this behavior if `setFieldValue` internally sets field is touched (maybe optionally)?

[...]

PR #3947, opened 2024-01-25
Fix validation on stale values

Jestermaxrko:
Fixes issue  - https://github.com/jaredpalmer/formik/issues/2083 

The cause of the issue is in this line https://github.com/jaredpalmer/formik/blob/0f960aaeeb0bdbef8312b5107cd3374884a0e62b/packages/formik/src/Formik.tsx#L185

beause of assigning value from `stateRef.current` to the `state` variable this variable contains "old" value until next re-render
https://github.com/jaredpalmer/formik/blob/0f960aaeeb0bdbef8312b5107cd3374884a0e62b/packages/formik/src/Formik.tsx#L190

And then when `setFieldValue` and `setFieldTouched` are called consecutively `stateRef.current` references to updated value imediatelly after `setFieldValue` call, but `state` variable is still references to the old value and `setFieldTouched` runs validation on that old value

Using `stateRef.current` directly gurantees that actual values are taken

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -418,7 +418,12 @@ export function useFormik<Values extends FormikValues = FormikValues>({
         dispatchFn();
       }
     },
-    [props.initialErrors, props.initialStatus, props.initialTouched, props.onReset]
+    [
+      props.initialErrors,
+      props.initialStatus,
+      props.initialTouched,
+      props.onReset,
+    ]
   );
 
   React.useEffect(() => {
@@ -549,7 +554,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       const willValidate =
         shouldValidate === undefined ? validateOnBlur : shouldValidate;
       return willValidate
-        ? validateFormWithHighPriority(state.values)
+        ? validateFormWithHighPriority(stateRef.current.values)
         : Promise.resolve();
     }
   );
@@ -560,7 +565,9 @@ export function useFormik<Values extends FormikValues = FormikValues>({
 
   const setValues = useEventCallback(
     (values: React.SetStateAction<Values>, shouldValidate?: boolean) => {
-      const resolvedValues = isFunction(values) ? values(state.values) : values;
+      const resolvedValues = isFunction(values)
+        ? values(stateRef.current.values)
+        : values;
 
       dispatch({ type: 'SET_VALUES', payload: resolvedValues });
       const willValidate =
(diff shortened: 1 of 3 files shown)


tim-phillips (approved):

ypk4:
Helpful change. Are we pending one more review to get this merged?

pvcresin:
Thank you very much. I hope it will be released.

quantizor:
Would love a test please

packages/formik/src/Formik.tsx
```diff
@@ -548,7 +553,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
       const willValidate =
         shouldValidate === undefined ? validateOnBlur : shouldValidate;
       return willValidate
-        ? validateFormWithHighPriority(state.values)
```
quantizor, line 551: This is actually the exact same thing. If you look at L190

```
const state = stateRef.current;
```

gone-skiing:
Any chance this can get merged and released? Seems like a helpful change to keep on the shelf.

Jestermaxrko:
@quantizor @jaredpalmer I've updated this branch, also added test. Could you take a look ?

Christine-Pinto:
Hey! I'm experimenting with an AI-powered QA tool that reviews PRs for edge cases and missed      
  patterns. I ran it against this PR and thought the findings might be useful.                      
                                                                                                                                       
 
## Summary                                                                                        
                                                                                                    
  The PR correctly identifies the root cause: `state.values` is captured at render time and becomes 
  stale when multiple state updates happen in the same tick. The fix to use `stateRef.current` in   
  `setFieldTouched` is the right approach.                                                          
                                                                                                    
  However, the tool found the **same stale-state pattern in other methods** that weren't updated:   
                                                                                                    
  ## Findings                                                                                       
                                                                                                    
  | Location | Line | Issue |                                                                       
  |----------|------|-------|                                                                       
  | `setFieldValue` (function updaters) | 646 | Uses `state.values` instead of                      
  `stateRef.current.values` |                                                                       
  | `validateField` | 567, 583 | Uses `state.values` for validation lookup |                        
  | `validateForm` default param | 531 | Default `values = state.values` captures stale state |     
                                                                                                    
  ## Impact                                                                                         
                                                                                                    
  If a user calls:                                                                                  
                                                                                        
```
  setFieldValue('a', prev => prev + 1);                                                             
  setFieldValue('a', prev => prev + 1); // Second update uses stale value                           
  Or:                                                                                               
  setFieldValue('email', 'invalid');                                                                
  validateField('email'); // Validates old value, returns no error          
```                    
                                                                                                    
  Suggestion                                                                                        
                                                                                                    
  Consider applying the same stateRef.current pattern to all imperative methods that access form    
  values.  
                                                                                                    
  ---🤖 This analysis was generated by an experimental QA agent I'm building. Feedback welcome!

PR #3949, opened 2024-01-30
Fix #3948 - Changing state was also causing change of initial value

### LINKED ISSUE #3948: Formik.dirty is not reacting to nested changes
### Current Behavior

I have an object with three-level nested object
When I change it formik values is changing but formik.dirty is not reacting except if i add/remove elements

### Expected behavior
formik.dirty should react to nested objects changes 

### Reproducible example

https://codesandbox.io/p/sandbox/nested-schema-bug-gy44pw?file=/index.js

### My environment

| Formik           | 2.4.5 | 
| React            | nextjs 13.2.3 | 
| TypeScript       | no | 
| Browser          | arc 1.27.3| 
| npm/Yarn         | npm | 
| Operating System | macOS 13.5.1|

DeveloperRaj:
Resolves Issue #3948

This addresses the problem of the dirty field not updating when the value of a nested object changes. The root cause of this issue is that the `initialValues` coming from props were directly assigned to the `useRef`, which did not perform a deep copy. Without a deep copy, it was modifying the original `initialValues` props along with the current state. Consequently, when comparing them for equality, the result was true

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -1,5 +1,6 @@
 import deepmerge from 'deepmerge';
 import isPlainObject from 'lodash/isPlainObject';
+import cloneDeep from 'lodash/cloneDeep';
 import * as React from 'react';
 import isEqual from 'react-fast-compare';
 import invariant from 'tiny-warning';
@@ -29,7 +30,6 @@ import {
   setIn,
   setNestedObjectValues,
 } from './utils';
-import cloneDeep from 'lodash/cloneDeep';
 
 type FormikMessage<Values> =
   | { type: 'SUBMIT_ATTEMPT' }
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -29,6 +29,7 @@ import {
   setIn,
   setNestedObjectValues,
 } from './utils';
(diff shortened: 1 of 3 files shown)


packages/formik/src/Formik.tsx
```diff
@@ -29,6 +29,7 @@ import {
   setIn,
   setNestedObjectValues,
 } from './utils';
+import cloneDeep from 'lodash/cloneDeep';
```
quantizor, line 32: Can you move this import up with the other lodash one and add a changeset please?

DeveloperRaj: Done

quantizor:
@DeveloperRaj thank you, last thing can you add a unit test that verifies the `initialValues` object is not modified?

DeveloperRaj:
> @DeveloperRaj thank you, last thing can you add a unit test that verifies the `initialValues` object is not modified?

@quantizor Thanks for suggestion of test. I have added a new test that checks that initial values are not being modified when we do handleChange or setValues etc. It looks a bit messy but had no option since this bug would repro only happen when we have deeply nested objects otherwise it would work just fine even without my code change.

quantizor (approved):
Thanks!

PR #3961, opened 2024-03-14
Fix value type

devcrazygit:
Form values can also be number, boolean, especially when you customize the form control.

In React Native, I created a new customized form select and the value is number. In this case, this causes error because `eventOrTextValue.target` or `eventOrTextValue.currentTarget` is undefined.

```
if (isString(eventOrTextValue)) {
        // If we can, persist the event
        // @see https://reactjs.org/docs/events.html#event-pooling
        if ((eventOrTextValue as any).persist) {
          (eventOrTextValue as React.ChangeEvent<any>).persist();
        }
        const target = eventOrTextValue.target
          ? (eventOrTextValue as React.ChangeEvent<any>).target
          : (eventOrTextValue as React.ChangeEvent<any>).currentTarget;

```

Can we fix to support such cases?

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -598,7 +598,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   const executeChange = React.useCallback(
-    (eventOrTextValue: string | React.ChangeEvent<any>, maybePath?: string) => {
+    (eventOrTextValue: string | number | boolean | React.ChangeEvent<any>, maybePath?: string) => {
       // By default, assume that the first argument is a string. This allows us to use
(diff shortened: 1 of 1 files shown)


.changeset/khaki-cherries-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'formik': major
```
quantizor, line 2: Is widening a type a breaking change?

devcrazygit: @quantizor I just walk through changeset tutorial.
What should I put in instead of `major`? `minor` is ok?
Or could you write code suggestion here?

quantizor: I think probably just a `patch` is ok

Issue #3960, opened 2024-03-07
Call formik.setFieldValue() with previous value as parameter


wilhei:
### Current Behavior

It is not possible to call formik.setFieldValue('name', value) multiple times async. without overwriting the other values. Let's take a look at following simple example:

      const onButtonClick = () =>{
        Array.from(Array(5).keys())
         .forEach((num) => formik.setFieldValue("testvalue", formik.values.testvalue + num));
        };
      }

The problem is, that we end up with a result, where only one num was added, instead of a sum of all. All previous added nums were overwritten.

### Desired Behavior

Beeing able to pass a function containing the previous value as parameter (like we can do in react useState setter method).

    const onButtonClick = () =>{
        Array.from(Array(5).keys())
         .forEach((num) => formik.setFieldValue("testvalue", (prev) => prev + num));
        };
      }

So in this example we do not use the "start value" everywhere, but we always use the already updated value of previous called functions.

### Suggested Solution

see Desired Behavior

### Who does this impact? Who is this for?

All react developer, which need to run multiple async funtions. In my example I upload multiple images and update the uploading states async.

### Describe alternatives you've considered

Use a React.useState variable and use the setter method with previous value. But I need to sync it all the time with the formik values to be able to use the formik.dirty variable.

### Additional context

noting

ubaidkolad:
This would be really helpful, any update on this?

sovetski:
any update?

gdeggau:
any update?

stupidSheep:
It's been merged and is available in 2.4.9 (and possibly earlier)

@ChronicusUA Many thanks for the PR! This is definitely making things easier!

PR #3968, opened 2024-04-24
Feat: 3960. Update setFieldValue with setter function

ChronicusUA:
Feature for: #3960

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -583,7 +583,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
 
   const setFieldValue = useEventCallback(
     (field: string, value: React.SetStateAction<any>, shouldValidate?: boolean) => {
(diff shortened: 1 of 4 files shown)


packages/formik/src/Formik.tsx
```diff
@@ -582,18 +582,20 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   );
 
   const setFieldValue = useEventCallback(
-    (field: string, value: any, shouldValidate?: boolean) => {
+    (field: string, value: React.SetStateAction<any>, shouldValidate?: boolean) => {
+      const resolvedValue = isFunction(value) ? value(state.values[field]) : value;
```
erashu212, line 586: we can use `getIn` instead of `value(state.values[field])`.

E.g
```suggestion
const resolvedValue = isFunction(value) ? value(getIn(state.values, field)) : value;
```

ChronicusUA: Fixed

quantizor (approved):
Great idea!

PR #3972, opened 2024-05-14
fix(3942): added support for NativeSyntheticEvent

### LINKED ISSUE #3942: Support for React Native's SyntheticEvent type
## Feature request

The Formik docs say that Formik is 100% compatible with React Native, but this is slightly untrue if you're using TypeScript. Formik's functions don't accept React Native's event type `NativeSyntheticEvent`, and the structure of the event is also slightly different and the value of the change is stored elsewhere, so I'm not even sure if Formik supports `NativeSyntheticEvent`. Because of this I have to convert my event to `ChangeEvent`s before I can pass them to Formik's functions.

### Current Behavior

`NativeSyntheticEvent` not supported.

### Desired Behavior

`NativeSyntheticEvent` to be supported

### Suggested Solution

Have all Formik functions that accept an event also be able to accept and handle `NativeSyntheticEvent`

This impacts developers who use Formik with Typescript+React Native

Current alternative is to convert the `NativeSyntheticEvent` to a `ChangeEvent` before passing it to the Formik function

erashu212:
Closes #3942

### DIFF
--- a/packages/formik/src/Formik.tsx
+++ b/packages/formik/src/Formik.tsx
@@ -2,6 +2,7 @@ import deepmerge from 'deepmerge';
 import isPlainObject from 'lodash/isPlainObject';
 import cloneDeep from 'lodash/cloneDeep';
 import * as React from 'react';
+import { NativeSyntheticEvent } from 'react-native'
 import isEqual from 'react-fast-compare';
 import invariant from 'tiny-warning';
 import { FieldConfig } from './Field';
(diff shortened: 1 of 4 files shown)


packages/formik/src/types.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { NativeSyntheticEvent } from 'react-native';
```
quantizor, line 2:
```suggestion
import type { NativeSyntheticEvent } from 'react-native';
```

packages/formik/src/Formik.tsx
```diff
@@ -2,6 +2,7 @@ import deepmerge from 'deepmerge';
 import isPlainObject from 'lodash/isPlainObject';
 import cloneDeep from 'lodash/cloneDeep';
 import * as React from 'react';
+import { NativeSyntheticEvent } from 'react-native'
```
quantizor, line 5:
```suggestion
import type { NativeSyntheticEvent } from 'react-native'
```

quantizor:
Let's actually just inline the type instead of requiring the user to install `react-native` if they're not using it, etc:

From 

```
export interface NativeSyntheticEvent<T>
  extends React.BaseSyntheticEvent<
    T,
    React.ElementRef<React.ComponentType<unknown>>,
    React.ElementRef<React.ComponentType<unknown>>
  > {}```

PR #3976, opened 2024-06-25
Make formik sync when validation functions are sync

fbarbare:


packages/formik/src/Formik.tsx
```diff
@@ -300,26 +294,43 @@ export function useFormik<Values extends FormikValues = FormikValues>({
             prev = setIn(prev, fieldKeysWithValidation[index], curr);
           }
           return prev;
-        }, {})
-      );
+        }, {} as FormikErrors<Values>);
```
bonimba, line 297: you should be able to move this type in the `.reduce<FormikErrors<Values>>`
```suggestion
        }, {});
```

fbarbare:
@jaredpalmer This change goes in the direction that you are mentioning in https://github.com/jaredpalmer/formik/pull/2140 but is backwards compatible since it's not removing any functionality. The deprecation of the async validation methods could be done in a later stage so we can first make things more synchronous (which will improve performance of a lot of apps) and then proceed with the removal of the async stuff.
Any thoughts?

bonimba (approved):
👏🏻 Well done. Amazing update

packages/formik/src/Formik.tsx
fbarbare: I need it to be casted in the `.reduce` I think, no?

bonimba: this should give yo the same outcome
```
       fieldErrorsList.reduce<FormikErrors<Values>>((prev, curr, index) => {
          if (curr === 'DO_NOT_DELETE_YOU_WILL_BE_FIRED') {
            return prev;
          }
          if (curr) {
            prev = setIn(prev, fieldKeysWithValidation[index], curr);
          }
          return prev;
        }, {})
```

fbarbare: Is does. Generics and casting are essentially the same thing though 🙁 
I made the change anyway

quantizor:
@fbarbare could you update against `main` when you get a moment? This branch looks awesome!

quantizor:
Would love to see the benchmarks run for this vs current `main` as well

fbarbare:
@quantizor Thanks for looking at this PR! I have updated the branch with `main` and also added a changeset.

.changeset/make-formik-possibly-synchronous.md
```diff
@@ -0,0 +1,5 @@
+---
+'formik': minor
```
fbarbare, line 2: @quantizor we might want to make this a major version bumb in the end.

I had not thought of that initially but if anyone implemented things as `setFieldValue(..).catch(...)` then it would break if their validation functions would be sync. It would be completely fine if they would implement it as `await setFieldValue(..)` or `void setFieldValue(..)` though.

What are your thoughts?

PR #3978, opened 2024-07-17
Typo_Fixed_in_useField.md

ankur-2dp:
This Pull request fixes a typing error that was present in the /docs/api/useField.md .

### DIFF
--- a/docs/api/useField.md
+++ b/docs/api/useField.md
@@ -3,7 +3,7 @@ id: useField
 title: useField()
 ---
 
-`useField` is a custom React hook that will automagically help you hook up inputs to Formik. You can and should use it to build your own custom input primitives. There are 2 ways to use it.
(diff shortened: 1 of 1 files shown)


quantizor:
In this case it's not a typo, but it's also not "magic" so changing the wording probably makes sense.

docs/api/useField.md
```diff
@@ -3,7 +3,7 @@ id: useField
 title: useField()
 ---
 
-`useField` is a custom React hook that will automagically help you hook up inputs to Formik. You can and should use it to build your own custom input primitives. There are 2 ways to use it.
+`useField` is a custom React hook that will automatically help you hook up inputs to Formik. You can and should use it to build your own custom input primitives. There are 2 ways to use it.
```
quantizor, line 6:
```suggestion
`useField` is a React hook that will help you thread Formik behaviors into arbitrary field components. You can use it to build your own custom input primitives. There are two ways to use it.
```

quantizor (approved):

PR #3982, opened 2024-08-13
`FieldArray` - make array helpers more type-safe

burtek:
Currently used `push<X = T[number]>(obj: X): void;` allows to push `string` to `number[]`. By changing the signature to  `push<X extends T[number] = T[number]>(obj: X): void;` it won't be possible any more.


burtek:
Or, we can get rid of generics here altogether and just do `push(obj: [number]): void;`

quantizor (approved):
This seems right to me.

Issue #2172, opened 2020-01-07
validateOnMount: Initial value for is valid should be whether the initial values are valid, when validateOnMount is true


andycarrell:
### Current Behavior
When `validateOnMount` is true, initial value for `isValid` is 'true', then validation occurs on initial values (which could set `isValid` to 'false').

Screenshot taken from reproducible example - see console for inconsistency - `isValid` has a different value, for the same form value (empty string in the example):

![image](https://user-images.githubusercontent.com/22784983/71859563-35ed9d00-3154-11ea-9ea4-fbb0b5f8cb55.png)

### Expected behavior
The first render should return the result of validating the initial values for isValid.

### Reproducible example
https://codesandbox.io/s/formik-codesandbox-template-kegrs

### Additional context

Would this problem be solved by setting initial errors? I believe possibly, although that would require running validation manually, defeating the purpose of validateOnMount.

The example is a simplification - application code has a far more complex validation schema + more initial values.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Formik         | 2.1.1 |
| React          | 16.12.0  |
| TypeScript       | - |
| Browser          | Codesandbox? && Chrome |
| npm/Yarn         | 6.13.4 |
| Operating System | MacOSX 10.15.2 |

omkar-joshi:
I'd like to help with this

jaredpalmer:
Go for it!

dumbape:
> I'd like to help with this

Any progress on this?

omkar-joshi:
Yeah. Putting up a PR in a couple of days

koredefashokun:
@omkar-joshi is there any progress on this? If you're still working on the PR, I would really appreciate if you could explain any workarounds for this issue at this time. Also, is there any way to prevent Formik from setting errors after running `validateOnMount`. I would like to set the isValid value, but not have errors on the fields. Thanks for your anticipated help.

btiwaree:
> Yeah. Putting up a PR in a couple of days

Hi @omkar-joshi any update on this? appreciate your contribution towards OSS.

semopz:
> @omkar-joshi is there any progress on this? If you're still working on the PR, I would really appreciate if you could explain any workarounds for this issue at this time. Also, is there any way to prevent Formik from setting errors after running `validateOnMount`. I would like to set the isValid value, but not have errors on the fields. Thanks for your anticipated help.

I do this by wrapping my field component where I do the following
```ts
[...]

PR #3984, opened 2024-08-26
fix(formik): validateOnMount initialize isValid to false until validation performed

anonkey:
Related to [2172](https://github.com/jaredpalmer/formik/issues/2172)


packages/formik/src/Formik.tsx
```diff
@@ -152,6 +152,7 @@ export function useFormik<Values extends FormikValues = FormikValues>({
   const initialTouched = React.useRef(props.initialTouched || emptyTouched);
   const initialStatus = React.useRef(props.initialStatus);
   const isMounted = React.useRef<boolean>(false);
+  const validated = React.useRef<boolean>(false);
```
quantizor, line 155: I think your move to add a ref here is right, but we should be running validation synchronously on mount if `validateOnMount` is set such that validation has occurred within the first render pass and there is no potential for the boolean to flip either way.

anonkey: @quantizor  any hint ? this cause annoying UI issue for me

anonkey: Any chance to fix it ? I really need this fix

anonkey:
@quantizor can i have some clue on how you think handling sync validation if validate function is async please ? this bug cause ugly troubles in my ui i wish i solve those

PR #3998, opened 2024-10-08
Fix: Render children in MySelect component to display options

rohitKumar38344:
The original example did not render the **children** prop, which resulted in the **select** element not displaying the options passed as children. This PR updates the **MySelect** component to properly handle the **children** prop and render the options inside the dropdown.

### Changes Made
Updated MySelect component to include the children prop inside the select element.
This ensures that the options passed as children (e.g., <option> tags) are correctly displayed in the dropdown list.

### Issue Addressed
Incorrect example in the documentation where options passed to the MySelect component were not being rendered.

### Testing
Verified that the options now render correctly in the MySelect component by testing it with the provided example from the tutorial.

### Additional Notes
This PR only fixes the example in the docs and does not introduce any breaking changes.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -849,12 +849,13 @@ const MyCheckbox = ({ children, ...props }) => {
   );
 };
 
-const MySelect = ({ label, ...props }) => {
+const MySelect = ({children, label, ...props }) => {
   const [field, meta] = useField(props);
   return (
     <div>
       <label htmlFor={props.id || props.name}>{label}</label>
       <select {...field} {...props} />
+      {children}
       {meta.touched && meta.error ? (
         <div className="error">{meta.error}</div>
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -849,12 +849,13 @@ const MyCheckbox = ({ children, ...props }) => {
   );
 };
 
-const MySelect = ({ label, ...props }) => {
+const MySelect = ({children, label, ...props }) => {
```
quantizor, line 852:
```suggestion
const MySelect = ({ children, label, ...props }) => {
```

quantizor (approved):
Good catch, thanks 😄

rohitKumar38344:
Even after merging the commit why website is not reflecting the updated changes

PR #4001, opened 2024-10-18
chore: fix release workflow

quantizor:
This old style of tag seems to have been removed and just migrating to the recommended way to specify version based on the changeset action [README](https://github.com/changesets/action?tab=readme-ov-file#with-publishing).


Philipp91:
It would be great to merge this, to unblock the releases.

PR #4013, opened 2024-12-08
fix: make validateField return type consistent 

### LINKED ISSUE #2021: return value for validateField
### Current Behavior

Currently, calling `validateField('firstName')` returns `void` (or `Promise<void>`. Calling `validateForm()`, on the other hand, returns the object with errors for the entire form.

### Desired Behavior

`validateField` Returning a single key-value pair, i.e. `{ firstName: 'First name is required' }`, if invalid, or `undefined` if the field is valid.

### Suggested Solution

Adding a return value to the function call. Since it is returning `Promise<void>` right now, I imagine this would not be a breaking change.

Is there any reason in particular that it is set up this way?

Users triggering manual validation of fields. My use case is a custom "Wizard", and I want to validate a limited number of fields for each step, without validating fields on further steps. Something like this:

```ts
export const ContactInfo = () => {
  const formik = useFormikContext()
  const { validateField } = formik
  const { goToStep } = useSequence()
  const next = async () => {
    const validated = await Promise.all([
      validateField('firstName'),
      validateField('lastName'),
      validateField('emailAddress'),
      validateField('phone'),
      validateField('agreeToEmail'),
    ])
    const invalidFields = validated.filter(v => v !== undefined)
    if (!invalidFields) goToStep('nextStep')
  }

  return (
    <div>
      <Heading level={3}>Register for your visit</Heading>
      <Field name="firstName" label="First Name" />
      <Field name="lastName" label="Last Name" />
      <Field name="emailAddress" label="Email" type="email" required />
      <Field name="phone" label="Mobile Phone Number" type="tel" required />
      <Field
        name="agreeToEmail"
        type="checkbox"
        label="I agree to receive email of photos/video taken during the Event"
      />
      <button onClick={next}>Continue</button>
    </div>
  )
}
```

I can use some state to work around this for now, but it's a little hacky.

hcsum:
The validateField function now consistently returns `Promise<string | undefined>`, where `string` represents a validation error message, `undefined` represents a successful validation.

before this change:
- validation via `<Field validate={...}>` returned `Promise<string | undefined>`
- validation via schema returned `Promise<void>`

Closes https://github.com/jaredpalmer/formik/issues/2021

### DIFF
--- a/.changeset/nine-adults-turn.md
+++ b/.changeset/nine-adults-turn.md
@@ -1,0 +1,9 @@
+---
+'formik': patch
+---
+
+The validateField function now consistently returns Promise<string | undefined> across all validation paths, matching its type definition where `string` represents a validation error message, `undefined` represents a successful validation.
+
+while previous implementation:
+- field-level validation returned Promise<string | undefined>
+- default case and schema validation returned Promise<void>
--- a/.changeset/nine-adults-turn.md
+++ b/.changeset/nine-adults-turn.md
@@ -2,8 +2,8 @@
 'formik': patch
 ---
 
-The validateField function now consistently returns Promise<string | undefined> across all validation paths, matching its type definition where `string` represents a validation error message, `undefined` represents a successful validation.
+The validateField function now consistently returns Promise<string | undefined> across all validation paths, where `string` represents a validation error message, `undefined` represents a successful validation.
 
 while previous implementation:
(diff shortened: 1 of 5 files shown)


hcsum:
@quantizor @jaredpalmer happy new year! can help review this small PR?

shreyansh2495 (approved):

PR #4077, opened 2026-07-23
fix: add optional @types/react peerDependency for pnpm

### LINKED ISSUE #4063: `formik` fails with pnpm strict mode — missing peerDependency on `@types/react`
`formik@2.4.9` ships `.d.ts` files that import from `react`:

```ts
```

TypeScript resolves these types through `@types/react`, but `formik` only declares `react` as a peer dependency — not `@types/react`. This works in flat `node_modules` layouts (npm, yarn) because `@types/react` is hoisted from the consuming project. When pnpm hoisting is disabled (`hoist: false`), TypeScript cannot find the React types and fails with:

```
error TS2307: Cannot find module 'react' or its corresponding type declarations.
```

**Workaround:**
```yaml
packageExtensions:
  "formik@2":
    peerDependencies:
      "@types/react": "*"
```

**Suggested fix:** Add `@types/react` to `peerDependencies` in `formik`'s `package.json`.

Hashim1999164:
## Summary
- Adds `@types/react` as an optional peer dependency so Formik's shipped `.d.ts` files resolve under strict pnpm (`hoist: false`).
- Marks it optional via `peerDependenciesMeta` so JS-only consumers are not warned.

Fixes #4063

## Test plan
- [ ] Confirm `packages/formik/package.json` lists `@types/react` under `peerDependencies` and `peerDependenciesMeta.optional`
- [ ] In a pnpm project with `hoist: false`, installing `formik` + `@types/react` should typecheck without `TS2307` for `react`
- [ ] Existing npm/yarn installs should be unchanged


Hashim1999164:
vercel deploy needs a formik team auth click on that link. bro i dont know that lmao, cant approve it from my side. code change itself is just the optional @types/react peerDep for pnpm

PR #4079, opened 2026-08-17
fix(types): remove void from SharedRenderProps.component union

### LINKED ISSUE #4061: [TypeScript] SharedRenderProps.component has incorrect type React.ComponentType<T | void> instead of React.ComponentType<T>
### Description:

The `component` prop in `SharedRenderProps` interface has an incorrect TypeScript type that includes `void` in the union.

**Current type in `types.d.ts:199`:**
```typescript
export interface SharedRenderProps<T> {
    component?: keyof React.JSX.IntrinsicElements | React.ComponentType<T | void>;
    render?: (props: T) => React.ReactNode;
    children?: (props: T) => React.ReactNode;
}
```

**Expected type:**
```typescript
export interface SharedRenderProps<T> {
    component?: keyof React.JSX.IntrinsicElements | React.ComponentType<T>;
    render?: (props: T) => React.ReactNode;
    children?: (props: T) => React.ReactNode;
}
```

### Why this is a bug:

Looking at the runtime implementation in `formik.esm.js` (FieldArray render method, line ~1818):

```javascript
var props = _extends({}, arrayHelpers, {
  form: restOfFormik,
  name: name
});

return component ? createElement(component, props) : render ? render(props) : children ? ...
```

The `component` receives **exactly the same props** as `render`. For `FieldArray`, this is `FieldArrayRenderProps`. The component will **always** receive these props - there's no code path where it receives `void`.

### Impact:

When using the `component` prop with a properly typed component, TypeScript produces false errors because it expects the component to accept `void` as props:

```tsx
interface MyComponentProps {
  form: FormikProps<any>;
  name: string;
  push: (obj: any) => void;
  // ... other FieldArrayRenderProps
}

const MyComponent: React.FC<MyComponentProps> = (props) => { ... };

// TypeScript error: Type 'FC<MyComponentProps>' is not assignable to type 'ComponentType<FieldArrayRenderProps | void>'
<FieldArray name="items" component={MyComponent} />
```

### Workaround:

Use `children` or `render` props instead of `component` for better type safety:

```tsx
<FieldArray name="items">
  {(props) => <MyComponent {...props} />}
</FieldArray>
```

### Suggested fix:

Remove `void` from the union type in `SharedRenderProps`:

```diff
export interface SharedRenderProps<T> {
-    component?: keyof React.JSX.IntrinsicElements | React.ComponentType<T | void>;
+    component?: keyof React.JSX.IntrinsicElements | React.ComponentType<T>;
    render?: (props: T) => React.ReactNode;
    children?: (props: T) => React.ReactNode;
}
```

### Environment:
- Formik version: 2.4.9
- TypeScript version: 4.x+

---

**Labels:** `bug`, `typescript`, `types`

ErfanBagheri404:
## Summary

SharedRenderProps.component was typed as React.ComponentType<T | void>, but at runtime the component prop always receives the same props as render/children (same object passed to createElement) -- no code path gives it void. The void union forces false type errors when passing a properly-typed React.FC<T>.

## Changes

Removed void from the ComponentType union in packages/formik/src/types.tsx.

## Type

Type-only change: keyof React.JSX.IntrinsicElements | React.ComponentType<T>, matching render/children.

Closes #4061


lbesecker195 (approved):
Verified in `Field.tsx`: a non-string `component` is always rendered via `createElement(component, { field, form, ...props, className })`, so it never receives `void`. Type-only tightening. LGTM.

